PR #9, opened 2019-05-04
Try to fix travis.

### LINKED ISSUE #7: Extend package description
Here are some clues from ``pyroma``:

```
(estimagic) PS C:\Users\tobia\git\estimagic> pyroma .
------------------------------
Checking .
Found estimagic
------------------------------
The package had no description!
The package's long_description is quite short.
You should specify what Python versions you support.
Your package does not have keywords data.
Your package does not have license data.
You should specify license in classifiers.
You are using Setuptools or Distribute but do not specify if this package is zip_safe or not. You should specify it, as it defaults to True, which you probably do not want.
------------------------------
Final rating: 0/10
This cheese seems to contain no dairy products
------------------------------
```

janosg:
Make estimagic compliant with our linters.

closes #7

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -31,6 +31,7 @@ install:
 
   # Configure conda
   - conda config --set always_yes yes --set changeps1 no --set quiet true
+  - conda update python=$env:PYTHON_VERSION
   - conda update conda
   - ps: conda install python=$env:PYTHON_VERSION
   - conda info -a
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -31,6 +31,7 @@ install:
 
   # Configure conda
   - conda config --set always_yes yes --set changeps1 no --set quiet true
+  - conda update conda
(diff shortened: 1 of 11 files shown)


tobiasraabe:
In the future, Appveyor will ship newer Miniconda installations with their images. Then, the fix can be removed. Here is their changelog: https://www.appveyor.com/updates/.

tobiasraabe:
The problem seems to be that DLL files are not found on Windows. Discussions can be found ``https://github.com/numpy/numpy/issues/12957`` and ``https://github.com/ContinuumIO/anaconda-issues/issues/10628``, but I cannot find a solution for Python 3.7. The following things have not worked:

- Update Python to the latest 3.7
- Set ``CONDA_DLL_SEARCH_MODIFICATION_ENABLE = 1``
- Choose Numpy 1.15 or latest 1.16.3

I cannot reproduce the problem on my machine.

janosg:
OK, thanks for trying!

PR #12, opened 2019-05-05
Add a dashboard with a convergence plot

### LINKED ISSUE #6: Add dashboard to monitor optimization
## Goal

We want a bokeh dashboard with the following tabs:

- Convergence Monitor: Interactive live convergence plots for different types of parameters. When clicking on one line, the others become opaque
- Profiling Information: show snakeviz profile of criterion function
- System Monitor: show resource usage
- Log: show whatever is printed during the optimization

## Resources

- https://towardsdatascience.com/data-visualization-with-bokeh-in-python-part-iii-a-complete-dashboard-dc6a86aa6e23
- https://github.com/lbaji/convergence_plot

roecla:
addresses parts of #6.


estimagic/dashboard/convergence_plot.py
```diff
@@ -0,0 +1,157 @@
...
+    ToDo: only plot parameters that are not fixed.
+
+    Args:
+        param_df (pandas DataFrame):
+            DataFrame with the initial parameter values, constraints etc.
+
+
+    """
+
+    assert "iteration" not in param_df.index, (
```
janosg: If at all, this error handling should be [here](https://github.com/janosg/estimagic/blob/0a57468954e20de59af420f596012a0b62622533/estimagic/optimization/optimize.py#L141)

But even better would be not to have this restriction. Is 'iteration' displayed somewher? Otherwise we could use a long and cryptic name that is essentially guaranteed to be unique.

estimagic/dashboard/convergence_plot.py
```diff
@@ -0,0 +1,157 @@
...
+
+
+def _choose_color_palettes(param_df):
+    # color tone palettes: bokeh.palettes.Blues9, Greens9, Reds9, Purples9.
+    long_colors = bokeh.palettes.Viridis256
+    return random.sample(long_colors, len(param_df))
+
+
+def _add_convergence_lines(figure, param_data, colors):
+    line_names = [str(x) for x in sorted(param_data.column_names) if x != "iteration"]
```
janosg: What are the parameter names if we have a MultiIndex?

estimagic/dashboard/convergence_plot.py
```diff
@@ -0,0 +1,157 @@
...
+import random
+from functools import partial
+
+import bokeh.palettes
+from bokeh.models import ColumnDataSource
+from bokeh.plotting import figure
+from tornado import gen
+
+
+def setup_convergence_plot(param_df, initial_fitness):
```
janosg: This should be called setup_convergence_plots (plural) or setup_convergence_tab, depending on which is more appropriate. I think I prefer the tab.

estimagic/dashboard/convergence_plot.py
```diff
@@ -0,0 +1,157 @@
...
+
+    """
+    fitness_plot, fitness_data = _fitness_plot(initial_fitness=initial_fitness)
+    param_plots, param_data = _parameter_plots(param_df=param_df)
+    plots = [fitness_plot] + param_plots
+    datasets = [fitness_data] + param_data
+    return plots, datasets
+
+
+def update_convergence_plot(doc, queue, datasets):
```
janosg: Also plural?

estimagic/dashboard/convergence_plot.py
```diff
@@ -0,0 +1,157 @@
...
+        iteration (int):
+            iteration of the parameter vector
+    """
+    entry = {"iteration": [iteration]}
+    entry.update({str(k): [v] for k, v in sr.to_dict().items()})
+    return entry
+
+
+@gen.coroutine
+def _update_convergence_plot(new_values, data):
```
janosg: Again, plural

estimagic/dashboard/convergence_plot.py
```diff
@@ -0,0 +1,157 @@
...
+    iteration = max(data.data["iteration"]) + 1
+    to_add = _convert_sr_for_cds(sr=new_values, iteration=iteration)
+    data.stream(to_add)
+
+
+def _wide_figure(title):
```
janosg: Is it robust to hardcode the size? How will it look on a beamer? We could try this by using a low resolution on a laptop.

estimagic/dashboard/dashboard.py
```diff
@@ -0,0 +1,94 @@
...
+        doc=doc, param_df=param_df, initial_fitness=initial_fitness
+    )
+
+    # this thread is necessary to not lock the server
+    callbacks = partial(_update_dashboard, doc=doc, dashboard_data=data, queue=queue)
+    update_data_thread = Thread(target=callbacks)
+    update_data_thread.start()
+
+
+def _setup_dashboard(doc, param_df, initial_fitness):
```
janosg: _configure_dashboard, _style_dashboard or something like this would be better. Setup could also mean to start running it, so it is too close to run_dashboard.

estimagic/dashboard/server_functions.py
```diff
@@ -0,0 +1,82 @@
...
+    """
+    asyncio.set_event_loop(asyncio.new_event_loop())
+
+    apps = {"/": Application(FunctionHandler(partial(run_dashboard, queue=queue)))}
+
+    server = _setup_server(apps, notebook, port)
+
+    server._loop.start()
+    server.start()
+    if notebook is True:
```
janosg: - The dashboard works from a notebook with notebook=False and notebook=True, but the lines in the convergence plot have different color scales. Any Idea why this is the case? Do we still need the notebook option? If we still need it, could we use something like [this](https://stackoverflow.com/a/24937408) instead?

estimagic/dashboard/convergence_plot.py
```diff
@@ -0,0 +1,157 @@
...
+    conv_p = _wide_figure(title="Parameter Values")
+
+    _add_convergence_lines(figure=conv_p, param_data=param_data, colors=colors)
+
+    return [conv_p], [param_data]
+
+
+def _choose_color_palettes(param_df):
+    # color tone palettes: bokeh.palettes.Blues9, Greens9, Reds9, Purples9.
+    long_colors = bokeh.palettes.Viridis256
```
janosg: Is there a reason why we haven't switched to the [category20 scale](https://bokeh.pydata.org/en/latest/docs/reference/palettes.html) yet? The colors between the lines for each parameter are too similar to distinguish them.

estimagic/dashboard/convergence_plot.py
roecla: At the moment it's not displayed anywhere but I am planning to put it as the x axis label later. I can hard code that though and avoid needing a "pretty" column name.

estimagic/dashboard/convergence_plot.py
janosg: Instead of using plural we will split up the functions.

estimagic/dashboard/convergence_plot.py
janosg: Instead of using plural we will split up the functions.

estimagic/dashboard/convergence_plot.py
janosg: Instead of using plural we will split up the functions.

estimagic/dashboard/convergence_plot.py
roecla: They are tuples. So if for one entry the first level of the Index was 'a' and for the second it was 1, then the parameter name (column name in the ColumnDataSource is ('a', 1).

janosg: That's ok for now and we can also use it in the legend.

estimagic/dashboard/dashboard.py
roecla: Agreed and changed in the next commit.

estimagic/dashboard/server_functions.py
roecla: You are right. I removed the notebook option everywhere.

estimagic/dashboard/convergence_plot.py
roecla: Switched now.

PR #16, opened 2019-05-06
Add tools for numerical differentiation.

janosg:
Notes:

- Use numdifftools to do all the hard work. Especially the richardson extrapolations.
- The [documentation on numdifftools](https://numdifftools.readthedocs.io/en/latest/src/numerical/main.html) is a good introduction to numerical  derivatives.
- The default of all options are what gives us most precision.
- We don't want to support anything that has a complex step in it because it is sometimes not clear what happens when we call numba or Fortran functions with complex numbers.


environment.yml.orig
```diff
@@ -0,0 +1,30 @@
+name: estimagic
```
janosg: Please resolve merge conflict or remove this file.

estimagic/differentiation/first_order_auxiliary.py
```diff
@@ -0,0 +1,37 @@
+"""
+This module contains auxiliary functions for a one step differentiation.
+"""
+
+
+def central(func, f_x0, params_sr, var, h, *func_args, **func_kwargs):
```
janosg: please add docstrings to all functions, documenting all inputs and specifying to which functions the args and kwargs will be passed.

estimagic/tests/examples/test_logit.py
```diff
@@ -0,0 +1,290 @@
...
+
+
+def test_loglikeobs(statsmodels_fixtures):
+    fix = statsmodels_fixtures
+    assert_array_almost_equal(
+        logit_loglikeobs(fix["params"], fix["y"], fix["x"]), fix["loglikeobs"]
+    )
+
+
+def test_gradient_forward(statsmodels_fixtures):
```
janosg: The tests for the numerical derivatives should not be in `test_logit.py` but in `tests/differentiation/test_{corresponding_module}.py

meta.yaml.orig
```diff
@@ -0,0 +1,44 @@
+{% set data = load_setup_py_data() %}
```
janosg: Please resolve merge conflicts or remove the file.

tox.ini.orig
```diff
@@ -0,0 +1,71 @@
+[tox]
```
janosg: merge conflicts

docs/source/differentiation.rst
```diff
@@ -0,0 +1,140 @@
...
+
+And therefore we get:
+
+.. math::
+        f_i(x) = \frac{ 32 (f(x + e_i * h_i) - f(x))  - 12  (f(x + e_i * 2h_i) - f(x)) +
+        f(x + e_i * 4h_i) - f(x)}{12h_i}
+
+Again this calculation can be easily adapted to central and backward differences and
+will therefore not be notated here. For the hessian matrix, this method is repeatedly
+called and the formulas are therefore straightforward.
```
janosg: Please add some references for interested readers.

estimagic/differentiation/hessian.py
```diff
@@ -0,0 +1,79 @@
...
+    func_kwargs = {} if func_kwargs is None else func_kwargs
+    # The hessian matrix is indexed by the vatiable names
+    hess = pd.DataFrame(index=params_sr.index, columns=params_sr.index, dtype=float)
+    # Calculate the value of the function for the observations.
+    f_x0 = func(params_sr, *func_args, **func_kwargs)
+    # Check if a extrapolation method is given.
+    if not extrapolant:
+        # As in the 1st differentiation case, a seperate function is set for each
+        # method.
+        if method == "forward":
```
janosg: you could use getattr instead of the if blocks to select the correct function. 

Also, f is not a very informative name, maybe you can find a better one.

estimagic/differentiation/first_order_auxiliary.py
```diff
@@ -0,0 +1,68 @@
...
+    """
+    This function calculates the central difference.
```
janosg: Docstring should not start with "This function calculates ..." but simply say "Calculate ..."

estimagic/differentiation/gradient.py
```diff
@@ -0,0 +1,56 @@
...
+    func_args = [] if func_args is None else func_args
+    func_kwargs = {} if func_kwargs is None else func_kwargs
+    if extrapolation == "richardson":
+        # For the richardson extrapolation we use, the numdifftools library.
+        grad_np = nd.Gradient(func, method=method)(params_sr, *func_args, *func_kwargs)
+        return pd.Series(data=grad_np, index=params_sr.index)
+    else:
+        grad = pd.Series(index=params_sr.index)
+        f_x0 = func(params_sr, *func_args, **func_kwargs)
+        if method == "forward":
```
janosg: you could write something like:
`import estimagic.differentiation.first_order_auxiliary as ax`
and then use getattr to replace the if, elif, else clauses
`f = getattr(aux, method)

Also, f is maybe not the besto name you can think of?!

estimagic/differentiation/hessian.py
```diff
@@ -0,0 +1,70 @@
...
+    Returns:
+        DataFrame: The index and columns are the index of params_sr.
+
+    """
+    if method != "central":
+        raise ValueError("The given method is not supported.")
+    # set default arguments
+    func_args = [] if func_args is None else func_args
+    func_kwargs = {} if func_kwargs is None else func_kwargs
+    if extrapolation == "richardson":
```
janosg: Extrapollation should be True/False, not a string. A string would only make sense if we support several methods. Do other methods exist?

estimagic/tests/differentiation/test_gradient.py
```diff
@@ -0,0 +1,101 @@
...
+
+
+def test_gradient_forward(statsmodels_fixtures):
+    fix = statsmodels_fixtures
+    assert_series_equal(
+        gradient(
+            logit_loglike,
+            fix["params"],
+            method="forward",
+            extrapolation="",
```
janosg: This is exactly the reason, why a bool is better than a string here.

estimagic/differentiation/first_order_auxiliary.py
MaxBlesch: Changed docstring in all three function of estimagic/differentiation/first_order_auxiliary.py .

estimagic/differentiation/gradient.py
MaxBlesch: Replaced in all functions the if condition with getattr.

estimagic/differentiation/hessian.py
MaxBlesch: Changed from string to bool.

estimagic/tests/differentiation/test_gradient.py
MaxBlesch: Adjusted tests.

PR #22, opened 2019-05-14
Wrap scipy optimizers.

janosg:
Wrap the scipy optimizers that accept bound constraints.

### DIFF
--- a/estimagic/optimization/optimize.py
+++ b/estimagic/optimization/optimize.py
@@ -8,6 +8,7 @@ from time import sleep
 
 import numpy as np
 import pygmo as pg
+import scipy
 
 from estimagic.dashboard.server_functions import run_server
 from estimagic.optimization.process_constraints import process_constraints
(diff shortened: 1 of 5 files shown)


estimagic/optimization/optimize.py
```diff
@@ -353,19 +387,24 @@ def _create_population(problem, algo_options, internal_params):
...
+        x = res.champion_x
+        f = res.champion_f
     params = _params_sr_from_x(x, internal_params, constraints, params)
 
-    res_dict = {
-        "x": params.to_numpy().tolist(),
-        "internal_x": x.tolist(),
-        "f": pygmo_res.champion_f,
-    }
+    res_dict = {"x": params.to_numpy().tolist(), "internal_x": x.tolist(), "f": f}
```
roecla: res_dict[f] has different types depending on the optimizer used. Using scipy (TNC), a np.array of shape (1,) is returned. If nlopt (bobyqa) is used, a float is returned. The later is probably more desirable.

roecla: Resolved in fb49447181d310b40a25b48455ec1abea31905db

janosg: Thanks for solving.

PR #23, opened 2019-05-24
More flexible constraint specification

### LINKED ISSUE #14: Add possibility to estimate variances + correlations instead of covariances.
Note: this can use the cholesky transformation but I'm not sure how we could propagate bounds through that transform.

janosg:
Add the following types of constraints:

- [x] fixed parameters and the value they are fixed have to be specified in the constraint list. Currently they are specified in the params_df.
- [x] pairwise equality constraints between two sets of parameters can be specified
- [x] it is possible to estimate standard deviations and a correlation matrix (currently only covariance matrices can be estimated.)

Moreover I add comprehensive checks that all specified constraints are compatible with each other.

closes #14


hmgaudecker:
>  fixed parameters and the value they are fixed to can be specified in the constraint list and the params_df. Currently only the params_df is possible.

Not having looked into the precise implementation, may I suggest allowing only one place to set parameters to fixed? Probably the constraints are the right place.

I have allowed for two different spots in the past and it is extremely annoying to realise only after a while that you freed up a parameter you had fixed in an initial round of optimisation only in one of the two places...

    There should be one-- and preferably only one --obvious way to do it.

:-)

janosg:
@hmgaudecker, I will do that and only allow for fixing parameters in the constraints list.

docs/source/estimation.rst
```diff
@@ -35,31 +35,28 @@ argument of any criterion function optimized with estimagic's :ref:`minimize`.
 
 It can have the following columns (most of them being optional)
 
-- 'value' (dtype=float): Start values for the optimization. If not provided,
-  start values based on constraints and other information in params are drawn
-  randomly. If provided, we ensure that these values    are taken as start
-  values for local optimizers and are in the initial population of genetic
-  algorithms or pseudo global optimizers.
-- 'lower' (dtype=float): Lower bounds enforced in optimization. Can take the
+- `'value'`` (dtype=float): Start values for the optimization. This column is
```
tobiasraabe: Here are a lot of unbalanced code ticks.

docs/source/estimation.rst
```diff
@@ -35,31 +35,28 @@ argument of any criterion function optimized with estimagic's :ref:`minimize`.
...
+  global optimizers.
+- `'group'`` (dtype=str or None): Indicates in which group (if any)
   a parameter's values will be plotted in the convergence tab of the dashboard.
   Parameters with value None are not plotted.
 
 It is important to distinguish three related but different concepts:
 
-- params_df: the DataFrame described above
-- params_sr: the 'value' column of params. This is the first argument of any
+- `params_df`: the DataFrame described above, sometimes just called `params`.
```
tobiasraabe: And I think you need to ticks for each side to make it literal. Otherwise it is italic or bold.

estimagic/optimization/process_constraints.py
```diff
@@ -95,22 +172,46 @@ def _process_cov_constraint(constraint, params):
...
+        case (str): takes the values 'all_fixed', 'uncorrelated', 'all_free'
+
+    """
+    dim = len(value_mat)
+    off_diagonal_zero = bool((value_mat[np.tril_indices(dim, k=-1)] == 0).all())
+
+    off_diagonal_fixed = bool(fixed_mat[np.tril_indices(dim, k=-1)].all())
+    all_fixed = bool(params_subset["__fixed__"].all())
 
     if all_fixed is True:
```
tobiasraabe: Only ``if all_fixed``.

estimagic/optimization/reparametrize.py
```diff
@@ -42,9 +47,9 @@ def reparametrize_to_internal(params, constraints):
             "ignore", message="indexing past lexsort depth may impact performance."
         )
         actually_fixed = internal["lower"] == internal["upper"]
-        internal.loc[actually_fixed, "fixed"] = True
+        internal.loc[actually_fixed, "__fixed__"] = True
 
-    internal = internal.query("~fixed")
+    internal = internal.query("~__fixed__")
 
     assert (internal["value"] >= internal["lower"]).all(), "Invalid lower bound."
```
tobiasraabe: For comparisons, pandas has a very readable api with ``internal["value"].ge(internal["lower"]).all()``. Here are more options: https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.DataFrame.gt.html. This would also make it clear, that these are pd.Series instead of Numpy arrays. And if you always use internal.lower, you could save three characters :).

estimagic/optimization/reparametrize.py
```diff
@@ -78,8 +83,10 @@ def reparametrize_from_internal(internal_params, constraints, original_params):
 
     for constr in constraints:
         params_subset = reindexed.loc[constr["index"]]
-        if constr["type"] == "covariance":
-            params_sr.update(_covariance_from_internal(params_subset, constr["case"]))
+        if constr["type"] in ["covariance", "sdcorr"]:
```
tobiasraabe: There is no else. I assume there are other checks before which would capture this, but maybe you want to add it so that you are sure. maybe ``else: raise NotImplementedError(f"Constraint {constr['type']} is not implemented")``

estimagic/optimization/optimize.py
```diff
@@ -146,7 +146,7 @@ def minimize(
 
     params = _process_params_df(params)
     fitness_eval = criterion(params["value"], *criterion_args, **criterion_kwargs)
-    constraints = process_constraints(constraints, params)
+    constraints, params = process_constraints(constraints, params)
     internal_params = reparametrize_to_internal(params, constraints)
 
     queue = Queue() if dashboard is True else None
```
tobiasraabe: ``if dashboard else None``

estimagic/optimization/reparametrize.py
```diff
@@ -50,13 +60,34 @@ def reparametrize_to_internal(params, constraints):
         internal.loc[actually_fixed, "__fixed__"] = True
 
     internal = internal.query("~__fixed__")
+    internal.drop(["__fixed__"], axis=1, inplace=True)
```
tobiasraabe: Since you are using pandas >= 0.24.0 you can use ``internal.drop(columns="__fixed__", inplace=True)``

estimagic/optimization/reparametrize.py
hmgaudecker: Seconded. I think you want to handle the entire block for the constraints (lines 32-53 in 647a18d) in one go rather than three steps for this very reason. Very easy to miss a mistyped constraint (yes, you probably *want* to check that elsewhere...) or so and very hard to debug. 

Similarly, lines 55-63 are unnecessarily hard to understand IMHO. Why not something like (untested):
```
internal = internal.query("(fixed == False) | (lower == upper)").drop(["__fixed__"], axis=1)
```

Btw, `_fixed` instead of `__fixed__` would look more pythonic to me

janosg: Thanks for your feedback. I'm currently working on this function and will try to make it more readable. Also, I will add an else statement. But it is not possible to handle all types of constraints in one go. It is important that fixes and equality constraints are applied before all other constraints because:

- In some cases they determine which transformation is applied to the parameters
- We have to check if fixes and equality constraints are compatible with the other constraints. 

I'll ask you for a second round of feedback on this function once I'm done.

hmgaudecker: > But it is not possible to handle all types of constraints in one go. It is important that fixes and equality constraints are applied before all other constraints

How about a custom sort of the list of constraints?

docs/source/estimation.rst
janosg: Thanks, fixed.

docs/source/estimation.rst
janosg: Thanks, fixed.

estimagic/optimization/reparametrize.py
janosg: Sorting the constraints and then running them through one block feels too implicit. Due to some restructuring I already reduced the dependencies between the constraints. Now only fixes have to be applied before equality constraints, otherwise the order is irrelevant. I propose the following:

instead of a column  `'_fixed'` in params, I keep this information in a separate Series called `fixed`. This will be an argument to some of the other constraint functions and thereby clearly shows the dependency. All other constraints will be handled in one block and I will have an else statement in the end. 

I think this dependency was also the reason why I originally wanted the user to specify fixes directly in params and not in constraints, but while using estimagic in skillmodels I discovered that specifying them in constraints is much more convenient. 

This solution would also allow for more code sharing between the application of the constraints and the processing and checking of the user provided constraints. 

@hmgaudecker maybe we can briefly discuss the details in the afternoon.

estimagic/optimization/process_constraints.py
janosg: Thanks, done.

estimagic/optimization/reparametrize.py
janosg: Thanks, done.

estimagic/optimization/optimize.py
janosg: Thanks.

estimagic/optimization/reparametrize.py
janosg: done.

docs/source/estimation.rst
```diff
@@ -76,33 +73,49 @@ Specification of Constraints
...
+      equality constraints are just syntactic sugar and are converted
+      to normal equality constraints internally.
+    - ``'fixed'``: A set of parameters is fixed to some values. In this case
+      the constraints dictionary has to contain a ``'value'`` entry which can
+      be a scalar or an iterable of suitable length.
 
 
-Lower and upper bounds as well as fixed parameters are specified in
-:ref:`params_df`
+Lower and upper bounds are specified in :ref:`params_df`.
```
tobiasraabe: Is it still ``params_df`` or ``params``?

estimagic/optimization/check_constraints.py
```diff
@@ -0,0 +1,123 @@
+import warnings
+
+import pandas as pd
+
+
+def check_compatibility_of_constraints(constraints, params, fixed):
+    """Compatibility checks for constraints.
+
+    Args:
+        constraint (dict)
```
tobiasraabe: constraints and add explanation.

estimagic/optimization/check_constraints.py
```diff
@@ -0,0 +1,123 @@
...
+def _check_no_overlapping_transforming_constraints(constraints, params):
+    counter = pd.Series(index=params.index, data=0, name="constraint_type")
+
+    transforming_types = ["covariance", "sdcorr", "sum", "probability", "increasing"]
+
+    for constr in constraints:
+        if constr["type"] in transforming_types:
+            counter.loc[constr["index"]] += 1
+
+    invalid = counter >= 2
```
tobiasraabe: Maybe ``counter.ge(2)``.

estimagic/optimization/check_constraints.py
```diff
@@ -0,0 +1,123 @@
...
+
+    if invalid.any() > 0:
```
tobiasraabe: Simply ``if invalid.any():``

docs/source/estimation.rst
janosg: params is better, thanks.

estimagic/optimization/check_constraints.py
janosg: Thanks. I also added docstrings to the remaining functions in that module.

estimagic/optimization/check_constraints.py
janosg: yes.

PR #26, opened 2019-06-09
Add comparison plot

### LINKED ISSUE #28: keep comparison_plot informative for very large numbers of models
### Current situation

When too many models are passed to comparison_plot the scatter plot becomes uninformative

### Goal and possible implementation

Keep comparison_plot informative even for very large numbers of models.

One way this could be done is to switch from a scatter_plot to a kde or histogram plot when the number of models passed to comparison_plot exceeds some threshold.

roecla:
## Situation

Often we have many model results and we want to compare them and are interested in patterns between groups of models.

## Solution

An interactive plot that displays (groups of) parameters that allows the user to select single models and groups of models and displays meta data when hovering over particular parameter points.

Thanks to @lbaji for providing an early version and example code for this plot.

## Capabilities of the plot

- When a group column is given it's used to group similar parameters together, just like in the dashboard.
- When a parameter is clicked all other parameters of the same model are highlighted 
- When a "model_class" column is given, groups of models are identified by markers (color coding is also possible) and these groups  can be selected by checking their respectively named box. 

### Look out of the box (with three model classes)

image1

### Selecting points displays their confidence intervals (if supplied) and highlights all parameters belonging to the same model(s):

image2

### If model classes are supplied, classes as a whole can be selected

image3

### When hovering over a point, information on that parameter is shown:

image4

closes #28 

## Remaining Issues:

- when classes exist there is a check in the first model class even though that is not the original state of the plot (compare "look out of the box" (Image1))

- when the user taps into the white only that parameter plot is reset to the original state. The other parameter plots maintain the selection from before. 
    - Workarounds are to either click the reset box on the right or to use the checkbox to unselect all model classes. Both result in the whole plot returning to the original state.


estimagic/tests/visualization/test_comparison_plot.py
```diff
@@ -0,0 +1,404 @@
...
+    res_dict["mod4"]["result_df"]["conf_int_lower"] = res_dict["mod4"]["result_df"][
+        "final_value"
+    ] - 0.5 * np.array(diff2)
+
+    return res_dict
+
+
+@pytest.fixture()
+def df():
+    with open("estimagic/tests/visualization/minimal_expected_df.json", "r") as f:
```
tobiasraabe: Pandas has read_json. no need for context manager.

estimagic/tests/visualization/test_comparison_plot.py
```diff
@@ -0,0 +1,404 @@
...
+        df = pd.DataFrame(json.load(f))
+        df["conf_int_upper"] = pd.np.nan
```
tobiasraabe: The nans should be ``np.nan``.

estimagic/tests/visualization/test_comparison_plot.py
```diff
@@ -0,0 +1,404 @@
...
+from estimagic.visualization.comparison_plot import _find_next_upper
+from estimagic.visualization.comparison_plot import _flatten_dict
+from estimagic.visualization.comparison_plot import _prep_result_df
+
+# ===========================================================================
+# FIXTURES
+# ===========================================================================
+
+
+@pytest.fixture()
```
tobiasraabe: The brackets are not necessary.

estimagic/tests/visualization/minimal_expected_df.json
```diff
@@ -0,0 +1,16 @@
+{
```
tobiasraabe: A csv would be more readable as the nested structure is confusing.

estimagic/tests/visualization/test_comparison_plot.py
```diff
@@ -0,0 +1,404 @@
...
+
+# _prep_result_df
+# ===========================================================================
+
+
+def test_prep_result_df(minimal_res_dict):
+    model = "mod1"
+    model_dict = minimal_res_dict[model]
+    res = _prep_result_df(model_dict, model)
+    expected = pd.DataFrame.from_dict(
```
tobiasraabe: Better save as csv and load it

estimagic/visualization/comparison_plot.py
```diff
@@ -0,0 +1,440 @@
...
+    lower = grouped[["conf_int_lower", "final_value"]].min().min(axis=1)
+    rect_widths = 0.02 * (upper - lower)
+
+    return lower, upper, rect_widths
+
+
+def _determine_plot_heights(df, figure_height):
+    grouped = df.groupby("group")
+    figure_height = _determine_figure_height(df, figure_height)
+    nr_params = grouped["full_name"].unique().apply(len)
```
tobiasraabe: ``n_params`` is more convenient.

estimagic/visualization/comparison_plot.py
```diff
@@ -0,0 +1,440 @@
...
+            step=rect_width,
+        )
+        param_names = df[df["group"] == group]["full_name"].unique()
+        for param in param_names:
+            _add_dodge_and_binned_x(df, param, bins)
+    df["binned_x"].fillna(df["final_value"], inplace=True)
+
+
+def _add_color_column(df, color_dict):
+    if color_dict is None:
```
tobiasraabe: Maybe ``color_dict = {} if color_dict is None else color_dict``

estimagic/visualization/comparison_plot.py
```diff
@@ -0,0 +1,440 @@
...
+        color_dict = {}
+    models = df["model_class"].unique()
+    for m in models:
+        if m not in color_dict.keys():
```
tobiasraabe: You can drop ``.keys()``

estimagic/visualization/comparison_plot.py
```diff
@@ -0,0 +1,440 @@
...
+        for param in param_names:
+            _add_dodge_and_binned_x(df, param, bins)
+    df["binned_x"].fillna(df["final_value"], inplace=True)
+
+
+def _add_color_column(df, color_dict):
+    if color_dict is None:
+        color_dict = {}
+    models = df["model_class"].unique()
+    for m in models:
```
tobiasraabe: ``color_dict = {color_dict.get(m, "#035096") for m in models}``

estimagic/visualization/comparison_plot.py
```diff
@@ -0,0 +1,440 @@
...
+from bokeh.models.callbacks import CustomJS
+from bokeh.models.widgets import CheckboxGroup
+from bokeh.plotting import figure
+from bokeh.plotting import show
+
+from estimagic.optimization.utilities import index_element_to_string
+
+
+def comparison_plot(
+    res_dict, color_dict=None, height=None, width=600, point_estimate_plot_kwargs=None
```
tobiasraabe: Why is width=600? Are these the normal matplotlib sizes in inches or works bokeh differently?

estimagic/visualization/comparison_plot.py
```diff
@@ -0,0 +1,440 @@
...
+        height (int):
+            height of the plot.
+
+        width (int):
+            width of the plot.
+    """
+    df = _df_with_all_results(res_dict)
+    heights = _determine_plot_heights(df, height)
+    lower, upper, rect_widths = _create_bounds_and_rect_widths(df)
+    _add_plot_specs_to_df(df, rect_widths, lower, upper, color_dict)
```
tobiasraabe: The function should return df so it is more explicit.

estimagic/visualization/comparison_plot.py
```diff
@@ -0,0 +1,440 @@
...
+from estimagic.optimization.utilities import index_element_to_string
+
+
+def comparison_plot(
+    res_dict, color_dict=None, height=None, width=600, point_estimate_plot_kwargs=None
+):
+    """Make a comparison plot from a dictionary containing optimization results.
+
+    Args:
+        res_dict (dict): The keys are the names of different models.
```
tobiasraabe: What is the difference between the keys of ``res_dict`` and ``model_class``?

estimagic/visualization/comparison_plot.py
```diff
@@ -0,0 +1,440 @@
...
+def _determine_figure_height(df, figure_height):
+    if figure_height is None:
+        nr_models = len(df["model"].unique())
+        nr_params = len(df["full_name"].unique())
+        figure_height = 8 * max(min(nr_models, 60), 10) * nr_params
+    return figure_height
+
+
+def _add_plot_specs_to_df(df, rect_widths, lower, upper, color_dict):
+    """Add color column, dodge column and binned_x colomun to *df*."""
```
tobiasraabe: typo

estimagic/visualization/comparison_plot.py
```diff
@@ -0,0 +1,440 @@
...
+    _add_color_column(df, color_dict)
```
tobiasraabe: Function should return the df.

estimagic/visualization/comparison_plot.py
```diff
@@ -0,0 +1,440 @@
...
+        df["group"] = "all"
+    if "conf_int_upper" not in df.columns:
+        df["conf_int_upper"] = np.nan
+    if "conf_int_lower" not in df.columns:
+        df["conf_int_lower"] = np.nan
+
+    return df.reset_index(drop=True)
+
+
+def _prep_result_df(model_dict, model):
```
tobiasraabe: In case a dataframe has no column ``final_value`` can we look for ``value`` instead? My use case is that if I have a model description which is not directly out of estimagics min max functions but an input, it does not have ``final_value`` but ``value``. Still, I might one to compare my parametrizations.

estimagic/visualization/comparison_plot.py
```diff
@@ -0,0 +1,440 @@
...
+        df = df.append(result_df, sort=False)
+
+    if "group" not in df.columns:
+        df["group"] = "all"
+    if "conf_int_upper" not in df.columns:
+        df["conf_int_upper"] = np.nan
+    if "conf_int_lower" not in df.columns:
+        df["conf_int_lower"] = np.nan
+
+    return df.reset_index(drop=True)
```
tobiasraabe: Function calls on return values are odd. In this case pass ``ignore_index=True`` to `´df.append`` to renew the index if frames are appended.

estimagic/visualization/comparison_plot.py
```diff
@@ -0,0 +1,440 @@
...
+
+
+def _prep_result_df(model_dict, model):
+    result_df = model_dict["result_df"].reset_index()
+    result_df["model"] = model
+    name_cols = [x for x in ["group", "name"] if x in result_df.columns]
+    result_df["full_name"] = result_df[name_cols].apply(
+        lambda x: index_element_to_string(tuple(x)), axis=1
+    )
+    if "model_class" in model_dict.keys():
```
tobiasraabe: Lines 117-121 become ``result_df["model_class"] = model_dict.get("model_class", "no_class")``.

estimagic/visualization/comparison_plot.py
```diff
@@ -0,0 +1,440 @@
...
+            figure_dict[param_group][param] = param_plot
+            source_dict[param_group][param] = param_src
+            glyph_dict[param_group][param] = point_glyph
+
+    return source_dict, figure_dict, glyph_dict
+
+
+def _add_hover_tool(plot, point_glyph, df):
+    top_cols = ["model", "full_name", "final_value", "model_class"]
+    if "conf_int_lower" in df.columns and "conf_int_upper" in df.columns:
```
tobiasraabe: Quick and dumb question because I do not have the overview. Are the confidence columns not filled with nans if they do not exist? Thus, they are always there.

estimagic/tests/visualization/test_comparison_plot.py
```diff
@@ -0,0 +1,404 @@
...
+    res.set_index(["model", "full_name"], inplace=True)
+    assert_frame_equal(res, expected)
+
+
+def test_df_with_results_with_model_classes(res_dict_with_model_class):
+    with open(
+        "estimagic/tests/visualization/with_model_class_expected_df.json", "r"
+    ) as f:
+        expected = pd.DataFrame(json.load(f))
+        expected["conf_int_upper"] = pd.np.nan
```
tobiasraabe: np.nan

estimagic/tests/visualization/test_comparison_plot.py
```diff
@@ -0,0 +1,404 @@
...
+# _add_color_column
+# ===========================================================================
+
+
+def test_add_color_column_no_dict():
+    color_dict = {}
+    df = pd.DataFrame()
+    df["model_class"] = list("ababab")
+    expected = df.copy(deep=True)
+    expected["color"] = "#035096"
```
tobiasraabe: If that is you default color, you should place it in some configuration file or make it a module variable of comparison_plot.py and import from there.

estimagic/tests/visualization/test_comparison_plot.py
```diff
@@ -0,0 +1,404 @@
...
+def test_add_dodge_and_binned_x_without_class():
+    df = pd.DataFrame()
+    df["final_value"] = [1, 5, 3, 3.5, 0.1, 2.5, 0.2, 0.3, 10]
+    df["full_name"] = "param"
+    df["model_class"] = "no class"
+    param = "param"
+    bins = np.array([-2, -1, 0, 1, 2, 3, 4, 5, 6], dtype=float)
+
+    expected = df.copy(deep=True)
+    expected["lower_edge"] = np.array([1, 5, 3, 3, 0, 2, 0, 0, 6], dtype=float)
```
tobiasraabe: you can pass a list to pandas. no need for a numpy array.

estimagic/tests/visualization/test_comparison_plot.py
```diff
@@ -0,0 +1,404 @@
+import json
+
+import numpy as np
+import pandas as pd
+import pytest
+from pandas.testing import assert_frame_equal
```
tobiasraabe: numpy does ``import numpy.testing as npt``. Maybe you can use ``import pandas.testing as pdt`` which is more explicit.

estimagic/tests/visualization/test_comparison_plot.py
```diff
@@ -0,0 +1,404 @@
...
+
+    _add_dodge_and_binned_x(df, param, bins)
+    assert_frame_equal(df, expected)
+
+
+# _find_next_lower
+# ===========================================================================
+
+
+def test_find_next_lower_with_sorted():
```
tobiasraabe: These repetitive patters can be simplified with [``pytest.mark.parametrize``](https://docs.pytest.org/en/latest/parametrize.html).

estimagic/tests/visualization/test_comparison_plot.py
roecla: changed in 0846b2d5aa68a1880f930eade981bce6930216aa

roecla: Good point! Implemented in 0846b2d5aa68a1880f930eade981bce6930216aa.

estimagic/tests/visualization/test_comparison_plot.py
roecla: Done in 0846b2d5aa68a1880f930eade981bce6930216aa.

estimagic/tests/visualization/test_comparison_plot.py
roecla: 0846b2d5aa68a1880f930eade981bce6930216aa.

estimagic/visualization/comparison_plot.py
roecla: Much nicer! Replaced in 604f9b37a2dc4bc5123ddc29a8fceba432d60334.

estimagic/visualization/comparison_plot.py
roecla: Replaced in 604f9b37a2dc4bc5123ddc29a8fceba432d60334.

estimagic/visualization/comparison_plot.py
roecla: Changed in 604f9b37a2dc4bc5123ddc29a8fceba432d60334.

estimagic/visualization/comparison_plot.py
roecla: Fixed in 604f9b37a2dc4bc5123ddc29a8fceba432d60334.

estimagic/visualization/comparison_plot.py
roecla: Changed in 604f9b37a2dc4bc5123ddc29a8fceba432d60334.

estimagic/visualization/comparison_plot.py
roecla: Adopted in 604f9b37a2dc4bc5123ddc29a8fceba432d60334.

estimagic/visualization/comparison_plot.py
roecla: Adopted in 604f9b37a2dc4bc5123ddc29a8fceba432d60334.

estimagic/tests/visualization/test_comparison_plot.py
roecla: Changed in 0846b2d5aa68a1880f930eade981bce6930216aa.

estimagic/tests/visualization/test_comparison_plot.py
roecla: Done in 0846b2d5aa68a1880f930eade981bce6930216aa.

estimagic/tests/visualization/test_comparison_plot.py
roecla: 0846b2d5aa68a1880f930eade981bce6930216aa.

estimagic/tests/visualization/test_comparison_plot.py
roecla: Adopted in 0846b2d5aa68a1880f930eade981bce6930216aa.

estimagic/tests/visualization/test_comparison_plot.py
roecla: Yes, that makes it much more readable! Switched in a2e17d558f39fcfb95973f9a0559bb1c46f7fe81.

estimagic/tests/visualization/minimal_expected_df.json
roecla: Yes! Implemented in 0846b2d5aa68a1880f930eade981bce6930216aa.

estimagic/visualization/comparison_plot.py
roecla: ``model_class`` is a string, not a dictionary. 

The inputs of the comparison_plot might be be improved.

At the moment the ``res_dict`` is a nested dictionary with the model_name (which must be unique) as key and a dictionary as value. That inner dictionary at the moment has at most two entries: Firstly, ``result_df`` which maps to the params DataFrame returned by ``estimagic.optimization.maximize`` or ``minimize``. Secondly ``model_class`` which maps just to a string (that should be the same for more than one model_name). 

I have considered just having the ``model_class`` as an optional column in the ``result_df`` to have a flat dictionary as input. Then the ``res_dict`` would be flat and a little more user-friendly but it seems superfluous and unelegant to have a column with a hundred times or more the same string for every of possibly many DataFrames passed to comparison_plot. 

I think @janosg  and I talked about this trade-off some time ago and decided on the nested structure but if it is confusing we might want to reconsider, especially as it turns out that we really just have that single additional entry in the inner dictionary.

estimagic/visualization/comparison_plot.py
roecla: In bokeh plot sizes are given in points, not inches. I improved on the default width and height in 499d9ad7f8ee2c909fffeaa16095c6265f9ca048.

estimagic/visualization/comparison_plot.py
roecla: You are right, it should expect ``value`` instead of ``final_value``. Fixed in 504d93d28d61c9b123041da886811b7151cb9c58.

estimagic/visualization/comparison_plot.py
roecla: At the moment at least not.

roecla:
- non integer y ticks are handled starting in cf00a9ff839dbbff22d43a91bc7bbf025039352f. 
- Added the option to have x axis ticks for every x axis. They should be the same because the x axis is shared among parameters of the same group (bdd119547948cba71a50cc1c0c6abbbedcd33bbc).
- default sizes should now be better.

estimagic/visualization/comparison_plot.py
tobiasraabe: Thanks for the explanation. But, I think I was not clear. What is the difference between ``model_class`` and ``model_name``? For me it seems, that we would be fine with just one name for a set of parameters. Why do you need a second level?

If we only need one name, the input to comparison plot just has to be an iterable. Then, if elements of the iterable are of length 2, take the first as the model name and the second as the model's parameters. Thus, you can pass a list of tuples ``[("model_a", params_a), ("model_b", params_b)]`` as well as a dictionary. If the elements are of length 1, number the models and use that as the name.

Having the nested structure does not seem intuitive to me, but I might be wrong as you probably have thought about more use-cases than I have.

estimagic/visualization/comparison_plot.py
tobiasraabe: Not even in ``_df_with_all_results(res_dict)``?

estimagic/visualization/comparison_plot.py
roecla: ``model_name`` uniquely identifies one model, in this case one optimization result. ``model_class`` allows you to group models together. For example in the opening message L-BFGS-B, SLSQP and TNC (each run several times from different starting values) are compared. Other use cases could be having different sets of control variables, instrumenting a variable of interest or not, comparing ML vs GMM ... 

I like your idea of having a list of tuples instead of the nested dictionary. That would make it much easier to use in case a user has no model_classes nor valuable model_names!

tobiasraabe: Now, I understand. What about a combination?

```
parameters = {
    "params_1": [
        ("TNC", params_1_tnc),
        ("L-BFGS-B", "params_1_lbfgsb")
    ],
    "params_2": [
        ("TNC", params_2_tnc),
        ("L-BFGS-B", "params_2_lbfgsb")
    ],
}
```
Or what do you think are the chances of having a plot where you are able to flip the dimensions. I mean not only sorting by ``params_1`` first, but by ``TNC`` and ``L-BFGS-B``? Then, as there is no predefined hierarchy anymore, you could do
```
parameters = [
    ("params_1", "TNC", params_1_tnv),
    ("params_2", "TNC", params_2_tnc),
    ...
]
```
where dimensions are everything but the last tuple entry. I think this would be pretty easy to aggregate just with for loops and zip. Easier than the dictionary. Also it would not get deeply nested in case you add other dimensions.

estimagic/tests/visualization/test_comparison_plot.py
```diff
@@ -0,0 +1,332 @@
...
+from estimagic.visualization.comparison_plot import _find_next_lower
+from estimagic.visualization.comparison_plot import _find_next_upper
+from estimagic.visualization.comparison_plot import _flatten_dict
+from estimagic.visualization.comparison_plot import _prep_result_df
+from estimagic.visualization.comparison_plot import (
+    _reset_index_without_losing_information,
+)
+from estimagic.visualization.comparison_plot import MEDIUMELECTRICBLUE
+
+FIX_PATH = "estimagic/tests/visualization/comparison_plot_fixtures/"
```
tobiasraabe: Checkout ``pathlib`` which is amazing for cross-platform paths.
```
from pathlib import Path
from estimagic.visualization import comparison_plot


FIX_PATH = Path(comparison_plot.__path__, "comparison_plot_fixtures")
```
Then, you do not need ``os.path.join``. It is simply ``path = FIX_PATH / "minimal_res_dict.json"`` and if you need to have a string (most python functions are fine by accepting a path) do ``str(path)``.

estimagic/visualization/comparison_plot.py
roecla: You are right. Fixed in ec59c2a2b28e6c8928c82826fef5e2aa482b1a95.

estimagic/tests/visualization/test_comparison_plot.py
roecla: much more concise this way, too! Thanks for the suggestion!

I think with __ attributes ``__file__`` would be correct. However, [this SO answer](https://stackoverflow.com/a/6209894) suggests using the ``inspect`` module is more robust for getting the path to the current module so I use that. 

Adopted in 1e0580d06bc44eada85c4a4b7b3ff810264fda82.

janosg:
This is a very nice PR and I think the comparison plot will be very useful to diagnose pseudo global or genetic optimizations once we support them. After the few open questions are addressed we can merge.

- [ ] Change interface according to #54 
- [ ] Decide between `__file__` and inspect and use consistently
- [ ] Make all plotted rectangles within one figure (so even across parameter groups) the same size (visually, i.e. in pixels, not in units of the parameters)
- [ ] Let the y-scale go at least up to 5 even if we don't have any parameter values stacked on top of each other to make the comparison plot look more like a rugplot.

roecla:
- y scale goes at least to 5: 

image

- plotted rectangles and plot heights are the same across groups irrespective of the number of parameters in each plot:

image

PR #36, opened 2019-07-09
Fix kwargs differentiation error

MaxBlesch:
This error comes up when evaluation the numdifftools differentiation object:

estimagic/differentiation/jacobian.py in jacobian(func, params_sr, method, extrapolation, func_args, func_kwargs)
     43     if extrapolation:
     44         # For the richardson extrapolation we use, the numdifftools library.
---> 45         jac_np = nd.Jacobian(func, method=method)(params_sr, *func_args, **func_kwargs)
     46         if isinstance(f_x0, pd.Series):
     47             jac = pd.DataFrame(index=f_x0.index, columns=params_sr.index, data=jac_np)

TypeError: __call__() got multiple values for argument 'x'


estimagic/differentiation/differentiation_auxiliary.py
```diff
@@ -1,68 +1,63 @@
...
     """Calculate the central difference.
 
     Args:
-        func (func): A function that maps params_sr into a float.
-        f_x0 (func output): The function value at params_sr, func_args and func_kwargs.
-        params (DataFrame): see :ref:`parmas_df`
-        var (string): The variable of which the finite difference is calculated.
+        internal_func (func): A function wrapping the original function, s.t. it is
+        only dependent on the variable to differentiate.
+        f_x0 (func output): The function value at params_value.
```
janosg: func_output is not a Python type. What types are allowed? I didn't see this last time. Maybe you have to change it at several places.

estimagic/differentiation/differentiation_auxiliary.py
```diff
@@ -1,68 +1,63 @@
...
-def central(func, f_x0, params, var, h, *func_args, **func_kwargs):
+def central(internal_func, f_x0, params_value, i, h):
     """Calculate the central difference.
 
     Args:
-        func (func): A function that maps params_sr into a float.
-        f_x0 (func output): The function value at params_sr, func_args and func_kwargs.
-        params (DataFrame): see :ref:`parmas_df`
-        var (string): The variable of which the finite difference is calculated.
+        internal_func (func): A function wrapping the original function, s.t. it is
```
janosg: The docstring should not explain that `internal_func` it is a wrapper around some other function. It should only explain which type of function is a valid input and which isn't. Being wrapped is not a requirement, it is only necessary, that the function takes just one argument and this argument is a 1d numpy array. This probably applies at several places.

estimagic/differentiation/differentiation_auxiliary.py
MaxBlesch: Changed type to numpy array-

estimagic/differentiation/differentiation_auxiliary.py
MaxBlesch: Changed description to: A function with a 1d numpy array as it's only argument.

PR #44, opened 2019-08-07
Standard erros for SMM/MSM

MaxBlesch:
In this pull request I will implement, the calculation of standard errors for simulated method of moments and generalized method of moments.


estimagic/inference/moment_covs.py
```diff
@@ -0,0 +1,58 @@
...
+        cov (np.array): 2d array covariance matrix of the moments (nmoms, nmoms)
+
+    """
+
+    dev = mom_cond - np.mean(mom_cond, axis=0)
+    cov = dev.T @ dev / mom_cond.shape[0]
+    return cov
+
+
+def gmm_cov(mom_cond, mom_cond_jacob, mom_weight):
```
janosg: `gmm_cov` should be defined at the top of the module. All other functions should be private, i.e. start with and underscore.

estimagic/inference/moment_covs.py
```diff
@@ -0,0 +1,58 @@
...
+    """
+
+    Args:
+        mom_cond (np.array): 2d array matrix of the moment conditions of
+            dimension (nobs, nmoms).
+
+        mom_cond_jacob (np.array): 3d array of the moment condition derivatives
+            w.r.t. the parameters of dimension (nobs, nmoms, nparams).
+
+        mom_weight (np.array):
```
janosg: Please rename `mom_weight` to `weighting_matrix` everywhere and don't shorten it to W. I know that it's sometimes nice when the formula looks exactly like in the book but the #noqa statements make it ugly again.

estimagic/inference/moment_covs.py
```diff
@@ -0,0 +1,58 @@
...
+    Returns:
+        sandwich (np.array):
+        2d array variance-covariance matrix of the GMM estimator of dimension
+        (nparams, nparams)
+
+    """
+
+    # Use notation from Hansen book everywhere. Tell flake8 to ignore capital notation
+    W = mom_weight  # noqa: N806
+    Omega = covariance_moments(mom_cond)  # noqa: N806
```
janosg: Please rename `Omega` to `omega`

estimagic/inference/moment_covs.py
```diff
@@ -0,0 +1,58 @@
...
+    Q = np.mean(mom_cond_jacob, axis=0)  # noqa: N806
+
+    return sandwich_cov(Q, W, Omega, mom_cond.shape[0])
+
+
+def sandwich_cov(Q, W, Omega, nobs):  # noqa: N803
```
janosg: I would just call this `_make_sandwich`, but if you disagree you can leave it as it is.

estimagic/inference/moment_covs.py
```diff
@@ -0,0 +1,58 @@
...
+        sandwich (np.array):
+        2d array variance-covariance matrix of the GMM estimator of dimension
+        (nparams, nparams)
+
+    """
+
+    # Use notation from Hansen book everywhere. Tell flake8 to ignore capital notation
+    W = mom_weight  # noqa: N806
+    Omega = covariance_moments(mom_cond)  # noqa: N806
+    Q = np.mean(mom_cond_jacob, axis=0)  # noqa: N806
```
janosg: If we rename `Q` to `q_hat` we are even more consistent with the book and get rid of the last uppercase variable.

estimagic/tests/inference/test_moment_covs.py
```diff
@@ -0,0 +1,56 @@
...
+
+def test_covariance_moments_unit():
+    moment_cond = np.reshape(np.arange(12), (3, 4))
+    control = np.full((4, 4), 32, dtype=np.float) / 3
+    assert_array_almost_equal(covariance_moments(moment_cond), control)
+
+
+@pytest.fixture
+def statsmodels_fixtures_gmm_cov():
+    fix = {}
```
janosg: Please write a docstring that tells us how exactly you got the fixtures, maybe with link to the code or documentation of the statsmodels function you used.

estimagic/inference/moment_covs.py
MaxBlesch: Reordered.

estimagic/inference/moment_covs.py
MaxBlesch: Ok. Renamed all variables.

estimagic/inference/moment_covs.py
MaxBlesch: I sticked to your notation.

estimagic/tests/inference/test_moment_covs.py
MaxBlesch: I just did some calculations for the matrices by hand. Should I describe that in the docstring?

estimagic/inference/moment_covs.py
hmgaudecker: It definitely has to be public. `gmm_cov` includes additionally the sample dimension, which is not what you want if you are doing msm, where the `nobs`-dimension is already taken care of. 

I can see the benefit for leaving `gmm_cov` as is, although I am not sure whether that 3-d arrays are ever constructed or whether the user would have to set them up. 

In any case, the naming should be improved, but I had no great ideas last week.

estimagic/tests/inference/test_moment_covs.py
janosg: Then it should probably not be called `statsmodels fixtures` ... and yes,  I would prefer a comment that says that you calculated this by hand.

estimagic/inference/moment_covs.py
janosg: Ok, then we make it public. I will have to get more practical experience with MSM before I can really contribute to the discussion.

hmgaudecker: Just look at the what `gmm_cov` does before the call to sandwich. Just aggregating over the observations. And of course this is how it is written in a textbook. 

For GMM, the question is whether that aggregation is a feature or an inconvenience from the point of view what is already calculated. I start leaning towards the latter...

For MSM, the moments are pretty much defined without reference to the dimension of the sample, so it clearly is not usable.

hmgaudecker:
I think this is fine overall (I must confess I have not looked in detail at the very last version).

I would suggest:

- Adding a couple of tests that make sure we get the same results as established routines from statsmodels etc., both for special cases (OLS, IV) and slightly more complex GMM models (if possible there), both for `gmm_cov` and `sandwich_cov`
- Add a warning upon import (`experimental` ?) that the names / interfaces are subject to change without prior warning. I am pretty sure this is not the final word yet.

janosg:
I think we don't need a warning for that as long as the version number is `0.0.x` and the package is registered as `pre-alpha` status on PYPI.

estimagic/tests/inference/test_moment_covs.py
janosg: Thanks.

hmgaudecker:
Cool, I would still suggest adding the extra tests, even if it is just to make sure we do not accidentally divide by `nobs` once too much.

MaxBlesch:
I will implement the test tonight or tomorrow!

MaxBlesch:
I used an OLS example from https://github.com/statsmodels/statsmodels/blob/master/statsmodels/regression/tests/test_cov.py to check our results vs. the one from statsmodels. I am not really sure about the weighting matrix, so I just took the identity. The results don't match yet.

hmgaudecker:
Thanks! Identity matrix sounds right.

Does the estimator produce the same result? What is the magnitude of the discrepancy? Are the OLS standard errors the heteroskedasticity-robust ones?

MaxBlesch:
The robust heteroscedasticity robust covariance matrix agrees with our results. Thanks for the help.

Issue #15, opened 2019-05-06
Add pounders algorithm to our optimizers


janosg:
https://www.mcs.anl.gov/petsc/petsc-current/docs/manualpages/Tao/TAOPOUNDERS.html

Best way to do that is probably [petsc4py](https://anaconda.org/conda-forge/petsc4py/files) but the [documentation](https://www.mcs.anl.gov/petsc/petsc4py-current/docs/usrman/overview.html#petsc-efficient) is very incomplete. 

The relevant import seems to be: `from petsc4py.PETSc import TAO`

tobiasraabe:
While implementing, make sure that ``algo_options`` are passed down to the optimizer.

PR #49, opened 2019-08-22
Pass down algo_options for nlopt and pygmo.

### LINKED ISSUE #3: Rewrite _create_algorithm
## Problem

Currently create_algorithm is very minimal and does not allow to configure the algorithms. 

## Task

- pass the algorithm options through
- document the options for each algorithm
- make sure that all defaults are sensible and change them if necessary

tobiasraabe:
### Description

Currently, the options are only passed for ``scipy``, but we also use ``pygmo``'s and ``nlopt``'s algorithms. They do not have the same API for that.

### Todo

- [x] Add a section in the documentation which links to the three documentations and where to find parameter values.

closes #3

### DIFF
--- a/docs/source/estimation.rst
+++ b/docs/source/estimation.rst
@@ -124,9 +124,14 @@ fixed parameters. For details see :mod:`estimagic.optimization.reparametrize`.
 List of algorithms
 ------------------
 
-.. todo:: Document the algorithms and their arguments. Provide links to the pygmo
-          documentation.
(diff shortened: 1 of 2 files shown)


hmgaudecker:
Don't forget about POUNDERs #15 :-)

tobiasraabe:
@hmgaudecker I added it as a separate note to #15. I am not sure who is working on it, but it will not be overlooked :)

PR #52, opened 2019-08-27
Include the Pounders algorithm in estimagic 

### LINKED ISSUE #15: Add pounders algorithm to our optimizers

mo2561057:
closes #15 

Add the pounders algorithm from the toolkit of advanced optimization. This will be the default optimizer for moment based estimation.


janosg:
There is no windows version of petsc4py on conda-forge (or on any other channel). Quick poll on how we deal with this @hmgaudecker @peisenha @tobiasraabe @roecla 

- Try to add it to conda forge
- Just not support pounders on windows (probably by using [selectors](https://conda.io/projects/conda-build/en/latest/resources/define-metadata.html#preprocessing-selectors) in `meta.yaml` 
- Other ideas?

hmgaudecker:
Is it only petsc4py? That is a thin wrapper, I believe, might be worth asking the maintainer (Lisandro Dalcin, he was very helpful when I had some issues since time back) of he sees any problems.

Now, if we are talking of petsc not being available, stop right here. That is a huge animal written in C or C++...

janosg:
There is also no petsc for windows on conda. BTW: The first sentence on the official installation guide of petsc for windows is: `Are you sure you want to use Microsoft Windows?` :-D

hmgaudecker:
Good one. 

Just point people here: https://docs.microsoft.com/en-us/windows/wsl/install-win10

tobiasraabe:
I agree with Hans-Martin. People are usually on windows 10 so they should use Wsl. It is easy to install and run. And it only affects a special subset of estimagics usability. 

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

janosg:
We have to postpone merging until we find a way to make the conda selectors work. I posted under a related [Issue](https://github.com/conda/conda-build/issues/2093) and hope we get a quick reply. 

In the worst case we have to build the windows package on a remote windows computer.

environment.yml
```diff
@@ -24,6 +24,8 @@ dependencies:
   - seaborn
   - jupyter
   - fuzzywuzzy
+  - mpi
+  - petsc4py
```
hmgaudecker: Make sure you require >= 3.11, I just got bitten by the silent SeparableObjective -> Residual rename elsewhere...

janosg: Will do. I'll finish this over the weekend.

janosg:
I tried to solve our windows installation problems by uploading a fake `petsc4py` package to my anaconda channel (windows versions only). If this works on @tobiasraabe 's windows machine, we can merge. I know this is not ideal, but all the other solutions would be much more work.

PR #53, opened 2019-08-28
Fix bounds for sdcorr matrix.

tobiasraabe:
This is only a temporary solution and a stub PR for your actual solution. The change works for the Robinson Crusoe economy modeled with respy.

### Problem

The problem is that the bounds are applied expecting the sds first and then the elements of the lower triangular. Actually, the matrix is sorted like ``np.tril_indices``. Thus, 0 bounds are not applied to sds, but to correlations which may cause an error.


janosg:
This is actually the solution. I'll just adjust the tests as well.

PR #70, opened 2019-09-08
Logging

raholler:
Four comments:

1. logging function requires two additional arguments: 

     - log_interval (at which iteration to save results at, default right now, every 100th iteration)
     - path_to_sql: where the to save the database to, default right now "logging.db"

2. The Transposing of the parameter DataFrame may be a little inefficient, but I could not find a more straightforward way because of the multiindex,  suggestions very welcome.

3. Right now builds on sqlite3, but pandas also supports sqlalchemy and only keeps sqlite3 for "legacy" reasons. I find sqlite3 much more straightforward than sqlalchemy. In my opinion, sqlalchemy is alittle overkill for a small problem like this one. Nevertheless, if you prefer sqlalchemy, it should be easy enough to change it.

4. Upon restarting optimization, would append the results to the old results, if this is not the behavior that you would like to have, this should be easy enough to change.

### DIFF
--- a/estimagic/optimization/optimize.py
+++ b/estimagic/optimization/optimize.py
@@ -1,5 +1,6 @@
 """Functional wrapper around the pygmo, nlopt and scipy libraries."""
 import json
+import sqlite3
 from collections import namedtuple
 from multiprocessing import Event
 from multiprocessing import Process
@@ -7,6 +8,7 @@ from multiprocessing import Queue
 from pathlib import Path
 
 import numpy as np
+import pandas as pd
 import pygmo as pg
 from scipy.optimize import minimize as scipy_minimize
 
@@ -432,7 +434,7 @@ def _logging(params, fitness_eval, counter, log_interval=100, path_to_sql="loggi
         counter (int): number of current iteration
         log_interval (int): number of iterations until first logging shall
                             occur
-        path_to_sql: path to sql database
+        path_to_sql (str): path to sql database (needs to end with .db)
 
     """
     if counter % log_interval == 0:
@@ -446,14 +448,8 @@ def _logging(params, fitness_eval, counter, log_interval=100, path_to_sql="loggi
         p_final["fitness_eval"] = fitness_eval
         p_final["iteration"] = counter
 
-
         # Save results as sql database
         conn = sqlite3.connect(path_to_sql)
-        p_final.to_sql(
-                name="params",
-                con=conn,
-                if_exists="append",
-                index=False,
-        )
+        p_final.to_sql(name="params", con=conn, if_exists="append", index=False)
         conn.commit()
         conn.close()
--- a/estimagic/optimization/optimize.py
+++ b/estimagic/optimization/optimize.py
@@ -297,6 +297,7 @@ def _create_internal_criterion(
     def internal_criterion(x, counter=c):
         p = _params_from_x(x, internal_params, constraints, params)
         fitness_eval = criterion(p, *criterion_args, **criterion_kwargs)
+        _logging(p, fitness_eval, counter[0])
         if queue is not None:
             queue.put(QueueEntry(iteration=counter[0], params=p, fitness=fitness_eval))
         counter += 1
@@ -418,3 +419,41 @@ def _process_results(res, params, internal_params, constraints, origin):
 
     res_dict["internal_x"] = x.tolist()
     return res_dict, params
+
+
+def _logging(params, fitness_eval, counter, log_interval=100, path_to_sql="logging.db"):
+    """Log parameter values and fitness.
+
+    Saves optimization results every `log_interval` iteration to `path_to_sql`.
+
+    Arg:
+        params (DataFrame): See :ref:`params`
+        fitness_eval (float): current fitness evaluation
+        counter (int): number of current iteration
+        log_interval (int): number of iterations until first logging shall
+                            occur
+        path_to_sql: path to sql database
+
+    """
+    if counter % log_interval == 0:
+        # Reshaping Param DataFrame (Transposing)
+        p = params["value"].reset_index(drop=True).copy()
+        p_trans = pd.DataFrame(p).T
+        p_trans.columns = params["value"].index
+        p_final = p_trans.reset_index(drop=True)
+
+        # Save additional values
+        p_final["fitness_eval"] = fitness_eval
+        p_final["iteration"] = counter
+
+
+        # Save results as sql database
+        conn = sqlite3.connect(path_to_sql)
+        p_final.to_sql(
+                name="params",
+                con=conn,
+                if_exists="append",
+                index=False,
+        )
+        conn.commit()
+        conn.close()


estimagic/optimization/optimize.py
```diff
@@ -418,3 +421,35 @@ def _process_results(res, params, internal_params, constraints, origin):
...
+def _logging(params, fitness_eval, counter, log_interval=100, path_to_sql="logging.db"):
+    """Log parameter values and fitness.
+
+    Saves optimization results every `log_interval` iteration to `path_to_sql`.
+
+    Arg:
+        params (DataFrame): See :ref:`params`
+        fitness_eval (float): current fitness evaluation
+        counter (int): number of current iteration
+        log_interval (int): number of iterations until first logging shall
```
hmgaudecker: The more important feature here seems to be the spacing between triggering a log entry rather than when it starts.

Default should be 1 not 100, IMHO. Unless errors / invalid values mean things are logged post mortem, but that does not seem to be the case? [one usually wants to know from the logs where things went wrong, does not help if it is in between the intervals]

estimagic/optimization/optimize.py
```diff
@@ -418,3 +421,35 @@ def _process_results(res, params, internal_params, constraints, origin):
...
+        p_trans.columns = params["value"].index
+        p_final = p_trans.reset_index(drop=True)
+
+        # Save additional values
+        p_final["fitness_eval"] = fitness_eval
+        p_final["iteration"] = counter
+
+        # Save results as sql database
+        conn = sqlite3.connect(path_to_sql)
+        p_final.to_sql(name="params", con=conn, if_exists="append", index=False)
```
hmgaudecker: If you append to existing databases, one would need to have a unique identifier for which run this is... 

Maybe create a new table each time with the start date at minute precision, overwrite if same date?

estimagic/optimization/optimize.py
raholler: > 
> 
> General: Please leave the connection open all the time. I guess that would require splitting this up into "start_log", "do_log", "finish_log".

well, I guess this needs to be done at a different point in the code? Because, I do not see how the functoin is supposed to know when the optimization is finished...

estimagic/optimization/optimize.py
raholler: so sth like if counter == 1, then get date, save table under data?

janosg:
> General: Please leave the connection open all the time. I guess that would require splitting this up into "start_log", "do_log", "finish_log".

This sounds like we should write a [context manager](https://docs.python.org/3/library/contextlib.html). Then the "start_log", "do_log", "finish_log" becomes something like:

```
with logging as db:
    # code that generates the new entry
    db.add(entry)
```

estimagic/optimization/optimize.py
raholler: Can I ask why this is important?l is this a speed issue?

hmgaudecker: Yes. Most of the overhead comes from there.

Janos is completely right about the context manager, of course. In fact, I bet you can write `with sqlite.connect(path_to_db) as conn:`, at least that is how it works with the db connectors I have worked with. The logging function would then get the connection. Less happening inside of it, good.

raholler: yeah there is something like that in the documentation, but still not sure, what exactly you want from me :), maybe a little bit more ellaboration for noobs like me,

estimagic/optimization/optimize.py
hmgaudecker: I think you would just do that in the context manager, see above. 

I.e., in maximize/minimize:

1. Open the connection (better default name: `estimagic_optimization_log.sqlite`
2. Create the table `YYYY-MM-DD-HH-MM`
3. Pass the table and the connection as keyword arguments

Inside the internal criterion, pretty much as is, but only call it if a connection is around (`_log_params(table=kwargs['table'], conn=kwargs['conn'], params=params, fitness=f if kwargs['conn'] and true_iteration`).

I  like `_log_params` better than logging...

estimagic/optimization/optimize.py
hmgaudecker: Is the below clear?

estimagic/optimization/optimize.py
raholler: true_iteration?

raholler:
So I looked into it, the basic context managers of sqlite3 and slqalchemy (as far as I can tell) only commit or rollback in case of an exception. I think this is not a behavior that we would want, right? We would like it to open, commit and close and rollback only when something with the db goes wrong, not when anything else goes wrong?

So I need to program our own context manager, I guess. Do you want it as class or as a generator?

hmgaudecker:
We most certainly do not need to reinvent the wheel. 

Where do you see any problematic behaviour? We will need to manually trigger commits every now and then, but that should be doable from the connection object, IIRC.

While I am at it: Unless we want to catch exceptions and write the faulty parameters to the DB post mortem, we will need two sets of tables: One with the parameters that we write to before calling the objective function, one with the fitness. Iteration type can go in either one. @janosg -- you'll need to make a design choice here :-)

estimagic/optimization/optimize.py
hmgaudecker: `true_iteration` = iteration of the algorithm as opposed to function evaluation for calculating gradient and friends.

janosg:
> So I need to program our own context manager, I guess. Do you want it as class or as a generator?

This is one of the rare instances where I find a class slightly more intuitive. But you can use whatever is easier to implement for you.

janosg:
> While I am at it: Unless we want to catch exceptions and write the faulty parameters to the DB post mortem, we will need two sets of tables: One with the parameters that we write to before calling the objective function, one with the fitness. Iteration type can go in either one. @janosg -- you'll need to make a design choice here :-)

I think we want to put the actual function evaluation and possibly even the reparametrization step into a try except block. This is easy to implement in `internal_criterion` and useful for the following things:

- Only one case for logging (function value would be NaN in that case)
- Integrates well into what we want to do in numerical derivatives
- We can return a penalty and continue the optimization instead of just aborting it

hmgaudecker:
@janosg: In case you end up implementing the try/except block, please ping @mo2561057. I am pretty sure that exceptions in the objective function will not make it as exceptions through PETSc's C/C++ layer, so he will need to check return codes and manually raise an exception in the wrapper for POUNDERs to get similar behaviour to other optimisers. 
 
Caveat: Could be that I am mis-remembering because exceptions on the GPU did not make it from Numba into Python.

hmgaudecker:
> > So I need to program our own context manager, I guess. Do you want it as class or as a generator?
> 
> This is one of the rare instances where I find a class slightly more intuitive. But you can use whatever is easier to implement for you.

Again, I am 99.9% sure that we do not have a use case that would not be covered by SQLAlchemy's regular context manager. Never used sqlite3, I believe (or not in almost a decade).

janosg:
I close this because after #97 we will implement it very differently and it will be faster to start from the current master than to fix the merge conflicts.

Issue #59, opened 2019-08-30
Allow user to specify steps for numdifftools, scaling


hmgaudecker:
I set the Richardson extrapolation version running yesterday. Still running.

I would be willing to sacrifice some precision for speed here; surely for testing. So it would be great if it was possible to pass the options `step_ratio=2.0, step=1, order=1, num_terms=2` there. Although I must admit that I have not gotten my head what precisely they do yet.

I have not really verified, but it seems to me that the steps are absolute. They certainly were for the manually specified finite-differencing derivative. With large parameter values, this leads to numerical error dominating. To be on the safe side here, I would suggest dividing by the absolute value of the parameter value.

hmgaudecker:
Just a polite inquiry regarding the progress on this one... In my current application, calculating the Jacobian currently takes forever for no reason other than being tied to the defaults and I cannot specify the "central" option because bounds would end up being violated. 

If nobody gets around to this in the near future, pointers on what to change in estimagic would also be appreciated :-)

Issue #57, opened 2019-08-29
Robust standard error calculation for constraint estimations and better standard error interface


janosg:
## Current Situation

- Standard error functions are relatively low level. The user explicitly calculates numerical derivatives and then passes them into the covariance/standard error functions
- It is possible (and happens frequently) that calculation of numerical derivatives fails because the stepsize is so large that some constraints are violated

## Desired Situation

- Standard error calculation can be done in one step
- Numerical derivatives should respect constraints

## Proposed Implementation

Write high level standard error functions that have a similar interface to maximize and minimize. Numerical derivatives are calculated on on a reparametrized version of the objective functions, such that all constraints hold. Then first calculate the covariance matrix for internal_params and use a parametric bootstrap (i.e. monte carlo integration) to calculate the covariance matrix of the external parameters

## Discussion

- Is anybody aware of literature on the topic? Is there any other approach that is similarly easy to apply?
- How would we handle simple box constraints?

hmgaudecker:
- I do not think this is the way to go, see below. But if you do it anyhow, it seems just the quadratic version of #56, i.e., you multiply the each gradient in the product with the derivative of the constraint function. No bootstrap.
- At a more general level, I think this is adding too much complexity and scope for error. What do you do when you are on the border of the parameter space? 
- I would just pass the user's parameters and the constraints. If a constraint is violated, just go the other direction if doing finite differences (i.e. switch from central/forward to backward if upper bound constraint is violated); return `NaN` for the extrapolation (I would hope it can handle that, can't tell yet...). Sure this becomes more problematic for more general restrictions, but if you are calculating standard errors for a covariance matrix that becomes singular when adding &epsilon; you should not be calculating them in the first place.

hmgaudecker:
Update: My calculation of the VCOV-matrix with Richardson's method finished and I would think that at some point a `NaN` was produced -- Jacobian looks good.

janosg:
- It is the inverse of #56, but I thought we can do without cumbersome closed form solutions this time. But maybe you are right and we should do the work. 
- We would check boundary problems of course and refuse to calculate standard errors in that case.  
- Switch adaptively between forward central and backwards differences is hard to implement! We probably couldn't use numdifftools anymore. 
[...]

PR #74, opened 2019-09-15
Implement scaling for parameter vector of optimization.

### LINKED ISSUE #13: Add scaling of the parameter vector
Scale the parameter vector by the inverse of the start parameters in reparametrize_to_internal and undo the scaling in reparametrize_from_internal.

tobiasraabe:
Closes #13.

This PR implements two scaling features for the parameter vector of the optimization. Parameters can either be scaled by starting values or gradients.

### Todo

- [x] Add a little bit of documentation.

### DIFF
--- a/estimagic/optimization/reparametrize.py
+++ b/estimagic/optimization/reparametrize.py
@@ -464,7 +464,9 @@ def _rescale_to_internal(
     """
     scaling = general_options.get("scaling", None)
 
-    if scaling == "start_values":
+    if scaling is None:
+        pass
+    elif scaling == "start_values":
         internal["scaling_factor"] = internal["value"].abs().clip(1)
         internal[["value", "lower", "upper"]] = internal[
             ["value", "lower", "upper"]
@@ -498,8 +500,6 @@ def _rescale_to_internal(
         ].divide(internal["scaling_factor"], axis="index")
     else:
         raise NotImplementedError(f"Scaling method {scaling} is not implemented.")
-
-    breakpoint()
 
     return internal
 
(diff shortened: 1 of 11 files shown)


tobiasraabe:
@janos There is an issue documented starting in ``internal_func`` in differentiation. It is because for scaling I have to set up an internal criterion function and pass it to gradient. There it gets wrapped again.

janosg:
Whether to take derivatives for standard errors on external or internal params was discussed extensively in #57 and we decided that for now we take the derivatives on on the external params. 

Our derivative functions should work for functions that take numpy arrays as well. If they currently don't we have to rewrite `_create_internal_func`. The double wrapping is not very nice but the speed penalty is probably negligible. 

For the scaling we should probably use faster but less precise derivatives. So `extrapolation=False` and maybe even `method="forward"`. We should also use a larger stepsize. This can currently not be specified but I'll change it in #59.

estimagic/optimization/reparametrize.py
```diff
@@ -420,3 +438,89 @@ def _fixed_from_internal(params_subset, constr):
     if value is not None:
         res["value"] = value
     return res["value"]
+
+
+def _rescale_to_internal(
```
janosg: This should be broken up into two functions: One that calculates the scaling factor and one that does the rescaling based on the scaling factor. Then only the actual rescaling should be called in `reparametrize_to_internal`. The calculation of the scaling factor should be done in minimize and the resulting scaling factor should be part of the optimization output. 

My problem with the current implementation is that a call `reparametrize_to_internal` with a value that is different from the start values (for example if we want to calculate a gradient on the internal function), will trigger a re-calculation of the scaling factor. This not only slow but probably also not what the user wants. There should be at most one scaling factor per optimization.

tobiasraabe:
@janosg I implemented the scaling like you suggested. The linting catches some unused arguments. You can go through the list or I can disable it.

janosg:
@peisenha  I requested a review from you because you have implemented precondotioning for an optimizer before.

PR #76, opened 2019-09-26
Pass more arguments to numdifftools

### LINKED ISSUE #59: Allow user to specify steps for numdifftools, scaling

janosg:
closes #59

### DIFF
--- a/estimagic/optimization/optimize.py
+++ b/estimagic/optimization/optimize.py
@@ -177,7 +177,7 @@ def minimize(
         )
         outer_server_process.start()
 
-    result, timing_info = _minimize(
+    result = _minimize(
         criterion=criterion,
         criterion_args=criterion_args,
         criterion_kwargs=criterion_kwargs,
@@ -194,7 +194,7 @@ def minimize(
     if dashboard:
         stop_signal.set()
         outer_server_process.terminate()
-    return result, timing_info
+    return result
 
 
 def _minimize(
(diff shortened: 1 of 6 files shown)


estimagic/optimization/optimize.py
```diff
@@ -177,7 +177,7 @@ def minimize(
         )
         outer_server_process.start()
 
-    result, timing_info = _minimize(
```
hmgaudecker: Does this belong in this PR?

janosg: Not really. I started the PR to tackle several small issues and then realized that the other issues would need a different reviewer. But I would leave it in. It's really just a small clean-up.

hmgaudecker: Just wanted to make sure it would not have any unwanted side effects. I just did not have anything more substantive to say I guess :-)

PR #77, opened 2019-09-30
Multiple parallel optimizations

ChristianZimpelmann:
Run several optimizations in parallel (args, criterium etc. are then entered as a list)

First: all relevant arguments need to be entered as a list.


ChristianZimpelmann:
The code could be a bit shorter if also single optimizations are broadcasted (to lists of length 1).

Then this part is not needed: 

```
if n_opts == 1:
        # Run just a single optimization
        result = _single_minimize(
            criterion=criterion,
            params=params,
            algorithm=algorithm,
            criterion_args=criterion_args,
            criterion_kwargs=criterion_kwargs,
            constraints=constraints,
            general_options=general_options,
            algo_options=algo_options,
            dashboard=dashboard,
            db_options=db_options,
        )
```

However, I think it is clearer to make this distinction and don't initialize multiprocessing if it is unnecessary.

janosg:
Yes, we should keep single optimization as special case!

estimagic/optimization/optimize.py
```diff
@@ -95,29 +92,111 @@ def minimize(
...
+        criterion_kwargs=criterion_kwargs,
+        constraints=constraints,
+        general_options=general_options,
+        algo_options=algo_options,
+        dashboard=dashboard,
+        db_options=db_options,
+    )
+
+    # Find out number of optimizations
+    n_opts = (
```
janosg: it would be nicer if process_optimization_arguments already returns this information, i.e.:

`arguments, n_opts = process_optimization_arguments(...)`

Otherwise we have very mixed levels of abstraction here.

estimagic/optimization/optimize.py
```diff
@@ -207,23 +276,20 @@ def _minimize(
...
             argument and returns a scalar floating point value.
 
-        criterion_args (list or tuple):
-            additional positional arguments for criterion
-
         criterion_kwargs (dict):
             additional keyword arguments for criterion
 
         params (pd.DataFrame):
-            See :ref:`params`.
```
janosg: `:ref:` is correct and `: ref` does probably not generate the correct link.

estimagic/optimization/process_arguments.py
```diff
@@ -0,0 +1,311 @@
...
+            if not isinstance(c, (list, tuple)):
+                raise ValueError(msg)
+
+    # Non-empty 1-dimensional list/tuple
+    else:
+        num_opt = 1
+    return num_opt
+
+
+def broadcast_argument(argument, len_arg, n_opts_total, name):
```
janosg: This function should be private.

estimagic/optimization/process_arguments.py
```diff
@@ -0,0 +1,311 @@
...
+    elif len_arg == n_opts_total:
+        res = argument
+
+    # Argument was entered as too short list/tuple
+    else:
+        raise ValueError(msg)
+    return res
+
+
+def process_optimization_arguments(
```
janosg: This function should be at the top of the module because, so you see it first and can then decide if you also want to look at the implementation details.

estimagic/optimization/process_arguments.py
```diff
@@ -0,0 +1,311 @@
...
+    # (i) expected as scalar,
+    # (ii) expected as scalar or list/tuple,
+    # (iii) expected as list/tuple or nested list/tuple
+
+    args_non_list = [
+        {"candidate": general_options, "scalar_type": dict, "name": "general_options"},
+        {"candidate": dashboard, "scalar_type": bool, "name": "dashboard"},
+        {"candidate": db_options, "scalar_type": dict, "name": "db_options"},
+    ]
+    args_list = [
```
janosg: From here until line 279 the arguments are handled and matched based on their order in some list. Then in 299+ they are put back into a dictionary. 

The position based handling is error prone and in fact there was a mistake in the order of `algo_options` and `criterion_kwargs` that the `test_optimize.py` (merged from the master), discovered. I fixed it for now, but it would be nicer to always handle the arguments based on names, i.e. instead of `args_list` you would have a dictionary where the keys are the names and the values are dictionaries that contain the remaining information, including an entry called "type" or "kind" that contains the information that you currently encode by splitting the arguments into several lists.

estimagic/optimization/process_arguments.py
```diff
@@ -0,0 +1,311 @@
...
+    n_opts_args_nested_list = [
+        _get_n_opt_and_check_type_nested_list_argument(**arg)
+        for arg in args_nested_list
+    ]
+
+    # Calc number of optimizations that should be run
+    n_opts = max(n_opts_args_list + n_opts_args_nested_list)
+
+    # Broadcast inputs
+    criterion, params, algorithm, algo_options, criterion_kwargs, constraints = [
```
janosg: This is a matter of taste, but I prefer to use simple for loops and `append` instead of a list comprehension if the list comprehension spans more than one or two lines.

tox.ini
```diff
@@ -22,6 +22,8 @@ conda_deps =
     pytest-mock
     pytest-xdist
     scipy >= 1.2.1
+    pathos
```
janosg: Do we really use pathos? Otherwise we should remove it here and in the other places.

estimagic/optimization/optimize.py
```diff
@@ -95,19 +92,99 @@ def minimize(
...
+        if "n_cores" not in arguments["general_options"][0]:
+            raise ValueError(
+                "n_cores need to be specified in general_options"
+                + " if multiple optimizations should be run."
+            )
+        n_cores = arguments["general_options"][0]["n_cores"]
+        pool = Pool(processes=n_cores)
+
+        # `Transpose' arguments and run all optimizations in parallel
+        args_transposed = [{a: arguments[a][i] for a in arguments} for i in range(2)]
```
janosg: Two questions:
1. Why `range(2)` and not something that depends on the length of the lists in arguments
2. Is there a reason why we can't move the transpose step into `process_arguments`. If not, I would prefer that.

estimagic/optimization/process_arguments.py
```diff
@@ -0,0 +1,311 @@
...
+            additional configurations for the optimization
+
+        algo_options (dict or list of dicts):
+            algorithm specific configurations for the optimization
+
+        dashboard (bool):
+            whether to create and show a dashboard
+
+        db_options (dict):
+            dictionary with kwargs to be supplied to the run_server function.
```
janosg: This docstring needs a `Returns:` section. I did not write it myself because it depends on your answer to my suggestion of moving the transpose step into `process_arguments`.

ChristianZimpelmann:
> The changes made the code much more readable. Just two questions left, then we can merge. Thanks!

I took care of your remarks and simplified the code once more (the broadcasting function isn't needed anymore)

Also n_opts is now just len(arguments). So I suggest to not return it separately.

PR #79, opened 2019-10-03
Implement penalty interface and elastic net.

tobiasraabe:
This PR establishes an interface for penalty functions.

- The penalty function and options are passed down via ``general_options``.

*List of penalty functions:*
- Elastic Net (Generalization of Lasso and Ridge penalties): The additional parameters are ``lambda`` which scales the penalty and ``alpha`` which determines the ratio between Lasso and Ridge Penalty. ``alpha = 0`` means only Lasso and ``alpha = 1`` means only Ridge.

*Todo:*
- Remove ``_calculate_elastic_net_reference`` because ``internal_params`` in ``internal_criterion`` contains the initial parameter values.
- Ensure that the interaction with scaling works.
- What kind of penalty functions are there?


tobiasraabe:
Community feedback round 🎆! What kind of penalty functions do we need?

hmgaudecker:
Looks cool at a glance. But can you add some more context, please? 

E.g., for which model classes can this be useful? Do I need to transform my parameters? As a case in point, in economic models a "1" might mean risk neutrality and a "0" might even be outside of the parameter space. Usually these things would penalise deviations from "0", right? 

Relatedly, can I apply it only to a subset fo parameters that look more like a regression?

tobiasraabe:
- In my use case, I wanted to estimate better parameters which are only slightly different from another set of parameters. Thus, I added an additional parameter called ``reference`` which can be the start values or any other vector.
- Regarding elastic net, I think it can be useful in general to estimate sparse models, but actually I have no idea how it impacts inference.
- I think you have to standardize the parameters. Otherwise, smaller parameters with big impact are always preferred.
- The reference can be used to not penalize the distance to any other value. In conjunction with your last comment, we need to be able to restrict penalties to subset of parameters and to provide a reference for a subset of parameters. With scaling, we have to ensure that the user defined reference is rescaled to internal values.

hmgaudecker:
Thanks. 

>     * In my use case, I wanted to estimate better parameters which are only slightly different from another set of parameters. Thus, I added an additional parameter called `reference` which can be the start values or any other vector.

I do not quite understand. What is "better" ? What is that reference thing supposed to do? 

>     * Regarding elastic net, I think it can be useful in general to estimate sparse models, but actually I have no idea how it impacts inference.

Yes, and my point is that a sparse model is well-defined in a simple regression context, but not necessarily in the models estimagic seems to be designed for. 

>     * I think you have to standardize the parameters. Otherwise, smaller parameters with big impact are always preferred.

Sounds like this may be quit hard. 

>     * The reference can be used to not penalize the distance to any other value. In conjunction with your last comment, we need to be able to restrict penalties to subset of parameters and to provide a reference for a subset of parameters. With scaling, we have to ensure that the user defined reference is rescaled to internal values.

Same here. 

In general, I think this might be a useful addition eventually. But it looks to me like it requires a lot of very careful thought, both in terms of theory (in which model classes is this even well-defined? What do I do with constraints that do not include the boundary, e.g., variance strictly larger than zero?) and in terms of implementation. With @janosg's plans to first get everything working correctly and then add more, this seems to be rather something for the long run?

tobiasraabe:
I agree with your points. I am fine with postponing this feature until the idea is more mature and we really know if it is useful.

*Correction*: Using LASSO in linear models requires standardized variables, **not** parameters.

janosg:
Doing inference with penalties in arbitrary models is probably not even possible, but definitely out of the scope of estimagic. I see this more as a tool you use in an early stage of an optimization. A good example is skillmodels with bad starting values, where the optimizers are magically drawn towards indefinite covariance matrices. 

In the (not too) long run I'm planning to add estimation functions, e.g. ``estimate_ml`` and ``estimate_msm`` that do the optimization and calculate standard errors. A good trade-off would be if only ``maximize`` and ``minimize`` support the penalty interface but the  estimation functions do not. 

The interface I imagine is that ``maximize`` and ``minimize`` get one additional argument called ``penalties``, which is a list or tuple of callables with the ``params`` DataFrame (the external one) as their only argument. Then we evaluate those functions and add their result to the objective function. Moreover, we can have a module ``estimagic.penalties`` from which you can import pre-implemented penalties like the elastic net and the positive definiteness penalty. Those will have additional tuning parameters that have to be partialed out. 

One question is of course, whether this is something we should implement in estimagic or something that the user can just implement in his objective function. I'm strongly in favor of having it in estimagic because otherwise the user has to code different versions of his objective function, some with and some without penalty and it is much nicer to compose the building blocks in automatically. Moreover, if we stick to the interface I proposed it is extremely easy to implement. 

Another thing I find really important if we implement this is a concise but very good documentation with some background on the topic and a notebook with usage examples of the pre-implemented penalties. The background information is necessary because penalties are a powerful but potentially dangerous tool. Applied in the wrong way they can screw up the smoothness of an optimization problem or even introduce local optima. A good starting point for the background information is the [video](https://www.youtube.com/watch?v=geFER2oVvvU) by the author of the mystic package. 

Regarding the scale problem: I think standardizing variables and standardizing parameters would be substitutes in linear models but the standardization of variables is done because you can't build a mathematical theory that relies on standardizing something at a "good start value". We could keep the start values in ``params`` to make it available in the penalty functions. Then we could have a version of elastic net that weighs very large parameters down. Unfortunately this cannot be integrated with the scaling for the optimization, because penalties operate on external ``params`` while scaling operates on internal ``params``. 

I don't have time to work on this, but could do the review and discuss implementation problems.

hmgaudecker:
Okay, so the idea is more "let's have a softer version of bounds during estimation" rather than "let's look for a sparse model". That sounds reasonable to me and a useful feature indeed! Thanks for the explanation.

janosg:
I think this is outdated. We can re-visit the topic when all the core features are working.

PR #80, opened 2019-10-03
Update issue templates

janosg:
I generated the templates with GitHub's online tool for that (because they recommended it) but I would never do that again. Was way more complicated than writing a markdown file and at some point I had to re-do everything because my changes got lost.


.github/ISSUE_TEMPLATE/descri.md
```diff
@@ -0,0 +1,38 @@
+---
+name: Descri
```
tobiasraabe: I think this will be used as the heading of the template. So description would be better.

.github/ISSUE_TEMPLATE/descri.md
```diff
@@ -0,0 +1,38 @@
...
+
+**Screenshots**
+If applicable, add screenshots to help explain your problem.
+
+**Desktop (please complete the following information):**
+ - OS: [e.g. iOS]
+ - Browser [e.g. chrome, safari]
+ - Version [e.g. 22]
+
+**Smartphone (please complete the following information):**
```
tobiasraabe: Not necessary.

tobiasraabe:
Maybe you should also add an PR template.

.github/ISSUE_TEMPLATE/descri.md
tobiasraabe: Do you really need this and the bug report?

janosg: No, this should be deleted. I'm really too stupid for the GUI template generator on the GitHub page.

.github/PULL_REQUEST_TEMPLATE/add_feature.md
```diff
@@ -0,0 +1,12 @@
+**Current situation**
```
tobiasraabe: This is the same as the feature request. What about this: https://github.com/iza-institute-of-labor-economics/gettsim/blob/master/.github/pull_request_template.md?

janosg: Copied it, except for the `changes.rst` because we don't have one and I will only start to create one when we reach version `0.1.x`

PR #88, opened 2019-10-10
Add pounders to the estimagic interface

janosg:
I merged the pounders pr without actually adding pounders to `minimize` and `maximize`. This PR will change that.


tobiasraabe:
@janosg What is your opinion?

- Should the tests in `test_pounders` use the higher-level API via the general `minimize` or is `minimize_pounders` sufficient? Currently, the criterion functions do not need to handle a series, but a NumPy array.
- Do we want different names for the keys of the results dictionary?

janosg:
1. I think, testing minimize_pounders is enough. 
2. Harmonizing the results dictionaries across different types of estimators is on my [list](https://docs.google.com/document/d/14080eS5j2ug8040wgC3e9SmoGlNEhM246QFbdrr77Lo/edit?usp=sharing) but should be done in a separate PR. 

I would say everything is ready to merge but @mo2561057  should do a review first.

estimagic/optimization/pounders.py
```diff
@@ -21,52 +20,52 @@ def minimize_pounders(
...
+              - termination reason as int.
 
     """
     if sys.platform == "win32":
         raise NotImplementedError("The pounders algorithm is not available on Windows.")
 
-    # we want to get containers for the func verctor and the paras
+    # We want to get containers for the func verctor and the paras.
     size_paras = len(x)
     size_objective = len_out
```
mo2561057: I think we introduced the len_out arg only to avoid an additional evaluation of the objective. I however think it does not really make a difference whether we do that before or after the function call. So I would propose to kick the argument ourt and evaluate the objective in the function for consistency reasons. What do  you think?

janosg: No, I think it should stay in. If we do this evaluation, we want to do it before we call `minimize_pounders`. Otherwise the evaluation is written to the log file.

PR #92, opened 2019-11-05
Adjustments Needed for Skillmodels + Bug Fixes

janosg:
## Bug

The new `process_constraints` sometimes fixes all equality constrained parameters, but the first one has to remain free. This is due to a bug in the consolidation of equality constraints. 

## Skillmodels Adjustment

In skillmodels I have to call `process_constraints` with a params DataFrame where all entries in the "value" columns are np.nan. This raised an error because the constraints are not fulfilled. Now the checks are only performed if at least one "value" is not missing.

### DIFF
--- a/estimagic/optimization/check_constraints.py
+++ b/estimagic/optimization/check_constraints.py
@@ -16,6 +16,9 @@ def check_constraints_are_satisfied(pc, params):
     This should be called before the more specialized constraints are rewritten to
     linear constraints in order to get better error messages!
 
+    We let the checks pass if all parameter values are np.nan. This is useful for the
+    construction of star parameters.
+
     Args:
         pc (list): List of constraints with processed selectors.
         params (pd.DataFrame): See :ref:`params`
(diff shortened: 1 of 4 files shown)


estimagic/optimization/consolidate_constraints.py
```diff
@@ -163,7 +163,7 @@ def _consolidate_fixes_with_equality_constraints(fixed_pc, equality_pc, params):
             assert (
                 len(valcounts) == 1
             ), "Equality constrained parameters cannot be fixed to different values."
-            fixed_value.iloc[eq["index"]] = valcounts.idex[0]
```
tobiasraabe: Ups. Sorry, I have overlooked this gem in the last PR, right?🤦‍♂

estimagic/optimization/check_constraints.py
```diff
@@ -16,6 +16,9 @@ def check_constraints_are_satisfied(pc, params):
     This should be called before the more specialized constraints are rewritten to
     linear constraints in order to get better error messages!
 
+    We let the checks pass if all parameter values are np.nan. This is useful for the
+    construction of star parameters.
```
tobiasraabe: What are star parameters?

estimagic/optimization/consolidate_constraints.py
janosg: Yes ;-)

estimagic/optimization/check_constraints.py
janosg: They are star**t** parameters without a t in the end ...

PR #94, opened 2019-11-18
Miscellaneous small improvements

roecla:
see commits for content.

### DIFF
--- a/estimagic/examples/logit.py
+++ b/estimagic/examples/logit.py
@@ -6,7 +6,7 @@ def logit_loglike(params, y, x):
     """Log-likelihood function of a logit model.
 
     Args:
-        params (pd.DataFrame): The index consists of the parmater names,
+        params (pd.DataFrame): The index consists of the parameter names,
             the "value" column are the parameter values.
         y (np.array): 1d numpy array with the dependent variable
         x (np.array): 2d numpy array with the independent variables
--- a/estimagic/examples/logit.py
+++ b/estimagic/examples/logit.py
@@ -31,9 +31,8 @@ def logit_loglikeobs(params, y, x):
         loglike (np.array): 1d numpy array with likelihood contribution per individual
 
     """
-    param_arr = params["value"].to_numpy()
     q = 2 * y - 1
-    return np.log(1 / (1 + np.exp(-(q * np.dot(x, param_arr)))))
+    return np.log(1 / (1 + np.exp(-(q * np.dot(x, params["value"])))))
 
 
 def logit_gradient(params, y, x):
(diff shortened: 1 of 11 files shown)


estimagic/optimization/optimize.py
```diff
@@ -217,7 +223,12 @@ def _single_minimize(
     """
     simplefilter(action="ignore", category=pd.errors.PerformanceWarning)
     params = _process_params(params)
-    fitness_eval = criterion(params, **criterion_kwargs)
+
+    fitness_factor = -1 if criterion_kwargs.get("_maximization", False) else 1
+    if "_maximization" in criterion_kwargs:
+        del criterion_kwargs["_maximization"]
```
tobiasraabe: Why is ``criterion_kwargs["_maximization"]`` treated differently between https://github.com/OpenSourceEconomics/estimagic/pull/94/files#diff-4fc3d0959b0a8ca5740e893de422065bR75-R80 and the lines here? Can you handle them similary?

estimagic/examples/logit.py
```diff
@@ -22,44 +22,44 @@ def logit_loglikeobs(params, y, x):
...
+            the "value" column are the parameter values.
         y (np.array): 1d numpy array with the dependent variable
         x (np.array): 2d numpy array with the independent variables
 
     Returns:
         loglike (np.array): 1d numpy array with likelihood contribution per individual
 
     """
-    params = params["value"].to_numpy()
+    param_arr = params["value"].to_numpy()
```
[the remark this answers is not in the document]
roecla: You are right, that is more parsimonious! Changed in c4964f17c8e31d290f7e04cd6dbb9fde59183368

estimagic/optimization/optimize.py
roecla: I am not sure I understand what you want.

Would you prefer that I define a `fitness_factor` key with value 1 or -1 already in lines 75-80 instead of having `_maximization` as key and a boolean value?

tobiasraabe: I should have read the code more carefully. The line with the fitness factor is great and handles both maximization and minimization problems.

Why do remove the key in the next lines? If it is necessary, use ``pop`` with default instead of ``get``: ``fitness_factor = -1 if criterion_kwargs.pop("_maximization", False) else 1``.

roecla: Adding the maximization info to `criterion_kwargs` means I must remove it before they are passed on to `criterion` where the additional entry might cause problems. 

I guess the more general question is why pass `_maximization` as part of `criterion_kwargs` along and there is no good reason to do that anymore. Originally I tried to avoid changing the signature of `create_internal_criterion` but then decided it was worth to add `fitness_factor`.

Given that, I think it makes more sense to move `_maximization` to `general_options` where it does not have to me removed. Did that in bd369d47b0a8275b805a7b056b84af1d659cd512.

tobiasraabe: Very good.

PR #95, opened 2019-11-20
Move estimagic to OSE on Anaconda and add new release.py.

tobiasraabe:
### Changes

- Deleted master build from rtd because it is the same as latest. Badge in Readme also links to latest.
- New release.py.
- Moved to anaconda.org/opensourceeconomics.
- Added bump2version to bump version numbers.


tobiasraabe:
I deleted how-to-maintain. It is way to premature and gets battle-tested by gettsim over the next weeks.

Issue #71, opened 2019-09-12
Logging and output for extremely long running problems


janosg:
## Current situation

The optimization functions return the optimal parameters and some additional information. After #54 this will be just one object. Moreover, after #45 we will have a persistent log of parameters.

## Desired situation

When optimization takes very long (between a few days and potentially weeks) and we run different optimizations in parallel, it is very easy to lose track. It has definitely happened to me that I found an optimization result on my computer and had difficulties to find out to which of many models it belonged. 

In the worst case the script that started the optimization does not exist anymore or we forgot the commit we would have to check out to get it. 

Therefore, the result of the optimization functions should (in addition to what we have currently) contain the following information:

- The start and end time of the optimization
- All arguments to `maximize` or `minimize`
- All `criterion_args` and `criterion_kwargs`
- The conda environment (exported via condas python api)
- The complete parameter log
- Optionally: additional information the user provided and that help him to reproduce the same result (e.g. the commit at which he started the optimization or a message that explains why this particular optimization was started.)

## Implementation

Collecting the data is easy. The main question is in what format we want to store it on disk.

Ideally everything would be in one "immutable" file and not just a directory where a user could save or delete files manually.  `.pickle` is not a good idea because some pickle objects will only load in the right environment and the information on the environment is inside the pickle file. `.json` won't work because we have some binary objects. My current favorite is something like a zipped folder but I'm open to other ideas. Inside this folder we could choose the best file type for each information. The environment would be a `.yml` file, for all dictionaries and lists we could first try to store them as `.yml` and only if that fails resort to `.pickle`. 

This is based on a conversation with @fediskhakov. Feel free to comment below. Other people can also add ideas and wishes. Right now there are more pressing issues in estimagic but this will be a nice addition in the long run.

hmgaudecker:
Very nice. But... One place only: the database seems to be the right place for this. You can throw in any type of object there. 

I would put throw the current commit and the output of git diff in there, too. Maybe filtering out too large diffs.

If you are obsessed about immutability, you may be able to achieve that (but probably need a more complex backend than sqlite), but I would trust the user there.

janosg:
I always thought of a database as some kind of rectangular data format... Of course, then the database is the place. That also solves all other implementation problems I could think of. Maybe @raholler wants to give a very brief hackathon presentation on databases next month?

Adding the current commit and the git diff automatically is a really cool idea! 

[...]

PR #97, opened 2019-11-24
Add database functions

janosg:
## Goal

We will use an sql database for persistent logging and to share data between the optimization and the dashboard. However, most contributors to estimagic are not familiar with sqlite or sqlalchemy. Therefore, this PR provides functions to:

- create or load a database with all relevant tables
- update tables in the database in a thread safe way
- read from the database and convert the result to the formats we typically need. 

Given the functions, adding persistent logging to estimagic will be very quick, but I will do it in a separate PR because I want a review first. 

### Backend

We use sqlite3 as database backend. The main reasons for choosing it are that it comes pre-installed with Python and is the only sql implementation that stores all data in a single file. A potential drawback is that sqlite is not designed for a high degree of write concurrency. However, this would only be a problem if someone optimizes an extremely fast criterion function with a heavily parallelized optimizer. 

### Frontend

To interact with the database we use sqlalchemy core. sqlalchemy ORM does not have any obvious advantages for our very simple use case. Just using `pandas.to_sql` and `pandas.read_sql` is not enough because it loads complete tables in memory which would be problematic for long running optimizations. 

### Table Structure

The exact table structure is described in the docstring of `prepare_database`. We use one table per concept (e.g. parameters, criterion_values, ...) to keep the queries simple. Tables that get new data in each iteration have an autoincrement column called iteration. This will replace the counter we currently have in the criterion function. The `append_rows` function ensures that the counters are correctly aligned across tables even if multiple threads  or processes update the database. 

@tobiasraabe and @roecla in your reviews you do not have to focus very much on the implementation details (I have tests!). Instead I would like to hear if we have all tables we need and if the interfaces of the public functions are intuitive and cover all relevant use-cases.


estimagic/logging/read_database.py
```diff
@@ -0,0 +1,164 @@
+"""Functions to read from a database."""
```
hmgaudecker: There is a discrepancy here between the level of generality in the module docstring ("a database") and the two first functions (read rows of one very specific table in a very specific database).

estimagic/logging/create_database.py
```diff
@@ -0,0 +1,238 @@
...
+    gradient_status=0,
+    replace=False,
+):
+    """Return database metadata object with all relevant tables.
+
+    This should always be used to create entirely new databases or to create the
+    tables needed during optimization in an existing database.
+
+    The resulting database has the following tables:
+    - params_history: the complete history of parameters from the optimization. The
```
hmgaudecker: "The index column is 'iteration'"

estimagic/logging/create_database.py
```diff
@@ -0,0 +1,238 @@
...
+    - gradient_history: the complete history of gradient evaluations from the
+      optimization. Same columns as params_history.
+    - criterion_history: the complete history of criterion values from the optimization.
+      The first column is "iteration", the second column is "value".
+    - time_stamps: timestamps from the end of each criterion evaluation. Same columss as
+      criterion_history.
+    - convergence_history: the complete history of convergence criteria from the
+      optimization. The first column is iteration, the other columns are "ftol", "gtol"
+      and "xtol".
+    - start_params: copy of user provided params.
```
hmgaudecker: Is this not always the first row of `params_history`? A utility function to get that back would seem more helpful than storing the same thing twice.

estimagic/logging/create_database.py
```diff
@@ -0,0 +1,238 @@
...
+    - optimization_status: table with one row and one column called "value" which takes
+      the values "scheduled", "running", "success" or "failure". The initial value is
+      "scheduled".
+    - gradient_status: table with one row and one column called "value" which takes
+      can be any float between 0 and 1 and indicates the progress of the gradient
+      calculation. Initialized to 0.
+    - db_options: table with one row and one column called "value". It contains
```
hmgaudecker: If this can still be changed, I would go for the more verbose `dashboard_options` everywhere. To me, `db` is short for database; that had always confused me when using estimagic and it only becomes worse now that there actually is a database backend in it...

estimagic/logging/create_database.py
```diff
@@ -0,0 +1,238 @@
...
+      a dictionary with the dashboard options. Internally this is a PickleType, so the
+      dictionary must be pickle serializable.
+
+    If the function is called with an existing database, only the tables that do not
```
hmgaudecker: So if the history tables exist, they will be appended to?  How do I realise that this is a completely new optimisation then? What if I change the parameters I optimise over and use the same database. This looks like a recipe for disaster if I understand the structure correctly. See my overview comment, too.

estimagic/logging/update_database.py
```diff
@@ -0,0 +1,110 @@
...
+import pickle
+import warnings
+from datetime import datetime
+from pathlib import Path
+
+import sqlalchemy
+
+
+def append_rows(metadata, tables, values):
+    """As append values to tables in an atomic transaction.
```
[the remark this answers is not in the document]
janosg: I agree, thanks.

estimagic/logging/create_database.py
```diff
@@ -0,0 +1,238 @@
...
+
+def prepare_database(
+    path,
+    params,
+    db_options=None,
+    optimization_status="scheduled",
+    gradient_status=0,
+    replace=False,
+):
+    """Return database metadata object with all relevant tables.
```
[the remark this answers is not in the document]
janosg: I changed the docstring.

estimagic/logging/read_database.py
janosg: Changed it. The more general functions are not public (because they still require some knowledge of sqlalchemy), so in fact the module only contains specific functions.

estimagic/logging/create_database.py
janosg: Done.

estimagic/logging/create_database.py
janosg: No, it is not. "start_params" contains the full params DataFrame with bounds columns, the correct Index (whereas the log uses parameter names as index to avoid a MultiIndex), information on fixed parameters, ... 

I made it clearer in the docstring.

estimagic/logging/create_database.py
janosg: Yes, I realized the same thing when I worked on this PR. But I would do it in a separate PR because we it has to be done in many places! I created #98 to discuss solutions.

estimagic/logging/create_database.py
janosg: Addressed in the overview comment.

janosg:
Thanks for the very detailed review!

Regarding the structure: I think there should be one database per optimization (your solution 2). The main reasons are:

- performance: both read and write performance suffer with large and fragmented databases and I made some benchmarks to confirm that this is relevant. 
- limit write concurrency: If too many optimizations write to the same database, it might be busy and estimagic will fall back to pickling. This is not relevant for long running criterion functions, but skillmodels would probably hit the limit. 
- each database has the same structure: otherwise the dashboard code will be horrible. 
- As discussed in #71 it is nice to have everything that belongs to one information in one file that does not get touched anymore after the optimization is finished. 

 The dashboard will be able to monitor several optimizations with the same bokeh server (i.e. you only have to forward one port from a remote machine). There are also ways to connect to multiple databases and make them behave as one in sqlalchemy if I remember correctly, so if this is really a concern, we (probably @raholler) could add a helper function for that. 

I adjusted `prepare_database` so that it deletes and re-creates all tables needed for the optimization, but leaves any other table the user might have in the database untouched. 

I also renamed `metadata` to `database` everywhere because this is more intuitive for users who have never worked with sqlalchemy.

estimagic/logging/create_database.py
```diff
@@ -0,0 +1,253 @@
...
+      calculation. Initialized to ``gradient_status``
+    - db_options: table with one row and one column called "value". It contains
+      a dictionary with the dashboard options. Internally this is a PickleType, so the
+      dictionary must be pickle serializable. Initialized to db_options.
+
+
+    Args:
+        path (str or pathlib.Path): location of the database file. If the file does
+            not exist, it will be created.
+        params (pd.DataFrame): see :ref:`params`
```
tobiasraabe:
```suggestion
        params (pd.DataFrame): see :ref:`params`.
```

estimagic/logging/create_database.py
```diff
@@ -0,0 +1,253 @@
...
+        replace (bool): If true and the database exists, it will be overwritten.
+            Otherwise, data will be appended.
+
+    Returns:
+        database (sqlalchemy.sql.schema.MetaData). The engine that connects
+        to the database can be accessed via ``database.bind``.
+
+    """
+    db_options = {} if db_options is None else db_options
+    gradient_status = float(gradient_status)
```
tobiasraabe: Why float?

estimagic/logging/update_database.py
```diff
@@ -0,0 +1,118 @@
...
+def _handle_exception(statements, database):
+    directory = Path(str(database.bind.url)[10:])
+    if not directory.is_dir():
+        directory = Path(".")
+    directory = directory.resolve()
+
+    for stat in statements:
+        if isinstance(stat, (sqlalchemy.sql.dml.Insert, sqlalchemy.sql.dml.Update)):
+            values = stat.compile().params
+            filename = f"{stat.table.name}_{datetime.now()}.pickle"
```
tobiasraabe: `datetime.now()` returns something like`'2019-11-24 17:09:40.648559'`. The colon definitely causes problems on windows. Maybe something more like this: https://stackoverflow.com/a/10607768.

estimagic/logging/create_database.py
janosg: This is a percentage used to make a progress bar.

estimagic/logging/update_database.py
janosg: Didn't know, changed it.

estimagic/logging/update_database.py
```diff
@@ -0,0 +1,119 @@
...
+def _handle_exception(statements, database):
+    directory = Path(str(database.bind.url)[10:])
+    if not directory.is_dir():
+        directory = Path(".")
+    directory = directory.resolve()
+
+    for stat in statements:
+        if isinstance(stat, (sqlalchemy.sql.dml.Insert, sqlalchemy.sql.dml.Update)):
+            values = stat.compile().params
+            timestamp = time.strftime("%Y%m%d-%H%M%S")
```
tobiasraabe: Maybe even add milliseconds like before? Otherwise you might run into errors with skillmodels?
```suggestion
            timestamp = time.strftime("%Y%m%d-%H%M%S-%f")
```

tobiasraabe: Sorry %f only works with datetime not time.

janosg:
Yes, the user will specify a path to the database, i.e. a name. The timestamp thing is only for the pickle files that get saved when the database update fails. This will probably never be reached in practice but I wanted to make absolutely sure that the logging never raises any errors. 

The other two things will be done in a separate PR when we tackle #71

estimagic/logging/update_database.py
```diff
@@ -0,0 +1,119 @@
...
+import datetime
+import pickle
+import warnings
+from pathlib import Path
+
+import pandas as pd
+import sqlalchemy
+
+
+def append_rows(database, tables, rows):
```
roecla: This is of course partially a matter of taste but to me it would feel more intuitive and less error-prone to pass a dictionary to this with the `tables` as keys and the `rows` as values rather than two lists. 

Also, I think it would be even more user friendly if instead of a `MetaData` object a path could be supplied. The way it is now, I think people may be confused whether they have to save the database to disk after using `append_rows` or whether they have to "close" it (just like files should be closed after being opened).

estimagic/logging/create_database.py
```diff
@@ -0,0 +1,253 @@
...
+    replace=False,
+):
+    """Return database metadata object with all relevant tables for the optimization.
+
+    This should always be used to create entirely new databases or to create the
+    tables needed during optimization in an existing database.
+
+    A new database is created if path does not exist yet or replace=True. Otherwise the
+    existing database is loaded and all tables needed to log the optimization are
+    overwritten.
```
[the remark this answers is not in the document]
janosg: Here only the tables estimagic knows about will be re-created, but the database can have other tables. I clarified the docstring and removed the "replace" Argument. It is not really necessary, since we switched to

estimagic/logging/create_database.py
```diff
@@ -0,0 +1,253 @@
...
+
+    The resulting database has the following tables:
+
+    - params_history: the complete history of parameters from the optimization. The
+      index column is "iteration", the remaining columns are parameter names taken
+      from params["name"].
+    - gradient_history: the complete history of gradient evaluations from the
+      optimization. Same columns as params_history.
```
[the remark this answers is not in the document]
janosg: Iteration counters are not shared, but aligned as long as all updates that belong to one iteration are done in one transaction, i.e in one call to `append_rows`. 
Typically, gradient based optimizers make one gradient evaluation per "real" function evaluation. Do you know any optimizer where this is not the case? If so, I do not really see another solution as logging a separate parameter history for the gradient evaluations. But I would only go there when we get the problem.

estimagic/logging/create_database.py
```diff
@@ -0,0 +1,253 @@
...
+    This should always be used to create entirely new databases or to create the
+    tables needed during optimization in an existing database.
+
+    A new database is created if path does not exist yet or replace=True. Otherwise the
+    existing database is loaded and all tables needed to log the optimization are
+    overwritten.
+
+    The resulting database has the following tables:
+
+    - params_history: the complete history of parameters from the optimization. The
```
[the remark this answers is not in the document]
janosg: This is a usage question. Both options are supported by the database functions. Ideally, I would say we log everything and have an additional boolean Table called `"is_champion"` that is true if the parameter vector was the champion at the time it was written to the database. Then we could only plot the champions in the dashboard. But @tobiasraabe will have to find out if is possible to get this information out of pygmo.

estimagic/logging/read_database.py
```diff
@@ -0,0 +1,164 @@
...
+    return result
+
+
+def read_new_iterations(database, tables, last_retrieved, return_type, limit=None):
+    """Read all iterations after last_retrieved.
+
+    Args:
+        database (sqlalchemy.MetaData)
+        tables (list): List of tables names.
+        n (int): number of rows to retrieve
```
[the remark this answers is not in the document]
janosg: Fixed.

estimagic/logging/update_database.py
janosg: Most database functions have `database` as their first argument and `table` or `tables` as the second one. I agree that the dictionary might be more intuitive in isolation, but the current implementation is more consistent with the other functions. 

Regarding the path: No, for the following reasons:
1. This would introduce a lot of overhead and it is really not how databases are meant to be used. 
2. I find it quite intuitive that there are `prepare_database` and `load_database` functions that both give you exactly what you need to use any function with a database argument. 
3. I know that people will worry how often the database is synchronized with the file on disk. I get this question all the time. (The answer is whenever we do an update). But this is not the place to document it or signal it via an interface. No user will see or use those functions directly. I will let you do a review of the user documentation on the logging when I wrote it and then you can decide if it is clear enough.

estimagic/logging/create_database.py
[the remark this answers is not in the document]
hmgaudecker: Pounders.

It uses an internal model for the objective function and may do several iterations until it evaluates the gradient again.

estimagic/logging/create_database.py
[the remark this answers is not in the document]
hmgaudecker: What is a Boolean table?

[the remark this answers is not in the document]
janosg: I meant a table with one boolean column.

estimagic/logging/create_database.py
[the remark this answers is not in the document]
janosg: Ok, I will have to look into that. But I will merge now and handle the rest when we actually implement the gradient logging.

estimagic/logging/create_database.py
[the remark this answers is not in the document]
hmgaudecker: You do not need that then. Just store the indices of the champions and do an inner merge.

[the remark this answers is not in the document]
janosg: Smart!

PR #101, opened 2019-11-26
implement logging

janosg:
## Goal

Add a basic version of persistent logging. For now we will only save the criterion and parameter history. Getting the gradients and convergence criteria out of the different algorithms will be done in a separate PR. Using the database in the dashboard will also be done in a separate PR. 

## Tasks

- [x] Fix an interface and write documentation
- [ ] Create all databases (in the case of parallel optimizations) before the first optimization or the dashboard actually start. 
- [ ] Update the database in `internal_criterion`
- [ ] Replace the counter in the criterion_function by something thread safe based on the database.


tobiasraabe:
This PR is closed in favor of #102.

Issue #56, opened 2019-08-29
Allow for user specified derivative functions during optimization


janosg:
## Current situation

We do pass a user specified gradient or hessian functions through to to the optimizer, but this function will be called with an internal parameter vector.

## Desired situation

A user should not have to bother about the internals and be able to provide a gradient or hessian functions that work on an external parameter vector. Estimagic converts this automatically into a corresponding function that operates on an internal parameter vector.

## Implementation

We will have to write closed form derivative functions for the reparametrization steps (from_internal) and use the chain rule. Everyone who has seen me taking derivatives knows that I will need help with the math. The most difficult one should be the cholesky transformation, but below I post some references. 

## Some references

- [Theory](https://arxiv.org/abs/1602.07527)
- Of course there is a [python implementation](https://github.com/imurray/chol-rev)

hmgaudecker:
Can't you just--for most of the stuff--just run the constraints functions through sympy?

janosg:
I have never worked with sympy but will check it out.

janosg:
This is done in estimagic 0.1. Unfortunately we had to go for closed form implementations because jax does not run on windows and sympy could not do what we need.

PR #102, opened 2019-11-27
Add decorators for conversions and logging.

tobiasraabe:
This PR adds logging via databases to estimagic.

### Features

- Variable handling of paths to databases.
- Logging for parallel estimations.
- Logging by default. Database is called ``"logging.db"``. It can be disabled by passing something which evaluates to ``False``.
- Logging is disabled in tests via monkeypatch or explicitly as monkeypatches are not recognized in optimizations spawned with multiprocessing.
- Logging gradients and the gradient status. Numbering in case it is not clear how many calculations are made.
- If a database is present, all tables are dropped and the new information is written into the clean tables.
- A default numerical gradient is passed to scipy and pygmo optimizers.
- The ``results`` dictionary is a partially harmonized output for all optimizers.

### Todo

- [x] Reconsider names of decorators and new constructs in general.


janosg:
For derivatives we also need conversion from numpy to DataFrame but without repararametrization step.

janosg:
Just realized, this is already handled in your solution.

estimagic/optimization/optimize.py
```diff
@@ -33,40 +37,54 @@ def maximize(
...
     general_options=None,
     algo_options=None,
+    logfile=DEFAULT_DATABASE_NAME,
+    log_options=None,
     dashboard=False,
     db_options=None,
 ):
-    """
-    Maximize *criterion* using *algorithm* subject to *pc* and bounds.
+    """Maximize *criterion* using *algorithm* subject to *pc* and bounds.
```
janosg: This was a pycharm mistake. *pc* should be *constraints*

tobiasraabe:
I think that's it. The success field of the results dictionary is a little bit silly as it always returns success if the optimization completes.

tobiasraabe:
I tried to cross-reference all issues which should be addressed by this PR. @janosg maybe you can go through them and add them with closes to the PR description.

estimagic/decorators.py
```diff
@@ -0,0 +1,140 @@
...
+
+import numpy as np
+
+from estimagic.config import DEFAULT_CRITERION_PENALTY
+from estimagic.logging.update_database import append_rows
+from estimagic.logging.update_database import update_scalar_field
+from estimagic.optimization.reparametrize import reparametrize_from_internal
+
+
+def x_to_params(params, constraints=None):
```
janosg: I think this decorator should get one additional optional argument called `output_type` or `out_type`. If this is `pd.Series` or `pd.DataFrame` the result is converted to an array. 

Also, how about calling it `numpy_interface`? I think it looks intuitive to read something like:
```python
@numpy_interface(params, constraints, out_type=pd.Series)
def my_pandas_criterion_function(params):
    pass
```
This would also make it clearer that the decorator is something that operates on a function.

estimagic/decorators.py
```diff
@@ -0,0 +1,140 @@
...
+            out = func(p, *args, **kwargs)
+
+            return out
+
+        return wrapper_x_to_params
+
+    return decorator_x_to_params
+
+
+def log_parameters_and_criterion_value(database, tables):
```
janosg: I would call it `log_criterion_evaluation` or just `log_evaluation`. It should be clear enough that this logs the inputs and the output.

estimagic/decorators.py
```diff
@@ -0,0 +1,140 @@
...
+    def decorator_exception_handling(func):
+        @functools.wraps(func)
+        def wrapper_exception_handling(x, *args, **kwargs):
+            try:
+                out = func(x, *args, **kwargs)
+            except (KeyboardInterrupt, SystemExit):
+                raise
+            except Exception as e:
+                constant, slope = general_options.get(
+                    "criterion_exception_penalty", (np.nan, np.nan)
```
janosg: I would prefer to set the default penalty constant to something like 2 * criterion(start_params) which we can save in the copy of general_options after the initial function evaluation. The default slope could be `0.1 * criterion(start_params)` or some other relatively small fraction of the start function value. 

I think reasonable defaults that do not depend on the order of magnitude of the initial function evaluation are impossible. I also think that the slope should not be zero by default (which it currently is) because users would never think about switching it on.

estimagic/decorators.py
```diff
@@ -0,0 +1,140 @@
...
+                )
+                raise_exc = general_options.get("criterion_exception_raise", False)
+
+                if raise_exc:
+                    raise e
+                else:
+                    out = min(
+                        DEFAULT_CRITERION_PENALTY,
```
janosg: If you agree with my proposed changes, this should be called `MAX_CRITERION_PENALTY`

estimagic/differentiation/differentiation.py
```diff
@@ -195,13 +196,10 @@ def _no_extrapolation_hessian(internal_func, params_value, method):
...
 def _create_internal_func(func, params, func_kwargs):
-    def internal_func(x):
-        p = params.copy(deep=True)
-        p["value"] = x
-
+    @x_to_params(params)
+    def internal_func(p):
         func_value = func(p, **func_kwargs)
-        if isinstance(func_value, (pd.DataFrame, pd.Series)):
-            func_value = func_value.to_numpy()
```
janosg: I think this line should not have been deleted, but I proposed a change to `x_to_params` that would make it unnecessary.

estimagic/optimization/optimize.py
```diff
@@ -34,51 +45,71 @@ def maximize(
...
-            additional configurations for the optimization.
+            additional configurations for the optimization
 
-        algo_options (dict):
-            algorithm specific configurations for the optimization.
+        algo_options (dict or list of dicts):
+            algorithm specific configurations for the optimization
+
+        gradient (callable or None):
+            Gradient function.
```
janosg: If I understand `create_internal_gradient` correctly, `gradient` has to be a function that calculates a gradient numerically, given a function. But then it is not necessary that the user can change this function because in a few weeks we will just get a really good numerical gradient function. 

Thus I would suggest that for now we remove the gradient argument from the user interface and only keep gradient options (which contains arguments that influence the calculation of the numerical gradient). Then we always do what you currently do if `gradient=None`.

In the long run we can add a gradient argument, but this would then not be a numerical derivative, but a closed form gradient. However, to make this work, we need to solve #56 which should be relatively easy since we re-wrote the reparametrizations.

estimagic/optimization/optimize.py
```diff
@@ -318,13 +417,36 @@ def _internal_minimize(
             Set to -1 for maximizations to plot the fitness that is being maximized.
 
     """
+    logging_decorator = functools.partial(
```
janosg: In my next PR I will replace the `queue` argument of `create_internal_criterion` by the `database`. Thus I suggest you already add `database` as argument now. Then we don't need `logging_decorator` as argument because we have all available information to construct it inside `create_internal_criterion`.

estimagic/optimization/optimize.py
```diff
@@ -344,51 +466,67 @@ def _internal_minimize(
...
+        results = minimize_scipy_np(
+            internal_criterion,
+            internal_params,
+            bounds=bounds,
+            algo_name=algo_name,
+            algo_options=algo_options,
+            gradient=internal_gradient,
         )
+    elif origin == "tao":
+        crit_val = general_options["criterion"]
```
janosg: I think it would be better to call this "start_criterion_value" to be very explicit. Just `criterion` usually refers to the function in estimagic, not to a specific function value.

estimagic/optimization/optimize.py
tobiasraabe: I have implemented ``logging_decorator`` because the ``internal_criterion`` for the function evaluations receives a different decorator than the ``internal_criterion`` for the gradient.

estimagic/optimization/optimize.py
tobiasraabe: I changed it such that ``process_optimizations_arguments`` always passes ``gradient = None``. Thus, the keyword is already present on lower levels, but hidden from the interface. Is this what you suggested?

estimagic/decorators.py
tobiasraabe: I am not sure whether out_type is really necessary.

1. ``out_type`` is not very intuitive as I would expect that the return after decorating is a pandas series.
2. numpy interface suggests by its name that the return is a numpy array, thus it is clear to which object we want to transform.
3. Every ``out_type`` needs its own conversion to NumPy anyway. So no need to pass it in. We can just check inside the function.

But, probably you have a use-case in mind which I do not.

estimagic/decorators.py
tobiasraabe: Went with the second.

estimagic/decorators.py
janosg: No, you are right. Checking the type is better than the argument. It is just important that the output is converted to numpy if it is a pandas object and If I'm correct this is currently not done.

estimagic/optimization/optimize.py
janosg: Yes, exactly. Thanks.

estimagic/optimization/optimize.py
janosg: I understand. That's a good solution.

PR #106, opened 2019-12-09
Fix Small Bugs

janosg:
- The dashboard scaling did not work for models with many parameters
- Making deepcopies of the `criterion_kwargs` is not compatible with skillmodels. I think it is ok to just remove that. 
- The counter did not work because we used a float instead of a length one array.


janosg:
We use `np.linspace` with an array for `start` and `stop` and according to the documentation this is only supported since `0.16.0`.

tobiasraabe:
@janosg Ready for merge.

PR #107, opened 2019-12-10
Implement interface for likelihood estimation.

tobiasraabe:
It was intended that this PR only adds a ``estimate_likelihood`` function. We could not hold onto ourselves.

### Features

- ``estimate_likelihood`` function which needs a criterion function which returns log like obs instead of a scalar.
- There is a decorator to wrap any criterion function and turn the first return to a scalar which is necessary for likelihood estimation and msm using a scalar optimizer.
- Every criterion function is allowed to have two returns
  1. criterion value (or array matrix whatever which is converted via a decorator within likelihood and msm esitmation)
  2. ``comparison_plot_data`` which can be a NumPy array, Series or DataFrame.
- ``"bounds_distance"`` is documented in the notebook explaining constraints.
- New table in database called ``comparison_plot`` which holds the data used for the comparison plot in the dashboard alongside the parameter frame. The first entry contains various invariant information (index, other cols) whereas subsequent entries store only the value column.
- The fitness in the results dictionary corresponds to the log likelihood of the sample.
- Maximize is now able to be parallelized.
- Since decorated functions cannot be used with ``multiprocessing``, ``mp.Pool`` is replaced with ``joblib.Parallel``. It does the same job but uses ``cloudpickle`` instead of ``pickle`` to serialize objects and pass them to the newly spawned processes. This works with decorated functions and all types of arguments.

### Todo

- [ ] In ``internal_criterion`` there is still the hotfix for POUNDERs. The reason is that the criterion function returns the errors, but we need a scalar. We cannot wrap the criterion with the ``aggregate_criterion_output`` because then POUNDERs does not see the errors anymore. In the long run, we need a aggregation function within ``log_evaluation`` decorator which stores the criterion value in the database. I would postpone it until we completely move to the database.

  I do not think that logging POUNDERs has ever worked because I missed this case. We need to test the whole interface with every optimizer. The NumPy interface is just not enough.


estimagic/estimation/estimate.py
```diff
@@ -0,0 +1,51 @@
...
+        general_options,
+        algo_options,
+        gradient_options,
+        logging,
+        log_options,
+        dashboard,
+        db_options,
+    )
+
+    return results
```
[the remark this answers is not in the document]
tobiasraabe: What do you mean by adjust?

estimagic/logging/create_database.py
```diff
@@ -214,9 +234,21 @@ def _define_convergence_history_table(database):
...
 
 def _define_start_params_table(database):
-    params_table = Table(
+    start_params_table = Table(
         "start_params", database, Column("value", PickleType), extend_existing=True
     )
+    return start_params_table
+
+
+def _define_table_formatted_with_pickled_scalar(database, table):
```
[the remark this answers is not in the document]
tobiasraabe: Done

estimagic/decorators.py
```diff
@@ -48,22 +48,64 @@ def wrapper_numpy_interface(x, *args, **kwargs):
...
-            criterion_value = func(params, *args, **kwargs)
+        def wrapper_log_evaluation(params, *args, **kwargs):
+            out = func(params, *args, **kwargs)
+
+            if np.isscalar(out):
+                criterion_value = out
+                cp_data = {"value": np.array([np.nan])}
+            else:
+                criterion_value, comparison_plot_data = out
+                if isinstance(comparison_plot_data, np.ndarray):
```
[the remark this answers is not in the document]
tobiasraabe: Done.

estimagic/estimation/estimate.py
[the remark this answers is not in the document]
janosg: I mean we have to multiply it with the number of elements over which we took the average. So we get the log likelihood and not the average log likelihood contribution.

docs/source/getting_started/estimate_likelihood.ipynb
```diff
@@ -0,0 +1,324 @@
...
+    "form = \"apply ~ pared + public + gpa\""
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": 5,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "info, params = ordered_logit(form, df, dashboard=False)"
```
roecla: This does not run on my machine. The error is a KeyError in `estimate_likelihood` in line 81. `result` is already a dictionary with an entry "fitness".

estimagic/decorators.py
```diff
@@ -47,23 +47,78 @@ def wrapper_numpy_interface(x, *args, **kwargs):
...
-            criterion_value = func(params, *args, **kwargs)
+        def wrapper_log_evaluation(params, *args, **kwargs):
+            out = func(params, *args, **kwargs)
+
+            if np.isscalar(out):
+                criterion_value = out
+                cp_data = {"value": pd.DataFrame({"value": [np.nan]})}
+            else:
+                criterion_value, comparison_plot_data = out
+                cp_data = {"value": comparison_plot_data["value"].to_numpy()}
```
roecla: You handle several times the "expansion" from one to two outputs (e.g. in also in lines 56-61). Why not have one `expand_output` decorator and "concatenate"? This might also allow us to avoid case handling elsewhere (e.g. in `optimize` lines 310-320?).

estimagic/estimation/estimate.py
```diff
@@ -0,0 +1,83 @@
...
+    constraints=None,
+    general_options=None,
+    algo_options=None,
+    gradient_options=None,
+    logging=DEFAULT_DATABASE_NAME,
+    log_options=None,
+    dashboard=False,
+    db_options=None,
+):
+    """Estimate parameters via likelihood.
```
roecla: I think we should call `estimate_likelihood` `maximize_log_likelihood` instead. That makes it immediately obvious that we are maximizing. If we decide to keep the name as is, I would clarify "via maximizing the likelihood" in the first line of the docstring.

estimagic/estimation/estimate.py
```diff
@@ -0,0 +1,83 @@
+import numpy as np
+
+from estimagic.config import DEFAULT_DATABASE_NAME
+from estimagic.decorators import aggregate_criterion_output
+from estimagic.optimization.optimize import maximize
+from estimagic.optimization.process_arguments import process_optimization_arguments
+
+
+def estimate_likelihood(
+    criterion,
```
roecla: I think it is confusing to call the first argument of `criterion`. What the user passes here is not what the user would pass to `maximize` if she wanted to do the same thing. I would suggest `criterion_contributions` or `contribution_level_criterion`.

estimagic/estimation/estimate.py
tobiasraabe: Done

estimagic/decorators.py
tobiasraabe: Done.

docs/source/getting_started/estimate_likelihood.ipynb
tobiasraabe: Fixed.

estimagic/estimation/estimate.py
tobiasraabe: I changed it to ``loglikeobs``. Maybe we want to find a more general name in the future when there is a GMM interface.

hmgaudecker: I think it definitely should have a plural in there. And then a clear docstring stating that this is supposed to be an array of log-likelihood contributions (I am mostly guessing here). `criterions_per_obs` ? That should work for GMM & friends, too.

You probably discussed that elsewhere, but is there a reason for `logging` & `log_options` instead of `logging_options`? Not that I would think there should be many options for taking logarithms, but expanding `log` to `logarithm` is just hard-wired in my brain when a likelihood is close by. But maybe I am just getting too old...

tobiasraabe: I think we chose ``log_options`` instead of ``logging_options`` for brevity. But, this is probably not strictly necessary and not a good reason if it causes confusion.

I moved the rest of the discussion to #112 as I would like to have this merged and Janos should join the discussion.

PR #109, opened 2019-12-14
Log exceptions.

tobiasraabe:
We catch exceptions in the internal criterion function, but do not store the traceback. This makes debugging an optimization a little bit easier, because failures can be due to the optimizer or the model.

It helped to debug this case: https://github.com/OpenSourceEconomics/respy/issues/305.

Relies on #106.

### Features

- Exceptions in the internal criterion function are stored in the database.
- Exceptions while updating the database are printed to the console.


tobiasraabe:
@janosg Docs will be fixed in #111. Thus, ready for review.

PR #111, opened 2019-12-17
Fix memory consumption error on rtd.

tobiasraabe:
The conda environment takes more memory than the 2GB provided by rtd and thus, the build sometimes crashes. The changes make the documentation more static by mocking imports and using the notebooks as they are and reducing packages in the environment.

### Features

- The documentation has its own, smaller environment. Imports of NumPy, etc. are mocked.
- sphinx check are now executed on Linux
- Warnings while building docs are displayed but are not recognized as errors. Needs to be resolved at some other point in time.

### DIFF
--- a/.readthedocs.yml
+++ b/.readthedocs.yml
@@ -1,0 +1,10 @@
+version: 2
+
+build:
+  image: latest
+
+python:
+  version: 3.7
+
+conda:
+  environment: docs/environment.yml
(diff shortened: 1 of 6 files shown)


tobiasraabe:
I was not quite at the point where I would have reused commit messages over and over again :). Yes, we need to commit notebooks with output and we have to revisit them from time to time just to make sure they are still working.

Maybe one should write a script for that and execute every notebook before a release.

PR #124, opened 2020-02-12
Fix bounds

janosg:
## Bug Description

the processed params DataFrame containts two types of columns for bounds:

1. `lower` and `upper`, which are what the user provides
2. `_internal_lower` and `_internal_upper`, which in addition contain bounds that are implied by other constraints, e.g. 0 lower bounds for diagonal elements of a covariance matrix. 

The internal bounds should be passed to the optimizers. Since estimagic `0.0.23` we passed the user provided ones. 

## Solution

Pass the internal bounds

## Remarks

While debugging this, we found that constraints are not yet saved in the database. This is changed now.

@hmgaudecker, this should solve your problems in skillmodels
@mo2561057, this should solve your problem in respy


estimagic/optimization/optimize.py
```diff
@@ -492,7 +492,11 @@ def _internal_minimize(
             f"{algorithm} is not a valid choice. Did you mean one of {proposals}?"
         )
 
-    bounds = tuple(params.query("_internal_free")[["lower", "upper"]].to_numpy().T)
+    bounds = tuple(
```
tobiasraabe: Maybe a function for this thing because it is done two times and if we had not forget that it exists in two places we might have fixed only one call?

mo2561057: That makes sense to me since I think it contains exactly the same logic across the two examples!

janosg: Done. Thanks!

PR #130, opened 2020-02-15
Fix error with start_criterion_value and POUNDERs.

### LINKED ISSUE #129: Output length
https://github.com/OpenSourceEconomics/estimagic/blob/a9523fd730b32dacbb838f2bfe24ab2c01946685/estimagic/optimization/optimize.py#L522
This will always return 1 for n_errors leading to a broadcasting error downstream.
We should maybe also adapt testing pounders to detect such issues faster.

mo2561057:
Fixes #129 
We used a scalar object to pass the length of our output to estimagic. 
I delted that check for now pounders will evaluate twice before it starts.


tobiasraabe:
We need ``"start_criterion_value"`` in case the criterion function raises an exception (e.g., `LinAlgError` due to extreme parameters) to give back a penalized criterion value to the optimizer. You can see this in `_handle_exceptions`.

What else do you want to change? What is the issue with the current code?

mo2561057:
Sorry forgot to reference the issue and the name doesnt make sense.
Until now we always passed 1 as output length to pounders!

janosg:
I don't understand why the code you deleted always passed 1 as output length to pounders. To me it looks correct. Can you explain more?

mo2561057:
https://github.com/OpenSourceEconomics/estimagic/blob/a9523fd730b32dacbb838f2bfe24ab2c01946685/estimagic/optimization/optimize.py#L322-L340
general_options["start_criterion_value"]  will always be a scalar since we already take the square mean before but then we use it to get the length of the output of the criterion for pounders.

tobiasraabe:
You are right. We should probably save criterion output and the fitness factor and adjust `handle_exception` to work with an array and a scalar.

janosg:
@mo2561057 thanks for the example. @tobiasraabe can you make the change you proposed? Since you wrote the code that has to be changed it would be too much work for Moritz. I can do the review.

tobiasraabe:
I'll do it.

The only thing which bothers me is that if we get a vector as criterion output we assume the correct scalar is the mean of squared terms. We should get rid of it, but I do not know how.

tobiasraabe:
@mo2561057 I hope this works for you.

janosg:
> The only thing which bothers me is that if we get a vector as criterion output we assume the correct scalar is the mean of squared terms. We should get rid of it, but I do not know how.

I think in the estimagic interface we should not adhere that strictly to the old tao pounders interface. Multi objective functions should always be such that the sum of the entries is the criterion. Then we write another decorator that converts this into a pounders compatible function. This would be more intuitive. I think that is also what pygmo does and how BHHH would be implemented.

janosg:
@mo2561057 @amageh could one of you please resolve the conflicts and write a test that makes the coverage checks pass?

Issue #143, opened 2020-03-02
Make estimagic work with pandas 1.0


janosg:
### Current situation

Estimagic is only tested for pandas <= 0.25

### Desired Situation

Estimagic works for pandas 1.0 and older versions. 

### To-Do

I'm not sure which steps are necessary, but at least we need the following:

- [ ] Some pickled DataFrames cannot be loaded with the new pandas. We have to load them with an older pandas and save as something that is not .pickle. Then load it with the new pandas and pickle them. 
- [ ] Upgrade the test environments in tox.ini to use pandas 1.0 on all operating systems
- [ ] Add a new environment that uses an older pandas version. I think it's enough to do this only on linux, where tests much quicker than on the other systems. 

@tobiasraabe, @roecla @mo2561057 @amageh @ChristianZimpelmann Please add to-dos if you run into problems when upgrading pandas in your projects and think we have to do something in estimagic.

PR #144, opened 2020-03-02
Small Fixes

janosg:
- Restrict pandas to >=0.24 and <1.0 in meta.yaml and other files. Otherwise conda build fails. See #143 
- Add enhancement issue template from respy

### DIFF
--- a/meta.yaml
+++ b/meta.yaml
@@ -20,7 +20,7 @@ requirements:
     - python {{ python }}
     - joblib
     - numpy>=1.16.0
-    - pandas>=0.24
+    - pandas<=0.25
     - pygmo>=2.10
     - bokeh>=1.3
     - scipy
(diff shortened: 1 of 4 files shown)


meta.yaml
```diff
@@ -20,7 +20,7 @@ requirements:
     - python {{ python }}
     - joblib
     - numpy>=1.16.0
-    - pandas>=0.24
+    - pandas<=0.25
```
tobiasraabe: Same for environment.yml and tox? Actually, it would be great if you can fix the issues with pandas v1 very quickly.

janosg: Done, thanks. I'll try to do it soon (or even better, find someone who wants to do it).

hmgaudecker: Nit-picky here, but don't you want
```
- pandas>=0.24,<1.0
```
here to be on the safe side? https://docs.conda.io/projects/conda-build/en/latest/resources/package-spec.html#package-match-specifications

janosg: Yes. Just didn't know that this works ;-)

PR #149, opened 2020-03-07
Implement local moment sensitivity measures

loikein:
Hi Professor, this is for the Effective Programming course, final project.

Some comments from me:

1. `pytest` does not pass, because the calculated precision is too high, and precision adjustment for `pandas` does not work. I tried to update to version 1.0, still does not work, but you may want to consider updating anyway, since in the environment it's currently 0.24 or something.
1. When `importing matplotlib`, the estimagic environment does not have `matplotlib`, and after adding it still cannot not find, which is beyond my knowledge. Therefore, the plotting part of the demo is done in my virtual environment.

Cheers!

~Leiqiong Wan


EconBeast:
Hey,
1) Why not use numpy methods such as np.allclose, np.testing.assert_almost_equal? I know in numpy there are methods in which you may specify decimal places.

2) Try installing matplotlib to your base environment using the anaconda prompt. I use spyder and this is always the case for me - even when I install through the powershell on the base environment.

janosg:
Hi @loikein, I finally have some time to look at this. Looks all very good. I already addressed a few of the above comments and will do some further polishing in the next days. 

To-Do:
- [ ] Split the test into one per sensitivity measure
- [ ] Fix deprecation warning with `check_less_precise` argument
- [ ] Make sure plotting in notebook works in the environment (see comment above)
- [ ] Add `numdiff_kwargs` (check how they are implemented in optimization or inference)

estimagic/inference/likelihood_covs.py
```diff
@@ -17,60 +17,60 @@ def cov_hessian(hessian):
...
-    info_matrix = np.dot((jacobian.T), jacobian)
+    info_matrix = np.dot((jac.T), jac)
     cov_jac = robust_inverse(info_matrix, msg=INVALID_INFERENCE_MSG)
 
     return cov_jac
 
 
-def cov_sandwich(jacobian, hessian):
+def cov_sandwich(jac, hess):
     """Covariance of parameters based on HJJH dot product.
```
mo2561057, line 55: This is the most robust version, isn't it? Would it make sense to rename this to cov_robust such that users see that it can make sense to use this method? 
Also once someone has the Hessian it is basically for free to calculate different versions of the standard errors. Is there any value in comparing these different standard errors?

estimagic/inference/msm_covs.py
```diff
@@ -0,0 +1,68 @@
...
+            of simulate_moments with respect to params. The derivative needs to be taken
+            at the estimated parameters. Has shape n_moments, n_params.
+        weights (np.ndarray): The weighting matrix for msm estimation.
+        moments_cov (np.ndarray): The covariance matrix of the empirical moments.
+
+    Returns:
+        numpy.ndarray: numpy array with covariance matrix.
+
+    """
+    _jac, _weights, names = process_pandas_arguments(jac=jac, weights=weights)
```
mo2561057, line 61: Would it make sense to delegate these steps to a decorator such that this function is purely NumPy-based? Or do you think that performance never plays a role here? Would also save some space and make functions easier maybe?

estimagic/inference/shared.py
```diff
@@ -43,20 +47,21 @@ def transform_covariance(
     processed_constraints, processed_params = process_constraints(constraints, params)
     free_index = processed_params.query("_internal_free").index
 
+    if isinstance(internal_cov, pd.DataFrame):
```
mo2561057, line 50: Similar here. I would be wondering whether it is nicer to handle these issues outside?

estimagic/tests/sensitivity/test_msm_sensitivity.py
```diff
@@ -0,0 +1,209 @@
...
+from estimagic.sensitivity.msm_sensitivity import (
+    calculate_fundamental_sensitivity_to_noise,
+)
+from estimagic.sensitivity.msm_sensitivity import (
+    calculate_fundamental_sensitivity_to_removal,
+)
+from estimagic.sensitivity.msm_sensitivity import calculate_sensitivity_to_bias
+from estimagic.sensitivity.msm_sensitivity import calculate_sensitivity_to_weighting
+
+
```
mo2561057, line 23: Maybe we could add a test that checks some random edge cases. For example when all the variance is one moment only etc?

mo2561057 (approved):
Looks great! 
I did not see many issues. 
In general, it might make sense to think about unifying MSM standard errors and likelihood standard errors in one function and exploit special cases that allow easier standard errors in the medium run. Then we could allow estimation of standard errors for all M or extremum estimators. 
In the more advanced tutorial, we should add something about generalized indirect inference.

estimagic/inference/likelihood_covs.py
janosg: For MSM this is really the only one of practical relevance since nobody does optimal MSM. In the optimal case the two different versions are identical (up to floating point imprecision), in any other case using the optimal formula is invalid. But I like the idea of calling it `cov_robust`.

estimagic/inference/msm_covs.py
janosg: Another way would be to put the type handling in the function that calculates all sensitivity measures and to make the individual functions numpy only and private. That would be most efficient and most pretty. I didn't do it because I wanted the individual functions public but I think I changed my mind.

estimagic/inference/shared.py
janosg: In this case it would be very hard.

estimagic/tests/sensitivity/test_msm_sensitivity.py
janosg: Would postpone that for now.

estimagic/inference/msm_covs.py
janosg: Will also postpone this until RAs reimplement the sensitivity measures. Adjusting the tests would take too long.

PR #150, opened 2020-03-08
Running the dashboard becomes a CLI command.

### LINKED ISSUE #142: Command Line Interface to `run_dashboard`
### Current situation

Currently the dashboard can only be started by `maximize` or `minimize` (or with a python script that uses undocumented functions). Thus it is hard to start a dashboard for an optimization that was already started earlier. 

### Desired Situation

The propose way of starting a dashboard is from the command line. 

### Proposed implementation

Use click to achieve the following command:

`estimagic dashboard db1.db logging_db`

Which works for any number of databases. Moreover, the command should have the optional arguments:
--port 1234
--no_browser

This should be really easy because [`run_dashboard`](https://github.com/OpenSourceEconomics/estimagic/blob/f1b8525c2d336b6fd2b88e61a720b5970a9663f4/estimagic/dashboard/run_dashboard.py#L48) already exists and does exactly what we need. 

The documentation has to be adjusted accordingly.

tobiasraabe:
### What do you want to achieve?

Resolves #142.

Under https://anaconda.org/brimborium/estimagic, you can find a preliminary release with the dashboard command.

After installation, checkout the CLI by typing ``estimagic`` and ``estimagic dashboard -h``.

*Questions*

- Should it be possible to start the dashboard without databases? Probably?


roecla:
Yes, it is planned that the dashboard will support that the user can supply paths to databases that do not exist yet and waits for them to be created.

tobiasraabe:
Great comment! And should the dashboard start even if no paths are supplied? Do you plan to add a button to the dashboard to locate dbs on the disk?

roecla:
I think if no path is supplied to the dashboard, there should be an error. I don't think it makes sense to allow users to add paths ex-post (via a button or otherwise). If the user wants add one or several paths, he or she should just close the current dashboard and open a new dashboard with the old and new paths. 

What we might consider is allowing users to supply a directory path and then automatically open all databases in all subfolders of that directory. That would probably add a lot of comfort for things like boostrap runs. 

Regarding the button: No, I would not let the user "manually" check for the databases. 
Instead if databases are missing we put them in the "scheduled" tab of the overview page. The tabs are not implemented yet but this is what we're planning it to look like:

image

Then, we keep the opened tab updated to display the optimizations that are in that stage. Thus, once the optimization starts, its link is moved to the "running" section and the user can normally open the convergence plot etc. because by then the database will exist.

janosg:
Thanks for the PR. A few comments:

1. Adding new dashboards dynamically is not possible with a bokeh server. The underlying tornado webserver seems to support it and the bokeh developers are open to a PR. However, I think it's not worth the trouble, because it is so easy to just start a new server. 
2. The idea to specify a file and automatically get a dashboard with all `.db` files in that directory is great. In addition, we could also support [glob patterns](https://docs.python.org/3/library/glob.html) because specifying the path list is the most annoying part of running a dashboard.

PR #151, opened 2020-03-17
Re-structure packaging.

janosg:
## Current situation

- Package size is relatively large
- We have a custom build script (release.py)
- Need different packages for different operating systems and python versions

## Desired Situation

- Use conda commands instead of build script
- Have a noarch package

## Solution 

See corresponding [respy PR](https://github.com/OpenSourceEconomics/respy/pull/340)


CITATION
```diff
@@ -0,0 +1,20 @@
...
+
+@Unpublished{Gabler2019,
+      Title  = {A Python Tool for the Estimation of (Structural) Econometric Models.},
+      Author = {Janos Gabler},
+      Year   = {2019},
+      Url    = {https://github.com/OpenSourceEconomics/estimagic}
+    }
+
+If you are unsure about which version of respy you are using run: `pip show
+respy`.
```
tobiasraabe:
```suggestion
If you are unsure about which version of estimagic you are using run: `conda list estimagic`.
```

PR #158, opened 2020-04-21
Allow for user provided gradient in `maximize` and `minimize`

janosg:
## Current Situation

User cannot specify a closed form gradient

## New Situation

If no constraints are used, the user can provide a closed form gradient that takes `params` as the only arguments and returns a numpy array or pandas Series with the gradient. In the long run we also want to support custom gradients for constrained optimizations, but this is too much work for now.


estimagic/decorators.py
```diff
@@ -114,6 +114,19 @@ def wrapper_negative_criterion(*args, **kwargs):
     return wrapper_negative_criterion
 
 
+def negative_gradient(gradient):
```
tobiasraabe, line 117: This looks awkward. Why dont you use functools.wraps as an decorator and put the returns below the decorated function. And why dont you return the wrapped function?

PR #159, opened 2020-04-30
Add first version of Richardson extrapolations.

timmens:
To increase the accuracy of numerically estimated sequences we employ the method of Richardson extrapolations, see https://en.wikipedia.org/wiki/Richardson_extrapolation.

The code is largely copied from ``numdifftools`` and transformed into a functional programming style.


estimagic/differentiation/numdiff_np.py
```diff
@@ -166,12 +180,162 @@ def _consolidate_one_step_derivatives(candidates, preference_order):
...
+        richardson_candidates (OrderedDict): Dictionary with derivative estimates and
+            error estimates from different methods.
+    """
+    richardson_candidates = OrderedDict()
+    for method in ["forward", "backward", "central"]:
+        for num_terms in range(1, n_steps):
+            derivative, error = richardson_extrapolation(
+                jac_candidates[method], steps, method, num_terms
+            )
+            richardson_candidates[method + str(num_terms)] = {
```
janosg, line 243: I have the feeling that an OrderedDict is not the best data structure to represent this. Since the keys are a composite of the metho and `str(num_terms)`, wouldn't it be better to have a dictionary of list, where the keys are just the methods and the values are lists with one entry per num_terms. The drawback is that num_terms starts at one and list indices at zero, but I still find it better. 

If not, we still don't need an explicit ordered dict for what you are doing. Dicts are ordered in cpython since 3.6 and in all other python implementations since 3.7. I think the estimagic user-base that uses a fancy 3.6 Python implementation is small enough to neglect them.

estimagic/differentiation/numdiff_np.py
```diff
@@ -166,12 +180,162 @@ def _consolidate_one_step_derivatives(candidates, preference_order):
...
+        steps (namedtuple): Namedtuple with the field names pos and neg. Each field
+            contains a numpy array of shape (n_steps, len(x)) with the steps in
+            the corresponding direction. The steps are always symmetric, in the sense
+            that steps.neg[i, j] = - steps.pos[i, j] unless one of them is NaN.
+        n_steps (int): Number of steps needed. For central methods, this is
+            the number of steps per direction. It is 1 if no Richardson extrapolation
+            is used.
+
+    Returns:
+        richardson_candidates (OrderedDict): Dictionary with derivative estimates and
```
janosg, line 234: Please describe keys and values in more detail.

estimagic/differentiation/numdiff_np.py
```diff
@@ -166,12 +180,162 @@ def _consolidate_one_step_derivatives(candidates, preference_order):
...
+        jac_minimal = np.squeeze(derivatives, axis=0)
+    else:
+        try:
+            minimizer = np.nanargmin(errors, axis=0)
+
+            jac_minimal = np.take_along_axis(
+                derivatives, minimizer[np.newaxis, :], axis=0
+            )
+            jac_minimal = np.squeeze(jac_minimal, axis=0)
+        except ValueError:
```
janosg, line 332: Why only except ValueErrors? Are they frequent here? What causes them?

estimagic/differentiation/richardson_extrapolation.py
```diff
@@ -0,0 +1,275 @@
...
+        tol = np.maximum(np.abs(new_seq[1:]), np.abs(new_seq[:-1])) * EPS * fact
+        converged = err <= tol
+        abserr = err + np.where(
+            converged, tol * 10, abs(new_seq[:-1] - old_seq[-new_seq_len + 1 :]) * fact,
+        )
+
+    return abserr
+
+
+def _get_order_and_exponentiation_step(method):
```
janosg, line 195: I think this is my favorite function in your PR! It just takes makes something that I could not understand after hours of reading numdifftools and makes it really simple to understand. Only thing that could be improved: Explain what order and exponentiation step are in vey first paragraph. Currently you explain it in the Returns section, which makes sense, but is a bit late.

estimagic/differentiation/numdiff_np.py
timmens: The ``ValueError`` can be caused if there are all-NaN slices in the errors, when computing the nanargmin, see https://numpy.org/doc/1.18/reference/generated/numpy.nanargmin.html.

I don;t think this occurs often.

I'm not sure anymore if we need a try-except block, it seems less clean / informative than just letting np.nanargmin raise the error.

estimagic/differentiation/numdiff_np.py
janosg: Ok, you convinced me here.

estimagic/differentiation/numdiff_np.py
janosg: Yes, the nanargmin error message is good enough.

timmens:
I incorporated @janosg  comments in (6510317bb7d658).

I noticed that for the function ``_compute_richardson_candidates`` ([link](https://github.com/OpenSourceEconomics/estimagic/blob/6510317bb7d6580ee1739306176cda3f1cae7ee5/estimagic/differentiation/numdiff_np.py#L219)), I use nested dictionaries.
Maybe it would be better to use for either the inner or outer dictionary a NamedTuple instead.
But this could be resolved when rewriting the code so that the we get a pandas DataFrame containing the derivatives.

PR #164, opened 2020-07-26
Implement derivatives of parameter transformations using only numpy.

timmens:
Implement closed form derivatives for the reparametrizations such that we can convert user specified gradients and jacobians (with respect to the external parameters) to gradients and jacobians with respect to the internal parameters. 

This is necessary because JAX is not available on windows and can thus not become a dependency of estimagic. 

The derivatives were tested once against JAX and are tested against our numerical `first_derivative` in the test suite.


janosg:
@timmens, Thank you very much for this PR. Despite the complexity of the problem, the code is quite readable and the comprehensive tests and resources you mention in the docstring convince me that the result is correct. 

@roecla, we can merge as soon as you approve.

Issue #138, opened 2020-02-18
Arguments for linear inequality constraints unintuitive


ChristianZimpelmann:
### Bug description

If I specify a linear constraint like
`{"loc": "b", "type": "linear", "weights": 1 / 3, "lower": 3}`
I read in my head `b-values lower 3`. What is actually specified is however `b-values greater than 3`.

### Expected behavior

I suggest naming the argument "lower_bound" instead of "lower" and "upper_bound" instead of  "upper" to avoid the misunderstanding.

janosg:
The original motivation for this was that in params we also call the bound columns just `"lower"` and `"upper"`, but I see your point and think we should change it. The question is if we then also rename it in params. The current name could be confusing once we get a `"ci_lower"` and `"ci_upper"` column for confidence intervals.

hmgaudecker:
+1 on removing ambiguity at the expense of longer column identifiers, too.

ChristianZimpelmann:
I would change it in params, too.

janosg:
I would also change it in params, but accept both for a while and emit a DeprecationWarning if the short names are used. The right place for the warning is probably [here](https://github.com/OpenSourceEconomics/estimagic/blob/0b31adacf2e251d4a2f4ca2f50247191c6381226/estimagic/optimization/transform_problem.py#L273)

ChristianZimpelmann:
Just to be sure. I assume we would like to do the same change `lower` -> `lower_bound` for parameter bounds?

Issue #139, opened 2020-02-18
"upper" column in parameter df with new pandas integer extension type, produces error


ChristianZimpelmann:
### Bug description

If the "upper" column in a params df is of the new pandas integer extension type, `.to_numpy()` in `_internal_bounds_from_params` returns an array of type object which produces the error below.

### To Reproduce

```
data = pd.read_pickle("ologit.pickle")
formula = "apply ~ pared + public + gpa"
params, y, x, constr = ordered_logit_processing(formula, data)
params["lower"] = -20.
params["upper"] = 1

# Convert to new extension type
params = params.convert_dtypes()
res = maximize(
    criterion=ordered_logit_loglike, 
    params=params, 
    algorithm='scipy_L-BFGS-B',
    constraints=[],
    criterion_kwargs={"y": y, "x": x}, 
    dashboard=False
)
```

### Expected behavior

No error

### Screenshots/Error messages

```
~/miniconda3/envs/ambig-beliefs/lib/python3.7/site-packages/estimagic/optimization/scipy.py in minimize_scipy_np(func, x0, bounds, algo_name, algo_options, gradient)
     25     # and requires a list of tuples for each parameter with lower and upper bound.
     26     bounds = np.column_stack(bounds)
---> 27     mask = ~np.isfinite(bounds)
     28     bounds = bounds.astype("object")
     29     bounds[mask] = None

TypeError: ufunc 'isfinite' not supported for the input types, and the inputs could not be safely coerced to any supported types according to the casting rule ''safe''
```

### System

 - OS: [e.g. Ubuntu 18.04]
 - Version [e.g. 0.0.1]

ChristianZimpelmann:
This is probably an error that needs to be fixed by numpy.

janosg:
Would this be solved if we convert bounds to float, as we already do with the value column? If so, we should add this conversion [here](https://github.com/OpenSourceEconomics/estimagic/blob/1b2a8770d8c5cbbd4eea9088f7e7d07cea4a7d64/estimagic/optimization/optimize.py#L628) after Klara finished her current PR (which refactors the whole arguments processing for the optimizations). Converting the bounds to float is probably safer anyways for optimizers like pounders.

ChristianZimpelmann:
That should solve it. Let me know once Klara is finished, I can take care of it.

janosg:
I merge Klara's PR later today. Thanks for solving this.

PR #168, opened 2020-08-28
fix issues #138 and #139

ChristianZimpelmann:
fix issues #138 and #139

Therefore the constraints key words for lower and upper bounds are adjusted and the values of the bounds are converted to float.


janosg (approved):
Thanks a lot!

PR #169, opened 2020-08-31
Add additional dashboard arguments

roecla:
- [x] do not show parameters with None, Nan, empty string and False as groups
- [x] add jump argument to skip to the last *rollover* elements
- [x] add update frequency and update chunks as arguments 
- [x] allow switching between a logarithmic and a non-logarithmic scale.

Notes:
- bokeh's log scale only supports positive values. If values are negative they are not displayed.


estimagic/cli.py
```diff
@@ -31,14 +31,34 @@ def cli():
...
-def dashboard(database, port, no_browser, rollover):
+@click.option(
+    "--frequency",
+    default=1,
+    help="Number of seconds to wait between checking for new entries in the database.",
+    type=float,
+    show_default=True,
+)
+@click.option(
+    "--update_chunk",
```
janosg, line 55: Since `--no-browser` uses a dash instead of an underscore, we should do the same for `--update-chunk`

estimagic/dashboard/monitoring_app.py
```diff
@@ -197,110 +106,182 @@ def _map_groups_to_params(params):
...
-        callback_dict["plot_periodic_data"] = doc.add_periodic_callback(
-            plot_new_data, period_milliseconds=200
+            table_name="optimization_iterations",
+            n_rows=1,
+            return_type="list_of_dicts",
         )
-        # change the button color
-        button.button_type = "success"
-        button.label = "Reset Plot"
+        session_data["last_retrieved"] = max(0, last_entry[0]["rowid"] - rollover)
```
janosg, line 171: This does not exactly do what you describe for the `--jump` option. It does not directly generate a plot with the last rollover observations but jumps to the first of those observations and then updates with the standard update-chunk.

estimagic/logging/database_utilities.py
```diff
@@ -223,6 +223,25 @@ def read_new_rows(
...
     limit=None,
 ):
+    """Read all iterations after last_retrieved up to a limit.
+
+    Args:
+        database (sqlalchemy.MetaData)
+        table_name (str): name of the table to retrieve.
+        last_retrieved (int): The last iteration that was retrieved.
+        return_type (str): either "list_of_dicts" or "dict_of_lists".
+        limit (int): Only the first ``limit`` rows will be retrieved. Default None.
```
janosg, line 233: Limit is documented twice

estimagic/cli.py
roecla: Adressed in 9f3c5cbb9b9c3fe40548ed30da551737dfee5523

estimagic/dashboard/monitoring_app.py
roecla: Addressed in 9f3c5cbb9b9c3fe40548ed30da551737dfee5523.

estimagic/logging/database_utilities.py
roecla: Addressed in 9f3c5cbb9b9c3fe40548ed30da551737dfee5523

PR #173, opened 2020-09-09
Add color palettes from impact lab

janosg:
We already used parts of the color palettes in the dashboard, but now I add the full `get_colors` function.


roecla (approved):
Thanks. Good extension that will be useful later on!

PR #174, opened 2020-09-14
Various dashboard improvements

roecla:
- [x] add id column to the dashboard and use it to uniquely identify parameters everywhere in the monitoring app
- [x] only save and update parameters that are actually plotted
- [x] minor aesthetic improvements (e.g. plot size)
- [x] small fixes in the dashboard how to guide
- [x] new gif of the styled dashboard in the how to guide

### DIFF
--- a/estimagic/dashboard/monitoring_app.py
+++ b/estimagic/dashboard/monitoring_app.py
@@ -48,8 +48,10 @@ def monitoring_app(
     database = load_database(path=session_data["database_path"])
     start_point = _calculate_start_point(database, rollover, jump)
     session_data["last_retrieved"] = start_point
-    start_params, group_to_params = _get_group_to_params_from_database(database)
-    criterion_history, params_history = _create_cds_for_monitoring_app(start_params)
+    start_params, group_to_param_ids = _get_group_to_param_ids_from_database(database)
+    criterion_history, params_history = _create_cds_for_monitoring_app(
+        start_params, group_to_param_ids
+    )
 
     # create elements
     button_row = _create_button_row(
@@ -64,7 +66,7 @@ def monitoring_app(
     monitoring_plots = _create_initial_convergence_plots(
         criterion_history=criterion_history,
         params_history=params_history,
-        group_to_params=group_to_params,
+        group_to_param_ids=group_to_param_ids,
         start_params=start_params,
     )
 
@@ -76,14 +78,14 @@ def monitoring_app(
     doc.add_root(tabs)
 
 
-def _get_group_to_params_from_database(database):
+def _get_group_to_param_ids_from_database(database):
     """Map each group name to the parameters' names that belong to it.
 
     Args:
         database (sqlalchemy.MetaData): Bound metadata object.
 
     Returns:
-        group_to_params (dict): keys are the group names, values are parameter names.
+        group_to_param_ids (dict): keys are the group names, values are parameter names.
 
     """
     optimization_problem = read_last_rows(
@@ -94,8 +96,8 @@ def _get_group_to_params_from_database(database):
     )
     start_params = optimization_problem["params"][0]
     start_params["id"] = _create_id_column(start_params)
-    group_to_params = _map_groups_to_param_ids(start_params)
-    return start_params, group_to_params
+    group_to_param_ids = _map_groups_to_param_ids(start_params)
+    return start_params, group_to_param_ids
 
 
 def _create_id_column(df):
@@ -121,21 +123,21 @@ def _map_groups_to_param_ids(params):
             "group" column and Index.
 
     Returns:
-        group_to_params (dict):
+        group_to_param_ids (dict):
             Keys are the values of the "group" column. The values are lists with
             bokeh friendly strings of the index tuples identifying the parameters
             that belong to this group. Parameters where group is None, "" or False
             are ignored.
 
     """
-    group_to_params = {}
+    group_to_param_ids = {}
     for group in params["group"].unique():
         if group is not None and group == group and group != "" and group is not False:
-            group_to_params[group] = list(params[params["group"] == group]["id"])
-    return group_to_params
+            group_to_param_ids[group] = list(params[params["group"] == group]["id"])
+    return group_to_param_ids
 
 
-def _create_cds_for_monitoring_app(start_params):
+def _create_cds_for_monitoring_app(start_params, group_to_param_ids):
     """Create the ColumnDataSources for saving the criterion and parameter values.
 
     They will be periodically updated from the database.
@@ -145,6 +147,7 @@ def _create_cds_for_monitoring_app(start_params):
     Args:
         start_params (pd.DataFrame): See :ref:`params`. It includes a dashboard
             specific id column.
+        group_to_param_ids ()
 
     Returns:
         criterion_history (bokeh.ColumnDataSource)
@@ -154,10 +157,10 @@ def _create_cds_for_monitoring_app(start_params):
     crit_data = {"iteration": [], "criterion": []}
     criterion_history = ColumnDataSource(crit_data, name="criterion_history_cds")
 
-    param_names = start_params["id"].tolist()
+    param_ids = [id_ for id_list in group_to_param_ids.values() for id_ in id_list]
     params_data = {"iteration": []}
-    for name in param_names:
-        params_data[name] = []
+    for id_ in param_ids:
+        params_data[id_] = []
     params_history = ColumnDataSource(params_data, name="params_history_cds")
 
     return criterion_history, params_history
@@ -189,14 +192,14 @@ def _calculate_start_point(database, rollover, jump):
 
 
 def _create_initial_convergence_plots(
-    criterion_history, params_history, group_to_params, start_params,
+    criterion_history, params_history, group_to_param_ids, start_params,
 ):
     """Create the initial convergence plots.
 
     Args:
         criterion_history (bokeh ColumnDataSource)
         params_history (bokeh ColumnDataSource)
-        group_to_params (dict):
+        group_to_param_ids (dict):
             Keys are the values of the "group" column. The values are lists with
             bokeh friendly strings of the index tuples identifying the parameters
             that belong to this group. Parameters where group is None, "" or False
@@ -210,7 +213,7 @@ def _create_initial_convergence_plots(
 
     """
     param_plots = []
-    for group, param_ids in group_to_params.items():
+    for group, param_ids in group_to_param_ids.items():
         param_names = [
             start_params.query(f"id == '{id_}'")["name"].iloc[0] for id_ in param_ids
         ]
--- a/estimagic/dashboard/monitoring_app.py
+++ b/estimagic/dashboard/monitoring_app.py
@@ -60,8 +60,8 @@ def monitoring_app(
     session_data["last_retrieved"] = start_point
     start_params = read_start_params(path_or_database=database)
     start_params["id"] = _create_id_column(start_params)
-    group_to_param_ids = _map_groups_to_other_column(start_params, "id")
-    group_to_param_names = _map_groups_to_other_column(start_params, "name")
+    group_to_param_ids = _map_group_to_other_column(start_params, "id")
+    group_to_param_names = _map_group_to_other_column(start_params, "name")
     criterion_history, params_history = _create_cds_for_monitoring_app(
         group_to_param_ids
     )
@@ -111,7 +111,7 @@ def _create_id_column(df):
     return ids.astype(str)
 
 
-def _map_groups_to_other_column(params, column_name):
+def _map_group_to_other_column(params, column_name):
     """Map the group name to lists of one column's values of the group's parameters.
 
     Args:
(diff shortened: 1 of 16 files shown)


estimagic/dashboard/monitoring_app.py
```diff
@@ -91,11 +95,26 @@ def _get_group_to_params_from_database(database):
...
     )
     start_params = optimization_problem["params"][0]
-    group_to_params = _map_groups_to_params(start_params)
-    return start_params, group_to_params
+    start_params["id"] = _create_id_column(start_params)
+    group_to_param_ids = _map_groups_to_param_ids(start_params)
+    return start_params, group_to_param_ids
+
+
+def _create_id_column(df):
```
janosg, line 103: If the ID is only used internally, why not just use integers (if necessary coded as string, i.e. `"0", "1", ...` as IDs?

estimagic/dashboard/monitoring_app.py
```diff
@@ -91,11 +95,26 @@ def _get_group_to_params_from_database(database):
...
+    if isinstance(df.index, pd.MultiIndex):
+        index_df = df.index.to_frame()
+        ids = index_df.astype(str).apply(lambda x: "_".join(x), axis=1)
+    else:
+        ids = df.index.to_series().astype(str)
+    # we check in optimize that there are no duplicates in the index but
+    # it is theoretically possible that the ids still have duplicate ids
+    # because everything is converted to strings,
+    # i.e. 1 and "1" would become identical and could introduce duplicates.
+    assert not ids.duplicated().any(), "Duplicates in the dashboard's id column."
```
janosg, line 113: The approach from above would also solve this

estimagic/dashboard/monitoring_app.py
```diff
@@ -136,10 +157,10 @@ def _create_cds_for_monitoring_app(start_params):
     crit_data = {"iteration": [], "criterion": []}
     criterion_history = ColumnDataSource(crit_data, name="criterion_history_cds")
 
-    param_names = start_params["name"].tolist()
+    param_ids = [id_ for id_list in group_to_param_ids.values() for id_ in id_list]
```
janosg, line 160: Such a nested list comprehension is too complex. Can't we just get the `params_ids` from the `start_params` or if we use integers as IDs just construct it as a range?

estimagic/dashboard/monitoring_app.py
```diff
@@ -136,10 +157,10 @@ def _create_cds_for_monitoring_app(start_params):
...
     params_data = {"iteration": []}
-    for name in param_names:
-        params_data[name] = []
+    for id_ in param_ids:
```
janosg, line 162: Here a dictionary comprehension would be more readable

estimagic/dashboard/monitoring_app.py
```diff
@@ -171,28 +192,37 @@ def _calculate_start_point(database, rollover, jump):
...
     Returns:
         convergence_plots (list): List of bokeh Row elements, each containing one
             convergence plot.
 
     """
     param_plots = []
-    for g, group_params in group_to_params.items():
+    for group, param_ids in group_to_param_ids.items():
+        param_names = [
+            start_params.query(f"id == '{id_}'")["name"].iloc[0] for id_ in param_ids
```
janosg, line 218: If we abstract the `group_to_params_ids` function a bit, such that it can map groups to any column we could re-use to get a mapping from groups to names. This would simplify the code here.

estimagic/dashboard/monitoring_callbacks.py
```diff
@@ -149,6 +151,7 @@ def _update_monitoring_tab(
         param_cds (bokeh.ColumnDataSource)
 
     """
+    clip_bound = 1e40
```
janosg, line 154: Please move this to `estimagic.config`

estimagic/dashboard/monitoring_callbacks.py
```diff
@@ -162,36 +165,48 @@ def _update_monitoring_tab(
...
 
     # update the parameter plots
-    param_names = start_params["name"].tolist()
-    params_data = _create_params_data_for_update(data, param_names)
-    param_cds.stream(params_data, rollover=rollover)
+    # Note: we need **all** parameter ids to correctly map them to the parameter entries
+    # in the database. Only after can we restrict them to the entries we need.
+    param_ids = start_params["id"].tolist()
+    params_data = _create_params_data_for_update(data, param_ids, clip_bound)
+    available_keys = param_cds.data.keys()
```
janosg, line 181: As discussed, I am in favor of defining a `stream_data` function that does the reduction automatically instead of calling `cds.stream` with a reduced dictionary, i.e:
```python
def stream_data(cds, data):
    # reduce data to present columns
    # call cds.stream with reduced data
```

estimagic/dashboard/plot_functions.py
```diff
@@ -37,6 +37,10 @@ def plot_time_series(
     plot = create_styled_figure(
         title=title, name=name, logscale=logscale, plot_width=plot_width
     )
+    # this ensures that the y range spans at least 0.1
+    plot.y_range.range_padding = 0.05
```
janosg, line 41: Please move to `estimagic.config`

estimagic/dashboard/plot_functions.py
```diff
@@ -37,6 +37,10 @@ def plot_time_series(
...
+    plot.y_range.range_padding_units = "absolute"
+
     colors = get_colors("categorical", len(y_keys))
 
     legend_items = [(" " * 60, [])]
```
janosg, line 46: Please move to `estimagic.config`

estimagic/dashboard/plot_functions.py
```diff
@@ -84,12 +88,12 @@ def create_styled_figure(
         fig (bokeh Figure)
 
     """
-    plot_width = plot_width if plot_width is not None else 800
+    plot_width = plot_width if plot_width is not None else 750
```
janosg, line 91: Please move to `estimagic.config`

estimagic/dashboard/plot_functions.py
```diff
@@ -84,12 +88,12 @@ def create_styled_figure(
...
 
     name = name if name is not None else title
     y_axis_type = "log" if logscale else "linear"
     fig = figure(
-        plot_height=300,
+        plot_height=250,
```
janosg, line 96: Please move to `estimagic.config`

estimagic/optimization/optimize.py
```diff
@@ -534,7 +534,8 @@ def _single_optimize(
         try:
             algorithm = AVAILABLE_ALGORITHMS[algorithm]
         except KeyError:
-            proposed = propose_algorithms(algorithm, list(AVAILABLE_ALGORITHMS))
+            possibilities = [v.__name__ for v in AVAILABLE_ALGORITHMS.values()]
```
janosg, line 537: Can't we use the keys instead of the `__name__` attribute of the values? This attribute is not always there and might also not be correct. For example, if an algorithm is a partialed function, it does not necessarily have the right `__name__`

estimagic/tests/dashboard/test_monitoring_app.py
```diff
@@ -75,85 +84,112 @@ def fake_read_last_rows(**kwargs):
...
+    multi_params["group"] = ["a", "a", "b", "b"]
+    multi_params["name"] = ["this", "repeats"] * 2
+    multi_params["3rd level"] = [3, 4, 5, 6]
+    multi_params.set_index(["group", "name", "3rd level"], inplace=True)
+
+    res = monitoring._create_id_column(multi_params)
+    expected = pd.Series(
+        ["a_this_3", "a_repeats_4", "b_this_5", "b_repeats_6"], index=multi_params.index
+    )
+    pdt.assert_series_equal(res, expected)
```
janosg, line 111: The distinction between the two cases becomes irrelevant if we move to the proposed solution.

estimagic/tests/dashboard/test_monitoring_app.py
```diff
@@ -75,85 +84,112 @@ def fake_read_last_rows(**kwargs):
...
+    params["id"] = ["a", "b", "c", "d"]
     params.index = ["a", "b", "c", "d"]
     expected = {}
-    res = monitoring._map_groups_to_params(params)
+    res = monitoring._map_groups_to_param_ids(params)
     assert expected == res
 
 
-def test_map_groups_to_params_group_empty():
+def test_map_groups_to_param_ids_group_empty():
```
janosg, line 141: Can we parametrize those tests? They look very similar!

estimagic/dashboard/monitoring_callbacks.py
roecla: I implemented the solution proposed by @hmgaudecker (0afeae5).

estimagic/dashboard/monitoring_callbacks.py
roecla: Done in 0afeae5.

estimagic/dashboard/monitoring_app.py
roecla: I like that solution because then the identifiers give the location in the parameter array. I also set the id to `'None'` for parameters that are not plotted. (0081c97)

estimagic/dashboard/monitoring_app.py
roecla: Would also be possible but the `param_names` are actually not used anymore in this function so I stuck to creating the `param_ids` in a loop from `group_to_param_ids` and got rid of the `start_params` as argument. (0081c97)

estimagic/dashboard/monitoring_app.py
roecla: Done (0081c97).

estimagic/dashboard/monitoring_app.py
roecla: Agreed, much nicer this way. (0081c97)

estimagic/dashboard/plot_functions.py
roecla: I moved all style settings to the config in fbbe96d.

estimagic/optimization/optimize.py
roecla: You are perfectly right (b5fcf5d).

estimagic/tests/dashboard/test_monitoring_app.py
roecla: Yes, that simplified the tests a lot. I parametrized the tests and kept the MultiIndex case in since I switched to using `groupby`. (b5fcf5d)

estimagic/dashboard/monitoring_app.py
```diff
@@ -47,8 +58,13 @@ def monitoring_app(
     database = load_database(path=session_data["database_path"])
     start_point = _calculate_start_point(database, rollover, jump)
     session_data["last_retrieved"] = start_point
-    start_params, group_to_params = _get_group_to_params_from_database(database)
-    criterion_history, params_history = _create_cds_for_monitoring_app(start_params)
+    start_params = load_start_params(path_or_database=database)
+    start_params["id"] = _create_id_column(start_params)
+    group_to_param_ids = _map_groups_to_param_values(start_params, "id")
```
janosg, line 63: `_map_group_to_param_values` is not very good, because it can be any column and rarely will be the value column. 

`_map_group_to_other_column` would be better.

estimagic/dashboard/monitoring_app.py
```diff
@@ -73,60 +90,58 @@ def monitoring_app(
...
+            to the particular group.
 
     """
-    group_to_params = {}
-    for group in params["group"].unique():
-        if group is not None and group == group and group != "" and group is not False:
-            group_to_params[group] = list(params[params["group"] == group]["name"])
-    return group_to_params
+    to_plot = params[~params["group"].isin([None, False, np.nan, ""])]
+    group_to_inds = to_plot.groupby("group").groups
```
janosg, line 128: what does `_inds` stand for?

estimagic/logging/read_log.py
```diff
@@ -35,7 +36,7 @@ def read_optimization_iteration(path, iteration, include_internals=False):
         KeyError: if the iteration is out of bounds.
 
     """
-    database = load_database(path=path, fast_logging=False)
+    database = load_database(**_process_path_or_database(path_or_database))
     optimization_problem = read_last_rows(
```
janosg, line 40: Here we should also use the `read_start_params` function then

estimagic/logging/read_log.py
```diff
@@ -81,3 +82,56 @@ def read_optimization_iteration(path, iteration, include_internals=False):
             del data[key]
 
     return data
+
+
+def load_start_params(path_or_database):
```
janosg, line 87: rename to `read_start_params` to be more consistent.

estimagic/dashboard/monitoring_app.py
roecla: indices. I changed it to group_to_indices in a675a68.

estimagic/dashboard/monitoring_app.py
roecla: Done in a675a68.

estimagic/logging/read_log.py
roecla: True. (a675a68)

estimagic/logging/read_log.py
roecla: Done in a675a68 .

janosg (approved):
Thanks!

PR #176, opened 2020-09-16
Fix plot width.

janosg:
Fix the default plot width of dashboard plots. It was zero/None


roecla (approved):
Thanks for fixing this slip of mine! Works perfectly on three different databases I tried.

PR #182, opened 2020-10-14
Constrained standard errors

### LINKED ISSUE #57: Robust standard error calculation for constraint estimations and better standard error interface

janosg:
Currently, we don't support calculation of likelihood standard errors with constraints.  Mainly, because there is no clear and theoretically clean way to do this. This PR adds a very rudimentary version of a likelihood inference function that works with estimagic constraints. The constraints are handled by calculating a covariance on the internal parameter vector and then using monte carlo integration to calculate the covariance matrix of the free but external parameters.

It is clearly experimental at this stage, but having a basic version out means that we can improve on it instead of writing different scripts in each project. 

closes #57


estimagic/inference/shared.py
```diff
@@ -0,0 +1,118 @@
...
+
+from estimagic.optimization.process_constraints import process_constraints
+from estimagic.optimization.reparametrize import reparametrize_from_internal
+from estimagic.optimization.reparametrize import reparametrize_to_internal
+
+
+def transform_covariance(
+    params, internal_cov, constraints, n_samples, bounds_handling,
+):
+    """Transform in internal covariance matrix to an external one, given constraints.
```
roecla, line 13:
```suggestion
    """Transform the internal covariance matrix to an external one, given constraints.
```

estimagic/inference/shared.py
```diff
@@ -0,0 +1,118 @@
...
+            transformed = reparametrize_from_internal(
+                internal=params_vec,
+                fixed_values=fixed_values,
+                pre_replacements=pre_replacements,
+                processed_constraints=processed_constraints,
+                post_replacements=post_replacements,
+            )
+            transformed_free.append(transformed[is_free])
+
+            free_cov = np.cov(np.array(transformed_free), rowvar=False,)
```
roecla, line 78: I think you can indent `free_cov = np.cov(np.array(transformed_free), rowvar=False,)` one level out and just calculate the covariance once after the loop.

roecla (approved):
Just some very minor comments. Very nice! I think it makes a lot of sense to set up the structure and add a basic version now.

estimagic/inference/shared.py
janosg: Done

PR #185, opened 2020-11-15
Remove name column from params.

janosg:
The `name` and `group` columns are actually only needed in the dashboard and can cause problems in the criterion function. Now we just add them before saving start parameters in the database and never call the criterion functions with the extended parameters.


janosg:
Add changes in #182 instead.

PR #192, opened 2021-02-10
Add option to scale optimization problems

janosg:
## To Do

- [x] use scaling factor and offset in reparametrize functions
- [x] use scaling factor and offset in derivative conversion functions
- [x] use scaling factor and offset when calculating internal bounds
- [x] implement and test functions in `parameter_conversion.py`
- [x] use parameter handling functions in `optimize.py` (and potentially in derivatives)
- [x] use parameter handling functions in tests for reparametrize
- [x] implement and test functions to calculate scaling factor and offset with different methods
- [x] adjust `maximize` and `minimize`
- [x] Add test cases that use scaling for:
    - [x] reparametrization of parameters
    - [x] reparametrization of derivatives
    - [x] optimization


docs/source/how_to_guides/optimization/how_to_scale_optimization_problems.rst
```diff
@@ -0,0 +1,167 @@
...
+        scaling_options={"method": "bounds", clipping_value: 0.0},
+    )
+
+
+Divide by gradient
+------------------
+
+Dividing all parameters by the gradient of the criterion function at the start values
+means that around the start values the problem is scaled optimally. In practice, we do
+not take the exact gradient, but a numerical gradient calculated with a very large step
```
hmgaudecker, line 124: Not obvious how this is supposed to be done. Is "large" absolute or relative to start values? What about bounds? Be explicit in relation previous two approaches, I guess you will have to pick one or another for coming up with "correct" step sizes.

docs/source/how_to_guides/optimization/how_to_scale_optimization_problems.rst
```diff
@@ -0,0 +1,167 @@
...
+    start_params = pd.DataFrame(data=np.arange(5), columns=["value"])
+    start_params["lower_bound"] = 0
+    start_params["upper_bound"] = 2 * np.arange(5) + 1
+
+    minimize(
+        criterion=sphere_with_noise,
+        params=start_params,
+        algorithm="scipy_lbfgsb",
+        scaling=True,
+        scaling_options={"method": "gradient", "clipping_value": 0.1},
```
hmgaudecker, line 156: Clipping value not obvious to me.

hmgaudecker (commented):
Looks great, thanks! 

I am not sure yet about the gradient calculation in the last case.

Also, what precisely is meant by `clipping_value`? Is it obvious that this is the same across all three methods?

hmgaudecker:
While you are at it: Not sure whether it is easier to solve that `delta` issue on the POUNDerS side or whether the quick'n'dirty version would be to add a `magnitude` argument here, which would multiply the start values with that factor (default is 1).

docs/source/how_to_guides/optimization/how_to_scale_optimization_problems.rst
janosg: Has an explanation section now.

docs/source/how_to_guides/optimization/how_to_scale_optimization_problems.rst
janosg: Explained more precisely now.

janosg:
Yes, will add the magnitude argument. It's already documented now.

roecla (approved):

PR #195, opened 2021-03-24
documentation on visualization of simple optimizers

sofyaakimova:
We added our 8 gifs into the source/static/images folder and also modified documentation. Our rst file was added into explanations/optimization folder (the index.rst file was also modified). 
By the way, we were concerned if we should also add the pyfiles? If so where should we put them?


janosg:
Hi Sofya, yes, your py files should also be part of the PR. 

I suggest you create a new folder in `estimagic/docs` called `scripts`. This should contain a reduced version of your pytask project that does not build the documentation, but generates all images and saves them at the appropriate place (i.e. in `estimagic/docs/source/static/images`).

effieHAN:
Hi, Janos. We have added the reduced pytask project as you asked, please have a look. Thank you!

janosg (approved):

PR #199, opened 2021-04-18
Return func evals and info as tidy frame

timmens:
Implements ability to return function evaluations plus information on steps from the derivative function in a tidy data frame.

### DIFF
--- a/docs/source/explanations/differentiation/index.rst
+++ b/docs/source/explanations/differentiation/index.rst
@@ -6,3 +6,4 @@ Differentiation
 
 
    background_numerical_differentiation
+   richardson_extrapolation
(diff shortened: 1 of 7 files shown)


timmens:
Added page on Richardson's extrapolation method to *explanation* section in documentation.

timmens:
Documentation error:

```
(how_to_guides/optimization/how_to_specify_algorithm_and_algo_options: line    8)
 broken    https://gitlab.com/petsc/petsc/-/tree/master/src/binding/petsc4py/ 
```

janosg (approved):
Very nice PR. The explanation of Richardson extrapolation is great. In the long run we can improve the explanation of simple finite differences to make them more similar. Code is also very clean. 

The exact name `tidy_evals` (both as argument and in the output dictionary) might change in the future, but let's do that after the next PR when we have a better understanding of all things that might be returned by this function.

PR #200, opened 2021-05-04
Plot unconsolidated jacobian candidates.

timmens:
Tasks for this PR:

1. Return unconsolidated jacobian candidates from ``first_derivative`` function in a sensible data container
2. Use returned objects to create a "derivative plot"

Remarks.

1. Use either tidy frame or list of dicts
2. Function evaluations as scatter plot with derivative estimates as lines. For richardson method with multiple estimates including error let the line be more transparent depending on the error size


timmens:
I do not like my implementation of ``differentiation.derivative. _convert_evaluation_data_to_frame``, but I don't really know how to avoid the pandas mess. If this will be merged one may need to go back and rewrite some functions to increase readability. Below is an example of how to use the plotting function.

```python
import numpy as np
from estimagic.differentiation.derivatives import first_derivative
from estimagic.visualization.derivative_plot import derivative_plot
```
```python
def func(x):
    y1 = (x[0] - 1) ** 2 + x[1]
    y2 = (x[1] - 1) ** 3
    return np.array([y1, y2])
params = np.ones(3)
```

```python
n_steps, dim_x, dim_f = 5, len(params), len(func(params))

der, info = first_derivative(
    func,
    params,
    n_steps=n_steps,
    return_func_value=True,
    return_evals=True,
    return_jac_cand=True,
)
```

```python
_ = derivative_plot(
    info["df_evals"],
    info["df_jac_cand"],
    info["func_value"],
    params,
    dim_x=[0, 1],
    dim_f=[0, 1],
)
```
output_3_0

janosg (commented):
A few things that should be done before we can merge. I can do some of them if you want me to. 

- [x] One test fails under Python 3.6 and it is related to the new code
- [ ] `first_derivative` should always return just one dictionary with the entry `derivative` and potentially more. Unfortunately, this requires changes at all places where we already use this function internally (inference and optimization). The advantage would be that no matter what options one selects, the result type is the same. 
- [ ] The keys in the result dictionary need clearer names. I think "derivative", "func_value", "derivative_candidates", "func_evals". 
- [ ] Instead of the many optional arguments that determine what is returned we should just have "return_func_value" (default False) and "return_info" (default True). 
- [ ] If `return_info` is False we should not even generate that data because it probably is slow enough to matter in fast optimizations. Speed is the only reason why one would set `return_info` to False. 
- [ ] The `derivative_plot` function should only have `derivative_result` (the above mentioned dict) and `params` as mandatory arguments. Users who only want the plot do not need to know how exactly we store the necessary data. 
- [ ] We do not need arguments for figure height or width. Those can be changed afterwards if necessary. 
- [ ] For now I would remove dim_x and dim_y because they do not fit in with well with the estimagic philosophy. In the long run I would add functions to split plots and recombine them. 
- [ ] We need add a tutorial for first derivatives  [here](https://github.com/OpenSourceEconomics/estimagic/blob/return_plotdata_derivative/docs/source/how_to_guides/differentiation/how_to_calculate_first_derivatives.ipynb) and that's also where the explanatory code snippets from this PR should go.

janosg (approved):
Very nice, thanks for doing all the additional improvements!

PR #206, opened 2021-06-30
Estimation table flex

mpetrosian:
make estimation table more flexible with respect to inputs, render some attributes optional.


janosg (approved):
Thanks!

PR #207, opened 2021-07-01
Roadmap proposal

janosg:
Proposal for a realistic roadmap for the next year or two.

Since I do not have too much time for upfront design right now, the focus is on separating stable from unstable interfaces in the next release. Then users can rely on estimagic for the things that are already mature (optimization, first_derivative) but we are free to improve the rest even if that means breaking changes (inference, sensitivity analyisis). 

Looking for feedback @tobiasraabe @hmgaudecker  @roecla

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -119,5 +119,50 @@ Warning
 =======
 
 Estimagic is still in alpha status and the API might still change. We will try to keep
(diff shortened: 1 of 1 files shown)


janosg:
After many discussions we decided not to merge this.

PR #211, opened 2021-07-07
Raise exception in read_log functions if database does not exist.

janosg:


estimagic/logging/read_log.py
```diff
@@ -119,7 +119,9 @@ def _process_path_or_database(path_or_database):
     if isinstance(path_or_database, MetaData):
         res["metadata"] = path_or_database
     elif isinstance(path_or_database, (Path, str)):
-        res["path"] = Path(path_or_database)
+        res["path"] = Path(path_or_database).resolve()
+        if not res["path"].exists():
+            raise FileNotFoundError(f"No such file: {res['path']}")
```
hmgaudecker, line 124:
```suggestion
            raise FileNotFoundError(f"No such database file: {res['path']}")
```

hmgaudecker (approved):
Looks great, thanks! 

Just wondering about the error one would get in case the path exists, but does not point to a valid database file? (corrupted, just some random file, ...)

janosg:
You would get:

image

But this comes from sqlalchemy, not from estimagic.

hmgaudecker:
Looks good enough.

PR #212, opened 2021-07-15
Estimagic nlopt

### LINKED ISSUE #226: Add selected nlopt optimizers
### Current situation

Nlopt is not supported

### Desired Situation

Add local derivative free and derivative based optimizers from nlopt.

mpetrosian:
resolves #226


estimagic/optimization/nlopt_optimizers.py
```diff
@@ -37,19 +44,25 @@ def nlopt_bobyqa(
...
+    Physics, Cambridge England, technical report NA2009/06 (2009).
 
     ``nlopt_bobyqa`` supports the following ``algo_options``:
 
     - convergence.relative_params_tolerance (float):  Stop when the relative movement
       between parameter vectors is smaller than this.
     - convergence.relative_criterion_tolerance (float): Stop when the relative
       improvement between two iterations is smaller than this.
-    - stopping.max_criterion_evaluations (int): If the maximum number of function
+    - stopping_max_criterion_evaluations (int): If the maximum number of function
```
janosg, line 63: We actually want to use a `.` instead of `_` in the docstring here. In general we use dots after the prefixes `stopping`, `convergence`, `trustregion`, `genetic` and `linesearch`. It does not matter that the arguments themselves have an underscore. We replace the dot internally, but we want it because it makes the grouping of algo options more visible. 

So please use dots after the above mentioned prefixes also in all the other docstrings.

estimagic/optimization/nlopt_optimizers.py
```diff
@@ -141,6 +140,899 @@ def nlopt_neldermead(
...
+):
+    """Minimize a scalar function using principal-axis method.
+
+    This is a gradient-free local optimizer originally described in:
+    Richard Brent, Algorithms for Minimization without Derivatives
+    (Prentice-Hall, 1972). (Reprinted by Dover, 2002.). It assumes quadratic
+    form of the optimized function and repeatedly updates a set of conjugate
+    search directions.
+
+    The algorithm, is not invariant to scaling of the objective function and may
```
janosg, line 161: Here you mention that this algorithm needs scaling but I think you still pass a copy of the default algorithm info where `needs_scaling=False`. I suggest you add the argument `algorithm_info` to `_minimize_nlopt` (with default `None`). 

In `_minimize_nlopt` you use the default algorithm info if None was passed, otherwise you use the one that was passed. 

Currently we don't use the info anywhere but in the long run we want to use it to automatically scale optimization problems that need scaling.

estimagic/optimization/nlopt_optimizers.py
```diff
@@ -141,6 +140,899 @@ def nlopt_neldermead(
...
+    - convergence.relative_params_tolerance (float):  Stop when the relative movement
+      between parameter vectors is smaller than this.
+    - convergence.relative_criterion_tolerance (float): Stop when the relative
+      improvement between two iterations is smaller than this.
+    - stopping_max_criterion_evaluations (int): If the maximum number of function
+      evaluation is reached, the optimization stops but we do not count this
+      as convergence.
+
+    """
+    if rank_variant == 1:
```
janosg, line 701: I would loosen the check a bit: `if rank_variant in (1, "1", "rank_1", "rank-1"): ...`

estimagic/optimization/nlopt_optimizers.py
```diff
@@ -141,6 +140,899 @@ def nlopt_neldermead(
...
+        algo = nlopt.LD_VAR1
+    elif rank_variant == 2:
+        algo = nlopt.LD_VAR2
+    else:
+        raise ValueError(
+            "nlopt supports only rank-1 and rank-2 methods of shifting variable-"
```
janosg, line 707: You need to tell the user more precisely what to do. In the current implementation, she might have passed "rank-1" and get an error that only "rank-1" and "rank-2" is supported. Ideally, error messages first tell you what is possible and then tell you what you did wrong and how you can fix it.

estimagic/optimization/nlopt_optimizers.py
```diff
@@ -141,6 +140,899 @@ def nlopt_neldermead(
...
+    x,
+    lower_bounds,
+    upper_bounds,
+    *,
+    convergence_relative_params_tolerance=CONVERGENCE_RELATIVE_PARAMS_TOLERANCE,
+    convergence_absolute_params_tolerance=CONVERGENCE_ABSOLUTE_PARAMS_TOLERANCE,
+    convergence_relative_criterion_tolerance=CONVERGENCE_RELATIVE_CRITERION_TOLERANCE,
+    convergence_absolute_criterion_tolerance=CONVERGENCE_ABSOLUTE_CRITERION_TOLERANCE,
+    stopping_max_criterion_evaluations=STOPPING_MAX_CRITERION_EVALUATIONS_GLOBAL,
+    nlopt_direct_version=0,
```
janosg, line 791: Since all of those are binary options that can be combined in any way, I would add three boolean arguments instead of the one version argument: `locally_biased`, `unscaled_bounds` and `random`. Then you determine the correct algorithm from that. 

More general: Never ever use integer options or returns that a user has to look up in the documentation. It might be justifiable when using C or Fortran but in Python we always want to automate the lookup.

estimagic/optimization/nlopt_optimizers.py
```diff
@@ -141,6 +140,899 @@ def nlopt_neldermead(
...
+    C. H. da Silva Santos, "Parallel and Bio-Inspired Computing Applied to Analyze
+    Microwave and Photonic Metamaterial Strucutures," Ph.D. thesis, University of
+    Campinas, (2010).
+    H.-G. Beyer and H.-P. Schwefel, "Evolution Strategies: A Comprehensive Introduction,
+    "Journal Natural Computing, 1 (1), pp. 3–52 (2002).
+    Ingo Rechenberg, "Evolutionsstrategie – Optimierung technischer Systeme nach
+    Prinzipien der biologischen Evolution," Ph.D. thesis (1971), Reprinted by
+    Fromman-Holzboog (1973).
+
+    ``nlopt_escht`` supports the following ``algo_options``:
```
janosg, line 890: Is it `esch` or `escht`?

estimagic/optimization/nlopt_optimizers.py
```diff
@@ -141,6 +140,899 @@ def nlopt_neldermead(
...
+    x,
+    lower_bounds,
+    upper_bounds,
+    *,
+    convergence_relative_params_tolerance=CONVERGENCE_RELATIVE_PARAMS_TOLERANCE,
+    convergence_absolute_params_tolerance=CONVERGENCE_ABSOLUTE_PARAMS_TOLERANCE,
+    convergence_relative_criterion_tolerance=CONVERGENCE_RELATIVE_CRITERION_TOLERANCE,
+    convergence_absolute_criterion_tolerance=CONVERGENCE_ABSOLUTE_CRITERION_TOLERANCE,
+    stopping_max_criterion_evaluations=STOPPING_MAX_CRITERION_EVALUATIONS_GLOBAL,
+    random_search_population_size=None,
```
janosg, line 983: We probably want to rename this to `genetic_popsize`. @roecla  is that the name you are using in pygmo? Maybe we also already have a way of calculating the default there and it would be nice if they are harmonized. 

I suggest that we just merge this PR first and then @roecla  makes sure that genetic algorithms from pygmo and nlopt have aligned names and default values.

estimagic/optimization/nlopt_optimizers.py
```diff
@@ -37,19 +44,25 @@ def nlopt_bobyqa(
...
     The implementation is derived from the BOBYQA subroutine of M. J. D. Powell.
 
-    ADD REMAINING EXPLANATIONS FROM NLOPT DOCUMENTATION HERE.
+    The algorithm performs derivative free bound-constrained optimization using
+    an iteratively constructed quadratic approximation for the objective function.
+    Due to its use of quadratic appoximation, the algorithm may perform poorly
+    for objective functions that are not twice-differentiable.
+
+    For details see:
+    M. J. D. Powell, "The BOBYQA algorithm for bound constrained optimization
```
janosg, line 53: If it's not too much it would be very nice if you add all the original papers to our [`refs.bib`](https://github.com/OpenSourceEconomics/estimagic/blob/master/docs/source/refs.bib) file and cite them properly (like [here](https://github.com/OpenSourceEconomics/estimagic/blob/770c9bdce524f269d94ec1a5e6516a82c3da63b8/estimagic/optimization/nag_optimizers.py#L100))

environment.yml
```diff
@@ -38,6 +38,7 @@ dependencies:
   - dill
   - pytask>=0.0.11
   - nlopt
+  - regex
```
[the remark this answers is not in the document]
janosg: Thanks.

estimagic/tests/optimization/test_many_algorithms.py
```diff
@@ -13,17 +13,28 @@
 
 from estimagic.examples.criterion_functions import sos_dict_criterion
 from estimagic.optimization import AVAILABLE_ALGORITHMS
+from estimagic.optimization import GLOBAL_ALGORITHMS
 from estimagic.optimization.optimize import minimize
 
 
+LOCAL_ALGORITHMS = {
+    name: AVAILABLE_ALGORITHMS[name]
```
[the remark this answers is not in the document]
janosg: Thanks.

estimagic/optimization/nlopt_optimizers.py
janosg: Postponed.

estimagic/optimization/nlopt_optimizers.py
janosg: Postponed to #234

estimagic/optimization/nlopt_optimizers.py
janosg: Thanks.

estimagic/optimization/nlopt_optimizers.py
janosg: Thanks

janosg (approved):

Issue #186, opened 2020-12-03
Combination of fixed and decreasing constraint leads to strange result


roecla:
### Bug description

When fixing one parameter and a decreasing constraint that includes the fixed parameter optimizations arrive at strange results.

### To Reproduce

```python
import pandas as pd
from estimagic import minimize

def sum_of_squares(params):
    return (params["value"].to_numpy() ** 2).sum()

params = pd.DataFrame()
params["value"] = [4, 3, 2, 1, 0]

sum_of_squares(params)  # 55

decreasing_constraint = [
    {"loc": params.index, "type": "decreasing"}]
fixed_constraint = [
    {"loc": 0, "type": "fixed", "value": 4},
]

# both constraints are fulfilled by params

# just the decreasing constraint works and arrives at the correct solution: 0 everywhere
just_decreasing_res = minimize(
    criterion=sum_of_squares,
    params=params,
    constraints=decreasing_constraint,
    algorithm="nag_pybobyqa"
)

assert just_decreasing_res["success"]
just_decreasing_res["solution_params"].round(5)

# just fixed works
just_fixed_res = minimize(
    criterion=sum_of_squares,
    params=params,
    constraints=fixed_constraint,
    algorithm="nag_pybobyqa"
)

assert just_fixed_res["success"]
just_fixed_res["solution_params"].round(5)

# specifying both leads to a warning that x0 is above the upper bound 
all_constraints_res = minimize(
    criterion=sum_of_squares,
    params=params,
    constraints=fixed_constraint + decreasing_constraint,
    algorithm="nag_pybobyqa"
)

assert all_constraints["success"]

# pybobyqa converges but to [4, -4, -4, -4, -4]

all_constraints_res["solution_params"]
```

### Expected behavior  & Error messages

see comments in the minimal example above.

### System

 - OS: Ubuntu 20.04
- Estimagic version: 
**branch:** fixes for sid, 2020-11-15 14:34-19
**commit** ed9fa4ca71183c0bb9fa8bb2d8c5375bde19295c

PR #214, opened 2021-07-20
Fix bug that ocurred if fixes and linear constraints overlapped

janosg:
A symptom of the bug was described in #186

### DIFF
--- a/estimagic/tests/optimization/test_all_algorithms_with_sum_of_squares.py
+++ b/estimagic/tests/optimization/test_all_algorithms_with_sum_of_squares.py
@@ -19,7 +19,6 @@ from estimagic.optimization import AVAILABLE_ALGORITHMS
 from estimagic.optimization.optimize import maximize
 from estimagic.optimization.optimize import minimize
 
-
 BOUNDS_FREE_ALGORITHMS = [
     "scipy_neldermead",
     "scipy_conjugate_gradient",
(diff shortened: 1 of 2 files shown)


estimagic/tests/optimization/test_all_algorithms_with_sum_of_squares.py
```diff
@@ -647,6 +646,74 @@ def test_with_sdcorr_constraint_bounds_distance(
...
+    params = pd.DataFrame(data=[[1], [4], [4], [2], [1]], columns=["value"])
+
+    constraints = [
+        {"loc": [1, 2, 3, 4], "type": "decreasing"},
+        {"loc": 2, "type": "fixed", "value": 4},
+    ]
+
+    optimize_func = minimize if direction == "minimize" else maximize
+
+    res = optimize_func(
```
roecla, line 665: Given that the expected result is hard coded to the argmin the test should fail in the case of maximize. Maybe simply `pass` in the case of maximize?

estimagic/tests/optimization/test_all_algorithms_with_sum_of_squares.py
```diff
@@ -647,6 +646,74 @@ def test_with_sdcorr_constraint_bounds_distance(
...
+    algo, direction, crit, deriv, crit_and_deriv
+):
+    params = pd.DataFrame(data=[[1], [2], [3], [4], [1]], columns=["value"])
+
+    constraints = [
+        {"loc": [0, 1, 2, 3], "type": "increasing"},
+        {"loc": 2, "type": "fixed", "value": 3},
+    ]
+
+    optimize_func = minimize if direction == "minimize" else maximize
```
roecla, line 697: Same as above.

roecla (approved):
Just two comments on the test cases. :tada:

estimagic/tests/optimization/test_all_algorithms_with_sum_of_squares.py
janosg: No, we switch the sign of the criterion function in that case. Same is done in all other test cases in that module.

Issue #232, opened 2021-09-20
Implement numerical second derivative


janosg:
### Desired Situation

- We can calculate the second derivative of vector and scalar valued functions that take parameters as numpy arrays or DataFrames
- The three standard hessian formulae are supported. See [ridout](https://www.jstor.org/stable/27644097?refreqid=excelsior%3Acfe40913f11dcbcb9ee25992109b5fa3) for formulae and [statsmodels](https://www.statsmodels.org/stable/_modules/statsmodels/tools/numdiff.html) for a code example. We do not need complex steps. 
- Function evaluations are done in parallel
- Interface and additional outputs similar to `first_derivative`
- Richardson extrapolation
- Check if it is possible to reuse evaluations returned by `first_derivative`

### Proposed implementation

I suggest to do several independent PRs for this

- [x] Add basic functionality without richardson extrapolation building on the step generation functions we also use for first  derivatives. It is important to design the code analogous to `first_derivative`, i.e. with parallel function evaluation and richardson extrapolation in mind. For now focus on numpy array interface. 
- [x] Add pandas interface, dict outputs of func and error handling as in `first_derivative`
- [ ] Add richardson extrapolation
- [ ] Add intermediate outputs and visualization

janosg:
Extensions are not planned anymore.

PR #247, opened 2021-10-23
Add basic option to compute second derivative

timmens:
This will be the first PR in a series to add basic and advanced features regarding the second derivative to estimagic. The PR-series is related to issue #232.

In this PR we tackle the first bullet point in #232, that is, we implement the numerical second derivative with parallel function evaluations.


timmens:
Things that remain to be done:

- [ ] Write tests
- [ ] Find bug which produces wrong results on off-diagonal
- [ ] Capture two and cross step evaluations in data frame

timmens:
Problem: statsmodels implementation is much more accurate which seems to be due to the fact that they are using more complicated steps. Before closing this PR we need to check if we want to adopt this or stick to the simpler version and sacrifice accuracy.

src/estimagic/differentiation/derivatives.py
```diff
@@ -244,11 +244,376 @@ def first_derivative(
...
+    error_handling="continue",
+    batch_evaluator="joblib",
+    return_func_value=False,
+    return_info=True,
+    key=None,
+):
+    """Evaluate second derivative of func at params according to method and step options
+
+    Internally, the function is converted such that it maps from a 1d array to a 1d
+    array. Then the Hessian of that function is calculated. The resulting derivative
```
janosg, line 277:
```suggestion
    array. Then the Hessians of that function are calculated. The resulting derivative
```

janosg (approved):
Well done! Very nice PR! Thanks a lot!

PR #252, opened 2021-11-08
Simplify dashboard and introduce new color theme.

roecla:
- [x] Switch colors 
- [x] Remove logscale button
- [x] Immediately start convergence plot when tab is opened. 
- [x] Require the user to always call the dashboard with **one** database and remove the Master App.


src/estimagic/visualization/colors.py
```diff
@@ -1,266 +1,44 @@
...
-    if not IS_MATPLOTLIB_INSTALLED and as_cmap:
-        raise ValueError("To use as_cmap, install matplotlib.")
+    blue = "#4e79a7"
+    orange = "#f28e2b"
+    red = "#e15759"
+    teal = "#76b7b2"
+    green = "#59a14f"
+    yellow = "#edc948"
+    purple = "#b07aa1"
+    brown = "#9c755f"
```
janosg, line 20: If I see correctly these colors are re-defined in `test_colors.py`. That should be avoided.

src/estimagic/visualization/colors.py
```diff
@@ -1,266 +1,44 @@
...
-            - For non-combined ordered palettes like blue it is between 0 and 12
-            - For the categorical palette it can by any non-negative integer. The colors
-              are repeated if more than 12 colors are requested.
-        as_cmap (bool): If True, the result is returned as matplotlib cmap.
-        skip_dark (int): How many colors to skip from the dark side. Only
-            available for monochrome and combined color palettes.
-        skip_bright (int): How many colors to skip from the bright side. Only
-            available for monochrome and combined color palettes.
+        palette (str): One of ["categorical", "ordered"]
+        number (int): Number of colors needed. Colors are repeated if more than eight
```
janosg, line 6: This should only be for categorical. Repeating ordered should probably rather interpolate a cmap somehow.

janosg (approved):
Looks very good. Only thing: We can rename a bit more, now that there is only going to be one app we do not need to call it monitoring app, for example.

src/estimagic/visualization/colors.py
roecla: Honestly, that was a little excessive testing for such simple functions. I reduced and simplified that.

src/estimagic/visualization/colors.py
roecla: We only use the categorical color scale in estimagic at the moment. However, I think it makes sense to support ordered colors for the future. One option would be to switch to the `nipy_spectral` cmap once we exceed the 6 ordered colors we have: 

image

janosg: Good solution.

PR #258, opened 2021-11-24
Add adaptive trust-region sampling

timmens:
In this PR we want to add a modular trust-region sampling algorithm. For now, we focus on Latin hypercube sampling procedures.

Things that remain to be done:

- [x] Add option that criteria are not evaluated on X but on poly(X, 2), since we want to train a quadratic model using these points
- [x] Finish and extend tests
- [x] Polish explanation notebook


src/estimagic/optimization/trust_region_management.py
```diff
@@ -0,0 +1,344 @@
...
+        candidates[:, -n_new_points:, :] += np.random.uniform(
+            size=(n_new_points, n_dim)
+        )
+
+    candidates = _scale_down_points(candidates, center, radius, n_points)
+
+    crit_func = partial(
+        compute_optimality_criterion, criterion=optimality_criterion, target=target
+    )
+    crit_vals = crit_func(candidates)
```
janosg, line 93: Why partial if we just do one function call afterwards and could also pass in all arguments in the call?

janosg (approved):
Very nice. Just one more thing: We should use the term trust region sampling instead of trust region management throughout. In the literature, trust region management refers to updating the trust region radius and deciding whether and how many points should be sampled. Not to the actual sampling process.

src/estimagic/optimization/trust_region_management.py
timmens: Yes, I agree, it is unnecessary.

PR #265, opened 2021-12-10
Refactor pounders

janosg:
Refactor pounders algorithm to prepare for later extensions.


tests/optimization/test_pounders_integration.py
```diff
@@ -40,12 +42,12 @@ def func(x: np.ndarray, exog: np.ndarray, endog: np.ndarray) -> np.ndarray:
 
 start_params = [
     np.array([0.15, 0.008, 0.01]),
-    np.ones(3) * 0.25,
+    # np.ones(3) * 0.25, # noqa: E800
```
janosg, line 45:
```suggestion
```

janosg (commented):
Thanks! Just a few minor comments. 

Can you please also replace all the pickle files with test fixtures by something human readable like `json`, `csv` or `yaml` (depending on what information it is)?

Since we both worked on this PR we'll also need a separate approval by @timmens.

src/estimagic/optimization/history.py
```diff
@@ -0,0 +1,247 @@
+"""History class for pounders and similar optimizers."""
+import numpy as np
+
+
+class LeastSquaresHistory:
+    """Container to save and retrieve history entries for a least-squares optimizer."""
```
timmens, line 6: Looking at this class alone its not entirely clear what a "history" is. Maybe add one or two sentences to the docstring.

timmens (approved):

tests/optimization/test_pounders_integration.py
segsell: I deleted the line ```# np.ones(3) * 0.25, # noqa: E800``` or do you want to keep it for future reference?

PR #274, opened 2022-01-20
Plot the surface of a criterion function on 2D grids

roecla:
This is the first plot with plotly. 

- [x] basic running version of the plots
- [ ] plot styling (labels, colors, ...)
- [ ] allow user specify a subset of parameters of interest
- [ ] allow user to specify a number of gridpoints for each dimension


janosg:
We will re-do this from scratch with pytree support.

PR #275, opened 2022-01-21
More flexible params specification

janosg:
# More flexible params specification

## Current Situation

Currently, parameters for optimization must be specified as params DataFrame, i.e. as a DataFrame with "value" column. The index can be arbitrary and additional columns are possible. The columns "lower_bound", "upper_bound", "name" and "group" are optional but used in some places if present.

Parameters for numerical differentiation can be specified as params DataFrame or 1d numpy array. In the array case, lower_bounds and upper_bounds  must be passed in separately if required.

This is problematic for several reasons:

1. For very simple criterion functions a 1d numpy array would be a more convenient way to specify start parameters than a DataFrame. It would also reduce the overhead.
2. For very large models, a DataFrame is too restrictive and has led to non-natural choices of index levels. A dictionary of DataFrames or dictionary of arrays could be a better choice.
3. Automatic differentiation tools and just in time compilers (e.g. Jax) do not play well with functions that take a `params` DataFrame.

## Background: What is a pytree

JAX uses the term pytree to refer to a tree-like structure built out of container-like Python objects such as lists, tuples and dicts. Anything that is not a container object is considered a leaf

Examples of pytrees are:

```python
[1, "a", object()]  # 3 leaves

(1, (2, 3), ())  # 3 leaves

[1, {"k1": 2, "k2": (3, 4)}, 5]  # 5 leaves
```

Typically, JAX functions work with pytrees of arrays, i.e. the only possible leaves are arrays, but the many of the `tree_util` functions are written more generally.

## Goal

Any pytree of arrays, numbers and params DataFrames are valid ways of specifying parameters in estimagic. This includes pytrees containing just one leaf, i.e. just a params DataFrame or just a numpy array. This is the same for numerical differentiation, estimation and optimization.

All changes are fully backwards compatible.

## How to specify bounds

Functions that work with lower and upper bounds (`maximize`, `minimize`, `first_derivative`, `second_derivative`) get new arguments `lower_bounds` and `upper_bounds`. The bounds can be specified as pytrees of the same structure as the params pytrees or a subset of those trees (if not all parameters have bounds). If DataFrames are used, the bounds can also be specified as columns within the `params` pytree. If both are specified, the bounds arguments have precedence.

## How to specify constraints

Currently, parameters to which a constraint is applied are selected via a "loc" or "query" entry in the constraint dictionary.

This keeps working as long as params are specified as one DataFrame. If a more general pytree is used we need a "selector" entry instead. The value of that entry is a callable that takes the pytree and returns selected parameters. Depending on the type of constraint the selected parameters might need to have a certain shape or dimensionality (e.g. for increasing constraints the order of selected parameters needs to be defined).

## Tree Utils and their implementations

Since JAX is not available on Windows, we cannot use it as a dependency. Therefore, we need to re-implement functions from the tree_util package. In particular we need:

- [x] tree_flatten
- [x] tree_unflatten
- [x] tree_map
- [x] tree_multimap

Moreover, we might need a few more functions:

- [x] `tree_update`: Update the leaves in one pytree with leaves from another one.
- [x] `leaf_names`: Construct (unique) names for the leaves of a pytree
- [x] `tree_equal`: Check if two pytrees are equal

We want to have interfaces similar to jax but not follow them religiously. 

## Deviation from JAX

JAX has one global pytree registry that determines which objects are considered containers. All others are considered leaves. An `is_leaf` argument sometimes allows to treat a type that is registered as container as leave. 

We sometimes want to treat arrays, DataFrames and Series as leaves and sometimes as containers. Moreover, we might need different flatten and unflatten functions for DataFrames in different situations. Thus a global registry for pytree containers is not suitable. 

Instead many of the function in `estimagic.tree_util` have an optional `registry` argument that allows to switch out the entire registry for a different one. The most frequently used registries can be accessed by not specifying a registry or by specifying "extended". 

Another deviation from JAX is the treedef, i.e. the information needed to reconstruct a pytree from a list of leaves. Thus while `tree_flatten` and `tree_unflatten` can be used in exactly the same way as in JAX, a tree that has been flattened with estimagic could not be unflattened with JAX. 

## Using JAX's autodiff during Optimization

While we allow for pytrees of arrays, numbers and DataFrames, JAX only allows pytrees of arrays and numbers for automatic differentiation.

If you want to use automatic differentiation with estimagic you will thus have to restrict yourself in the way you specify parameters.

We will try to find a way of extending JAX but it probably won't happen very soon.

## To-Do

- [ ] Write more tests for tree_util functions against JAX
- [ ] Use pytree functions during optimization
    - [ ] Adjust constraints processing
    - [ ] Adjust bounds processing
    - [ ] Adjust reparametrization of parameters 
    - [ ] Adjust reparametrization of derivatives
    - [ ] Adjust how name and group are stored in log
    - [ ] Convert benchmarking to 1d numpy array interface
    - [ ] Convert most tests to 1d numpy array interface
    - [ ] Adjust documentation
- [ ] Use pytree functions in numerical differentiation
    - [ ] First derivative
    - [ ] Second derivative
    - [ ] Adjust documentation
- [ ] Make sure pytrees work well with estimation
    - [ ] estimate_ml
    - [ ] estimate_msm
    - [ ] allow moments in `estimate_msm` to be pytrees
- [ ] Make sure pytrees work well for visualization
    - Estimation table (probably just needs better documentation and best practices)
    - `plot_univariate_effects` and `plot_bivariate_effects`


hmgaudecker:
Just trying to use this PR (great stuff!) to understand how these  pytrees actually work.

> JAX uses the term pytree to refer to a tree-like structure built out of container-like Python objects such as lists, tuples and dicts. Anything that is not a container object is considered a leaf

In case you want to re-use this for docs: I would suggest changing the language because NumPy arrays are also often referred to as containers (e.g., [here](https://scipy-lectures.org/intro/numpy/array_object.html)). Typical language use is not fully clear to me, which also implies later design choices. How does estimagic decide whether something is a container-like thing or a leaf? Clear for the examples (list, tuple, dict, set vs. NumPy array, Pandas DataFrame), but what about, say, cuDF DataFrames or Jax arrays?



> Typically, JAX functions work with pytrees of arrays, i.e. the only possible leaves are arrays, but the many of the `tree_util` functions are written more generally.

Below it says scalars or arrays -- would be better to pick one :-)

To what extent should this be compatible with Jax' pytrees? Is it worth checking with them and trying to copy their API tests they have?

janosg:
## Regarding the first question (What is a container, what is a leaf):

JAX decides this based on a global registry of containers and an `is_leaf` argument (a callable). We also have the `is_leaf` argument, but the registry is not a global variable. Details are described in the updated PR description. 

Whether numpy arrays are containers to be flattened depends on whether they are in the registry or not. How they are flattened is also defined in the registry. E.g. sometimes flattening a DataFrame might mean to extract the element of the "value" column into a list but in other occasions it might meant to extract all numerical values into a list. 

Jax arrays, cuDFs and custom classes can all be registered. So people could even use the Parameter Classes estimagic tried to make obsolete ;-)

## Regarding testing against Jax

Yes, apart from the deviations mentioned above we want to test against JAX.

janosg:
After writing the To-Do and updating the comment, I think this might be a good time to start a formal enhancement proposal process. A lot of the points mentioned above deserve more discussion and a more detailed description. Who else thinks we should move it to EEP1?

janosg:
The code has been moved to [pybaum](https://github.com/OpenSourceEconomics/pybaum). 

The design proposal and discussion has been moved to #278

PR #276, opened 2022-01-26
Add basic version of parallelized Nelder-Mead

jacekb95:
Python implementation of the algorithm by Lee D., Wiswall M., A parallel implementation of the simplex function minimization routine, Computational Economics, 2007.


janosg (approved):
Hi Jacek, thank you very much for this great PR. It has been a very nice experience working with you. I added `neldermead_parallel` to the documentation and registered it as available algorithm so it can be selected in `minimize` and `maximize`. Also did some minor renaming. Will make a release later but it can take a few hours until it becomes available on conda.

PR #282, opened 2022-02-21
Improve estimation table

mpetrosian:
- [x] Add comprehensive number formatting options
    - [x]  Allow for the scientific format of numbers
- [x]  Add a new return type of a data frame
- [x]  Improve docstrings, make them up to date with changes in functions.
- [x] Bundle operations from `estimation_table `into functions.
    - [x]  Add tests for these functions.
- [x] Now, with `return_type="render_inputs"`,  return a unified dictionary of `render_inputs` that works both in `render_html` and `render_latex`

TODO:
- [ ] Improve documentation


ChristianZimpelmann:
Thank you! Just a few quick comments:

-  Do you plan to keep both `custom_param_names` and `custom_index_names`? If no, I would very much prefer the dictionary (`custom_param_names`).
- Is there an option to use the dependent variable in each regression as column name?
- I still have the problem that statistics in the footer should be rounded differently than values in the body. E.g, Observations shouldn't be rounded at all. (I think the best would be to introduce a new argument that contains a mapping of statistics to number format. The existing `number_format` keyword would then only be applied to the body_df and to all statistics for which no specific number_format is specified.

mpetrosian:
> Thank you! Just a few quick comments:
> 
> * Do you plan to keep both `custom_param_names` and `custom_index_names`? If no, I would very much prefer the dictionary (`custom_param_names`).
> * Is there an option to use the dependent variable in each regression as column name?
> * I still have the problem that statistics in the footer should be rounded differently than values in the body. E.g, Observations shouldn't be rounded at all. (I think the best would be to introduce a new argument that contains a mapping of statistics to number format. The existing `number_format` keyword would then only be applied to the body_df and to all statistics for which no specific number_format is specified.
Hey. 
- I think, yes we want to keep both arguments since the first applies to index values (e.g. change "constant" to "intercept") and the second to index name (e.g. "parameters" to "variables")
- In principle, yes. One way of doing it would be to pass `custom_col_names = [model.endog.name for model in models]`
- n_obs is excluded from fromatting in `create_statistics_series`

hmgaudecker:
> n_obs is excluded from fromatting in `create_statistics_series`

I think that should be user-controllable. Sometimes you have N, N x T, some other numbers...  Whatever ends up in a regression table, let alone outputs of a structural model, can be very heterogeneous!

janosg (approved):
Overall very very nice!

I need two more changes:

1. Replace all asserts by proper errors
2. Reverse the stat_keys dict and rename it to stats_options. The new docstring entry should be:

stats_options (dict): A dictionary that determines which statistics (e.g. R-Squared,
No. of Observations) are displayed and how they are labeled. The keys are the names
of the statistics inside the ``model.info`` dictionary or attribute names of a 
statsmodels results object. The values are the new labels to be displayed for those 
statistics, i.e. the set of the values is used as row names in the table.

src/estimagic/visualization/estimation_table.py
```diff
@@ -265,184 +358,715 @@ def tabular_tex(
         )
     latex_str = latex_str.split("\\bottomrule")[0]
     if show_footer:
-        stats_str = footer_df.to_latex(**default_options)
-        stats_str = (
-            "\\midrule" + stats_str.split("\\midrule")[1].split("\\bottomrule")[0]
-        )
+        if "Observations" in stats.index.get_level_values(0):
+            stats = stats.copy(deep=True)
+            stats.loc[("Observations",)] = _add_multicolumn_left_format(
```
ChristianZimpelmann: It would make sense to apply this to all other integer statistics (e.g. if the "Observations" index entry is called differenctly. Alternatively, I would let the user enter a list of statistics which should be treated as integers (and hence left aligned)

src/estimagic/visualization/estimation_table.py
```diff
@@ -265,184 +358,715 @@ def tabular_tex(
...
+                )
+            name = model.name
+        else:
+            name = None
+
     else:
         if isinstance(model, dict):
-            params_df = model["params"].copy(deep=True)
-            info_dict = model.get("info", {})
+            params = model["params"].copy(deep=True)
```
ChristianZimpelmann, line 489:
```suggestion
            params = model["body"].copy(deep=True)
```

src/estimagic/visualization/estimation_table.py
```diff
@@ -598,24 +1240,29 @@ def _generate_notes_latex(append_notes, notes_label, sig_levels, custom_notes, d
     """
     n_levels = df.index.nlevels
     n_columns = len(df.columns)
-    sig_levels = sorted(sig_levels)
+    significance_levels = sorted(significance_levels)
     notes_text = "\\midrule\n"
```
ChristianZimpelmann, line 1244: This midrule is not needed if `append_notes == False`

ChristianZimpelmann: I suggest to move `\\midrule\n` in the if statement below

ChristianZimpelmann (approved):
Just three very small open suggestions (can be merged afterwards)

src/estimagic/visualization/estimation_table.py
```diff
@@ -265,184 +358,715 @@ def tabular_tex(
...
+            model column.
+
+    Returns:
+        column_names (list): The column names to display in the estimatino table.
+
+    """
+    if not custom_col_names:
+        col_names = default_col_names
+    elif isinstance(custom_col_names, dict):
+        col_names = list(pd.Series(custom_col_names).replace(custom_col_names))
```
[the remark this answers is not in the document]
mpetrosian: You're right, thanks!

src/estimagic/visualization/estimation_table.py
```diff
@@ -265,184 +358,715 @@ def tabular_tex(
...
+    elif isinstance(custom_col_names, list):
+        col_names = custom_col_names
```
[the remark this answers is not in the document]
mpetrosian: Done.

src/estimagic/visualization/estimation_table.py
mpetrosian: Done.

PR #283, opened 2022-02-21
Add quadratic trust-region subsolvers

segsell:
In this PR, we add the following two quadratic subproblem solvers, which are used by the POUNDERS algorithm:

1) ``trust-exact``: Nearly exact trust-region solver for a quadratic subproblem. Based on the original ``GQTPAR`` algorithm. Does not support bounds.
2) ``BNTR``: Bounded Newton Trust Region for nonlinear minimization with bound constraints.

To do:
- [x] Implement ``GQTPAR`` algorithm
- [x] Add Conjugate Gradient trust-region algorithm, which is called by ``BNTR``
- [x] Implement ``BNTR``
- [x] Adjust POUNDERS subproblem solver interface and remove solvers from ``scipy``
- [x] Adjust POUNDERS main interface and add additional convergence criteria for subproblem solvers


janosg (approved):
While not completely polished yet, this is a huge improvement over the current status. Thus we merge and postpone the rest to a follow up PR. Thanks a lot!

PR #278, opened 2022-01-28
Add estimagic enhancement proposal (EEP-01) for pytrees

janosg:

PR #286, opened 2022-03-02
Add pytree support for derivatives

timmens:
Following the EEP-01 (#278), in this PR we will add pytree support to the first derivative.

I rewrote first and second derivative functions to work with input (``params``) pytrees and output (``func(params)``) pytrees. The Jacobian and Hessian are now (block-) pytrees.


timmens:
At the moment the derivative functions expect a ``key`` argument whenever ``func(params)`` is a dictionary. Should we continue to support this behavior or should we check whether ``key`` is not None and if not take the derivative of ``func(params)[key]``? @janosg

src/estimagic/differentiation/derivatives.py
```diff
@@ -31,29 +38,24 @@ def first_derivative(
...
 
     Internally, the function is converted such that it maps from a 1d array to a 1d
-    array. Then the Jacobian of that function is calculated. The resulting derivative
-    estimate is always a :class:`numpy.ndarray`.
+    array. Then the Jacobian of that function is calculated.
 
-    The parameters and the function output can be pandas objects (Series or DataFrames
-    with value column). In that case the output of first_derivative is also a pandas
-    object and with appropriate index and columns.
+    The parameters and the function output can be general pytrees. By default the
```
janosg, line 49: We should call it `estimagic-pytrees` instead of general pytrees and link to corresponding section of the EEP. General pytrees makes it sound like a pytree of strings would work.

src/estimagic/differentiation/derivatives.py
```diff
@@ -127,19 +125,16 @@ def first_derivative(
...
-    # convert params to numpy, but keep label information
-    params_index = (
-        params.index if isinstance(params, (pd.DataFrame, pd.Series)) else None
-    )
-
-    x = params["value"].to_numpy() if isinstance(params, pd.DataFrame) else params
-    x = np.atleast_1d(x).astype(float)
+    # convert params to numpy
+    registry = get_registry(extended=True)
+    x, params_tree_def = tree_flatten(params, registry=registry)
```
janosg, line 136: `treedef` not `tree_def` to be consistent with rest.

src/estimagic/differentiation/derivatives.py
```diff
@@ -998,25 +934,36 @@ def _split_into_str_and_int(s):
...
+        # combine derivative candidates in accessible data frame
+        derivative_candidates = _convert_richardson_candidates_to_frame(
+            *updated_candidates
+        )
+        info["derivative_candidates"] = derivative_candidates
 
     return info
+
+
+def _is_exactly_nan(value):
```
janosg, line 960: Maybe rename to `_is_scalar_nan`?

src/estimagic/differentiation/derivatives.py
```diff
@@ -998,25 +934,36 @@ def _split_into_str_and_int(s):
...
+    return isinstance(value, float) and np.isnan(value)
+
+
+def _unflatten_if_ndarray(leaves, tree_def, registry):
```
janosg, line 964: I guess if it's not a ndarray it's a scalar nan? If so, I would rename to `_unflatten_if_not_nan`. Partially because I find it more readable, partially because it should work for alternative array implementations such as jax arrays.

src/estimagic/parameters/block_trees.py
```diff
@@ -7,114 +7,239 @@
 from pybaum import tree_unflatten
 
 
-def matrix_to_block_tree(matrix, tree1, tree2):
-    """Convert a matrix to block-tree.
+def matrix_to_block_tree(matrix, tree_outer, tree_inner):
```
janosg, line 10: I think `outer_tree` is better than `tree_outer` (and same for inner)

janosg (approved):
Just a few small comments. This is a very nice PR. Thank you very much! Especially the `get_bounds` function is a beautiful use of `pybaum`! The test cases are extremely thorough and still simple throughout!

tests/jax/test_compare_derivatives.py
```diff
@@ -0,0 +1,149 @@
...
+from pybaum import tree_equal
+
+if not IS_JAX_INSTALLED:
+    pytestmark = pytest.mark.skip(reason="jax is not installed.")
+else:
+    import jax
+    import jax.numpy as jnp
+
+
+# arrays have to be equal up to 5 decimals
```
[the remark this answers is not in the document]
timmens: The problem is that our second derivatives are imprecise. Hopefully, we can increase the precision with Richardson extrapolations for the second derivative.

src/estimagic/parameters/parameter_bounds.py
```diff
@@ -0,0 +1,61 @@
...
+        np.ndarray: Consolidated and flattened lower_bounds.
+        np.ndarray: Consolidated and flattened upper_bounds.
+
+    """
+    registry = get_registry(extended=True)
+    n_params = len(tree_leaves(params, registry=registry))
+
+    registry.pop(pd.DataFrame)
+    bounds_tree = tree_map(
+        lambda leaf: leaf if isinstance(leaf, pd.DataFrame) else np.nan,
```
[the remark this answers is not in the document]
timmens: Yes, very good catch. That would have been a bug.

tests/jax/test_compare_derivatives.py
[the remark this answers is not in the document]
timmens: 64 bit precision for jax is enabled now.

PR #287, opened 2022-03-02
added the option of pre-calculated hessian

PaulBehler:
Since my change for checking whether it´s already optimized was exactly the same as yours, I did only make one very small change, namely adding the option of using a pre-calculated hessian.
 Note: 2 Tests failed: "test_fides_stop after_tiny_time" "and test_algorithm_on_sum_of_squares_with_binding_bounds[fides]" with one of them (I think the latter) already failing before my changes, so I don´t know, if this has something to do with my changes. Also, I have not yet written a new test, mainly I didnt know how exactly I could implement this in the existing test for estimate_ml. But I could of course still try it if needed.


janosg:
I will tackle this problem as part of the switch to pytrees.

PR #290, opened 2022-03-10
Add linear trust-region subsolver TRSBOX

segsell:
In this PR, we add the ``TRSBOX`` algorithm for solving linear trust-region subproblems. The algorithm supports bound constraints.

It is based on [Powell (2009) "The BOBYQA algorithm for bound constrained optimization without derivatives."](https://www.damtp.cam.ac.uk/user/na/NA_papers/NA2009_06.pdf)


janosg (approved):
Thanks!

PR #295, opened 2022-03-23
Fixing a small bug in estimation_table

mpetrosian:
- [x] Fix a bug that arises in `_customize_col_names` when passing a list as `custom_col_names`.


janosg (approved):
Thanks!

PR #304, opened 2022-04-21
Remove chaospy dependency

segsell:
In this PR, we get rid of the `chaospy` dependency and rewrite the function [`draw_exporation_sample()`](https://github.com/OpenSourceEconomics/estimagic/blob/7484225bb7e86d28c4ab6a933a1c86943878f493/src/estimagic/optimization/tiktak.py#L222). 

As an alternative to the sampling schemes provided by `chaospy`, we make use of `scipy`'s [Quasi-Monte Carlo module](https://scipy.github.io/devdocs/reference/stats.qmc.html), which offers support for low discrepancy sequences.

To-Do
- [x] Implement sampling schemes provided by scipy's QMC module
    - [x] Sobol
    - [x] Halton
    - [x] Latin hypercube
    - [x] Random
- [x] Add support for the following sampling distributions:
    - [x] Uniform
    - [x] Triangular

### DIFF
--- a/.conda/meta.yaml
+++ b/.conda/meta.yaml
@@ -36,7 +36,6 @@ requirements:
     - sqlalchemy >=1.3
     - seaborn
     - dill
-    - chaospy
     - pybaum
 
 test:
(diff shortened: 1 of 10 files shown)


janosg (approved):
Thanks a lot!

PR #306, opened 2022-04-28
Polish quadratic subsolver BNTR

segsell:
In this PR, we polish the quadratic subsolver `BNTR`. In particular, we fix a bug where bound constraints were not handled properly.


src/estimagic/optimization/_trustregion_bounded_newton_quadratic.py
```diff
@@ -74,224 +93,259 @@ def take_preliminary_steepest_descent_step_and_check_for_solution(
...
     lower_bounds,
     upper_bounds,
     active_bounds_info,
     maxiter_steepest_descent,
+    options_update_radius,
 ):
     """Perform steepest descent step and update trust-region radius."""
-    x_trial = np.copy(x_candidate)
-    f_min = np.copy(f_candidate)
+    f_min = copy(f_candidate_initial)
```
janosg, line 275: You make many copies of arrays at places where it is not immediately clear why copying is needed. Are they all necessary or just defensive? If they are defensive, I would prefer to simply not do inplace changes anywhere and get rid of all the copying.

Also, you use `copy`, `np.copy` and `arr.copy`. I would prefer to use `arr.copy` everywhere.

janosg (commented):
Thanks a lot! Just a few small comments.

src/estimagic/optimization/_trustregion_bounded_newton_quadratic.py
segsell: Good point! Is a copy needed at all if we define a variable like this 

```python
if converged:
   hessian_inactive = model.square_terms

else:
   hessian_inactive = find_hessian_submatrix_where_bounds_inactive(
      model, active_bounds_info
      )
```
where `model` is a `namedtuple`?

janosg: It depends. In the `if` clause you would need a copy if later there is any place where `hessian_inactive` gets changed in place but you do not want to have that change in `model.square_terms`, e.g. if you set some elements of `hessian_inactive` to zero without making a copy first. 

That is why it is very hard to tell whether copies are needed just from looking at the definition, unless you embrace functional style where 1. everything happens in functions and 2. no function modifies its inputs. 

Below I make 3 examples and discuss why I prefer the last:

1. Really wrong: 
```python
hessian_inactive = model.square_terms 
# a lot of code here
hessian_inactive[0, 0] = 0
```
-> model.square_terms[0, 0] is zero

2. Correct but hard to maintain
```python
hessian_inactive = model.square_terms.copy()
# a lot of code here
hessian_inactive[0, 0] = 0
```
-> model.square_terms is not changed but there will come the moment where I want to make the code faster and wonder if I can remove the `.copy`. Then I have to read a lot of code to find that out.

3. Functional style
```
def set_first_element_to_zero(arr):
    arr = arr.copy()
    arr[0, 0] = 0
    return arr

hessian_inactive = model.square_terms
# a lot of code here
hessian_inactive = set_first_element_to_zero(hessian_inactive)
```
The reason why the last is best is that you directly see why the copy is needed.

In fact, unless you have an iterative algorithm that changes hessian_inactive multiple times it would be better to not even initialize it with `model.square_terms` and directly do `hessian_inactive = set_first_element_to_zero(model.square_terms).

segsell: Thanks for the detailed explanation!

segsell:
Ready to merge from my side.

src/estimagic/optimization/_trustregion_bounded_newton_quadratic.py
```diff
@@ -1,297 +1,365 @@
...
+                gtol_scaled=gtol_scaled,
             )
-        else:
-            x_accepted = x_candidate
-            f_accepted = f_candidate
 
-        trustregion_radius = max(trustregion_radius, max_trustregion_radius)
-        trustregion_radius = min(max(trustregion_radius, 1e-10), 1e10)
+        if not converged:
+            trustregion_radius = max(trustregion_radius, radius_upper_bound)
```
janosg, line 142: This line looks redundant since you do clipping in the next line.

janosg (approved):
Looks all good now. Just one minor comment.

src/estimagic/optimization/_trustregion_bounded_newton_quadratic.py
segsell: `radius_upper_bound` and `options_update_radius["max_radius"]` are two different things. 
1) `radius_upper_bound` gets - potentially - updated during the gradient descent iterations.
2) `options_update_radius["max_radius"]` is a default option and does not get changed (maybe I could have been clearer with the naming here. I chose the `dict` type over a `namedtuple` to signal that the entries of `options_update_radius` do not get overwritten)

In line 141, we set `trustregion_radius` equal to `radius_upper_bound` in case we have not converged. Clipping is then done as a safeguarding step to ensure `trustregion_radius` lies within the default bounds, `"min_radius"` and `"max_radius"`.

janosg: Ok!

PR #311, opened 2022-05-02
Move to plotly

timmens:
This PR extends #288.

In this PR we

- Remove matplotlib and seaborn from the dependencies
- Update the use of plotly figures in the documentation notebooks
- Update the ```derivative_plot```


environment.yml
```diff
@@ -54,4 +54,4 @@ dependencies:
       - DFO-LS
       - fides==0.7.4
       - gif
-      - gif[matplotlib]
```
janosg, line 57: I guess in the dev environment we still need to keep seaborn and gif[matplotlib] to generate some of the plots in the documentation. Can you confirm by running pytask once in the docs folder?

src/estimagic/visualization/colors.py
```diff
@@ -1,4 +1,11 @@
-import seaborn as sns
```
janosg, line 1: This whole module should go. Are we still using it anywhere?

src/estimagic/visualization/convergence_plot.py
```diff
@@ -30,6 +24,9 @@ def convergence_plot(
     stopping_criterion="y",
     x_precision=1e-4,
     y_precision=1e-4,
+    combine_plots_in_grid=True,
+    template=PLOTLY_TEMPLATE,
+    palette=px.colors.qualitative.Plotly,
```
janosg, line 29: For the color palette we should also have a global (imported from config) as default. I.e. similar to `PLOTLY_TEMPLATE`.

While we are at it: I think the colors look really bad in derivative plots and not too good in some others. Maybe we should try out D3, G10 and T10 from [here](https://plotly.com/python/discrete-color/) to see if we can get a nicer default without much effort.

janosg (approved):

environment.yml
timmens: You're right. For the documentation we need both seaborn and matplotlib.

src/estimagic/visualization/colors.py
timmens: Moved to ``src/dashboard``.

src/estimagic/visualization/convergence_plot.py
timmens: The Set2 template provides a slightly better default, but prospectively we may want to create our own.

hmgaudecker: My suggestion: Let's not get distracted. Economists should trust specialization. If anything, pick some color cycle from somewhere, but let's not take on the burden of discussions about this sort of thing.

Issue #323, opened 2022-05-11
Better result object for optimization


janosg:
### Current situation

- `maximize` and `minimize` return a dictionary with information
- The dictionary keys could be improved (e.g. `solution_params` should be just `params` if the whole dictionary the solution of the optimization problem
- When printed the result is not very readable
- The history of criterion evaluations is not part of the result

### Describe the solution you'd like

`maximize` and `minimize` should return a custom object (`OptimizationResult`). The main advantages will be:
- By implementing a `__repr__` we can make terminal/notebook output more informative. In particular, the success and message entries of the current result dictionary should be displayed very prominently. 
- All the current functionality stays available as attributes, but we can design the names more carefully than in the current dict.
- We can add a few methods to provide additional functionality:
    - `.convergence_plot()`
    - `.to_dict()`
    - `.from_dict()`
    - ...

### Backwards compatibility

We could make the results object fully backwards compatible by implementing `__getitem__` to make it behave like a dict. However, I think the userbase is currently small enough that we could simply make a breaking change. 

### Question

Do we need an enhancement proposal for this or can I make a PR? Main reason why an enhancement proposal would be cumbersome is that I think the design will evolve during implementation.

Issue #314, opened 2022-05-06
Improve Benchmarking 


janosg:
### Problems:

1. The optimization histories for benchmarks are from a log database. The intention behind this was to get thread safety for optimizers that parallelize. However, when running benchmarks it is almost always a good idea not to parallelize on the level of function evaluations but on the benchmark problem level. 
2. The processing of optimization histories for the benchmark assumes that parameters are a DataFrame. It also does all calculations in pandas which might not be ideal. It would currently not run with general pytree parameters.

### Solution

1. Add an option to `maximize` and `minimize` that allows to collect the optimization histories (i.e. function evaluations and parameters) in memory. The mechanism can be similar to how the cache works, i.e. a list is partialled into `internal_criterion_and_derivative_template` and in each iteration, the parameters and function value are appended to that list. In contrast to the cache, this behavior is switched on by the user and not determined from the `algo_info`. If it is on, the histories are added to the optimization result. The following things should be available:
    - the scalar criterion value 
    - the raw criterion output 
    - the external params
2. Rewrite the processing of the optimization histories for general pytrees. Add a fast path in case that parameters are numpy arrays. Use mainly numpy, not pandas for the calculations.
3. Adjust convergence_plot 
4. Adjust profile_plot

PR #324, opened 2022-05-12
Collect histories and add them to output of optimization

janosg:
## Goal

collect the optimizer histories (for criterion values and parameter vectors that have been tried) during optimization and add them to the result. 

For now, this will only worked for optimizers that do not parallelize. It will however work with parallel numerical derivatives or parallel multistart optimizations.

This is a preparation for #323 and #314 

## To-Do

- [x] Get rid of `first_eval` and `cache` in `internal_criterion_and_derivative`
- [x] Improve internal algorithm interface such that we can find out if an optimizer actually parallelizes
- [x] Add history collection for non parallelizing optimizers
    - [x] standard optimization
    - [x] multistart
- [x] Make history collection optional

### DIFF
--- a/src/estimagic/batch_evaluators.py
+++ b/src/estimagic/batch_evaluators.py
@@ -124,7 +124,10 @@ def joblib_batch_evaluator(
     def internal_func(*args, **kwargs):
         return func(*args, **kwargs)
 
-    res = Parallel(n_jobs=n_cores)(delayed(internal_func)(arg) for arg in arguments)
+    if n_cores == 1:
+        res = [internal_func(arg) for arg in arguments]
+    else:
+        res = Parallel(n_jobs=n_cores)(delayed(internal_func)(arg) for arg in arguments)
 
     return res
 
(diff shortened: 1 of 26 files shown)


hmgaudecker:
I am sure you thought about this, but why? There is a certain beauty in small-ish objects. Why add something that can easily reach thousands if I can have  simple convenience function reading the logs?

janosg:
There are two reasons:

1. Logging introduces a very large overhead (at least 1ms per criterion evaluation) when criterion functions are fast. This kills us when running benchmarks, where the typical criterion function takes microseconds. This is the primary reason, as it will really speed up the development of tranquilo.
2. I hope that more people look at convergence plots if creating them takes just one line in a notebook which will be the case after #323; Requiring logging to be active for that will be enough to discourage most users. 

From looking at student projects I also get the feeling that few people know about and use logging. 

But it's good to keep your point in mind and make the in-memory collection of histories optional.

timmens (approved):
Approved. Things that need to be done are listed in #327.

PR #325, opened 2022-05-12
Slice Plot

mpetrosian:
- [x] Split  `slice_plot` into functions that get a dictionary with plots for each parameter (`get_slice_plots`) and combine the plots (`combine_plots`)
- [x] Use subplots to combine plots instead of facetgrid => improved display of axes labels


janosg (approved):
Thanks! Very nice!

PR #328, opened 2022-05-16
Improve subsolver BNTR

segsell:
In this PR, we improve the conjugate gradient step of the BNTR (Bounded Newton Trust-Region) algorithm by adding two conjugate gradient methods. 
Previously, the only available option was an off-the-shelf conjugate gradient implementation.

The new methods are:
- [x] Steihaug-Toint (see [Steihaug, 1983](https://epubs.siam.org/doi/10.1137/0720042))
- [x] TRSBOX (see [Powell, 2009](https://www.damtp.cam.ac.uk/user/na/NA_papers/NA2009_06.pdf))

The respective flags for our three methods are:
- "cg"
- "steihaug_toint"
- "trsbox"

TRSBOX performs best on the ``more_wild`` benchmark set and is thus the recommended default.


.pre-commit-config.yaml
```diff
@@ -50,7 +50,7 @@ repos:
     rev: 22.3.0
     hooks:
       - id: black
-        language_version: python3.9
+        language_version: python3.8
```
segsell, line 53: I changed `language_version` to `python3.8`. Otherwise `pre-commit` would throw an error. 
Alternatively, pinning `python==3.9` in `environment.yml` and re-installing the conda environment worked. 

If this is a local problem, we can change `language_version` back to `python3.9`. Locally, I have to do one of the above before I can commit something.

docs/source/how_to_guides/optimization/how_to_specify_algorithm_and_algo_options.rst
```diff
@@ -679,18 +679,24 @@ noisy criterion functions.
...
       the input array *model_improving_points* is not zero.
     - **trustregion_subproblem_solver** (str): Solver to use for the trust-region
       subproblem. Two internal solvers are supported:
       - "bntr": Bounded Newton Trust-Region (default, supports bound constraints)
       - "gqtpar": (does not support bound constraints)
+    - **trustregion_subproblem_conjugate_gradient_step** (str): Routine for computing
+      the conjugate gradient step, when the subsolver "bntr" is used.
+      Available conjugate gradient routines are:
+      - "standard"
+      - "steihaug-toint"
```
janosg, line 692: I would prefer an underscore, i.e. `"steihaug_toint"`.

janosg: It should also be part of the `trustregion_subproblem_options` which I would probably rename to `trustregion_subsolver_options`. Inside that dict in can just be called `conjugate_gradient_step`.

.pre-commit-config.yaml
janosg: I think we can move on to 3.9 in environment.yml.

docs/source/how_to_guides/optimization/how_to_specify_algorithm_and_algo_options.rst
segsell: Will do! I prefer to call it ``conjugate_gradient_routine``, because of a name conflict inside ``minimize_bntr()``. There,``conjugate_gradient_step`` is the actual cg step taken.

janosg: Would `method` work? Routine sounds so much like Fortran.

segsell: Sure! Another candidate would be ``algo``. But I like ``method``, too.

src/estimagic/optimization/subsolvers/_trsbox_quadratic.py
```diff
@@ -0,0 +1,624 @@
...
+                n_fixed_variables,
+            )
+
+            if free_variable_reached_bound:
+                restart_alt_loop = True
+                break
+
+            hess_s = model_hessian @ search_direction
+
+            s_hess_s = np.sum(search_direction[x_bounded == 0] * hess_s[x_bounded == 0])
```
segsell, line 261: Lines ``261``- ``265`` shoould be mathematically equivalent to

```python
s_hess_s = search_direction[x_bounded == 0].T @ hess_s[x_bounded == 
x_hess_s = x_candidate[x_bounded == 0].T @ hess_s[x_bounded == 0]
x_hess_x = x_candidate[x_bounded == 0].T @ hessian_reduced[x_bounded == 0]
```
Numerically, however, there are slight differences in the ``more_wild`` benchmarks, where the ``@`` version performs slightly worse.

src/estimagic/optimization/subsolvers/_trsbox_quadratic.py
```diff
@@ -0,0 +1,624 @@
...
+            x_hess_s = np.sum(x_candidate[x_bounded == 0] * hess_s[x_bounded == 0])
+            x_hess_x = np.sum(
+                x_candidate[x_bounded == 0] * hessian_reduced[x_bounded == 0]
```
segsell, line 264: Lines ``261``- ``265`` shoould be mathematically equivalent to

```python
s_hess_s = search_direction[x_bounded == 0].T @ hess_s[x_bounded == 
x_hess_s = x_candidate[x_bounded == 0].T @ hess_s[x_bounded == 0]
x_hess_x = x_candidate[x_bounded == 0].T @ hessian_reduced[x_bounded == 0]
```
Numerically, however, there are slight differences in the ``more_wild`` benchmarks, where the ``@`` version performs slightly worse.

docs/source/how_to_guides/optimization/how_to_specify_algorithm_and_algo_options.rst
```diff
@@ -679,18 +679,24 @@ noisy criterion functions.
...
       *model_improving_points* is zero.
     - **c2** (int): Treshold for accepting the norm of our current x candidate.
-      Equal to 10 by default. Argument to find_affine_points() in case
+      Equal to 10 by default. Argument to *find_affine_points()* in case
       the input array *model_improving_points* is not zero.
     - **trustregion_subproblem_solver** (str): Solver to use for the trust-region
       subproblem. Two internal solvers are supported:
       - "bntr": Bounded Newton Trust-Region (default, supports bound constraints)
       - "gqtpar": (does not support bound constraints)
+    - **trustregion_subproblem_conjugate_gradient_step** (str): Routine for computing
```
janosg, line 688: I think you forgot to adjust the documentation here. To be sure: `trustregion_subproblem_conjugate_gradient_step` should not be a separate argument but part of the `trustregion_subsolver_options` and called `conjugate_gradient_method` or `bntr_conjugate_gradient_method`.

docs/source/how_to_guides/optimization/how_to_specify_algorithm_and_algo_options.rst
```diff
@@ -679,18 +679,24 @@ noisy criterion functions.
...
+      the conjugate gradient step, when the subsolver "bntr" is used.
+      Available conjugate gradient routines are:
+      - "standard"
+      - "steihaug-toint"
+      - "trsbox" (default)
     - **trustregion_subproblem_options** (dict): Options dictionary containing
```
janosg, line 694: I think this was renamed to `trustregion_subsolver_options`. Please go over the full signature of pounders and check that the docs are up to date.

src/estimagic/optimization/subsolvers/_conjugate_gradient_quadratic.py
```diff
@@ -87,22 +96,22 @@ def _get_distance_to_trustregion_boundary(candidate, direction, radius):
...
 
     Returns:
-        (float) Distance of the candidate vector to the trustregion
+        float: The candidate vector's distance to the trustregion
             boundary.
     """
-    cc = np.dot(candidate, candidate)
-    cd = np.dot(candidate, direction)
-    dd = np.dot(direction, direction)
+    cc = candidate.T @ candidate
```
janosg, line 112: Do you really need the transposes? I guess candidate is a 1d array so `.T` has no effect?

src/estimagic/optimization/subsolvers/_steihaug_toint_quadratic.py
```diff
@@ -0,0 +1,209 @@
...
+        model_gradient (np.ndarray): 1d array of shape (n,) containing the
+            gradient (i.e. linear terms) of the quadratic model.
+        model_hessian (np.ndarray): 2d array of shape (n, n) containing the
+            hessian (i.e .square terms) of the quadratic model.
+        trustregion_radius (float): Radius of the trust-region.
+
+    Returns:
+        np.ndarray: Solution vector of shape (n,).
+    """
+    reason = "Iterating"
```
janosg, line 28: Why do we need the reason if it is not returned? And if we need it, maybe it should be called `status`? Reading `reason = "Iterating"` is not very intuitive.

src/estimagic/optimization/subsolvers/_steihaug_toint_quadratic.py
```diff
@@ -0,0 +1,209 @@
...
+def _take_step_to_trustregion_boundary(x_candidate, p, dp, radius_sq, norm_d, norm_p):
+    """Take step to trust-region boundary."""
+    step = (np.sqrt(dp * dp + norm_p * (radius_sq - norm_d)) - dp) / norm_p
+    x_candidate = x_candidate + step * p
+
+    return x_candidate
+
+
+def _check_convergence(
+    rnorm, rnorm0, abstol, ttol, divtol, converged, diverged, message
```
janosg, line 195: Why does this take message as argument? It looks like the passed in message is not used and just redefined inside the function.

janosg: Ok, I see now. In some cases the message is not changed and just returned. Same for the other things. Maybe add an else clause to make that explicit.

src/estimagic/optimization/subsolvers/_trsbox_quadratic.py
janosg: Then we take the sum.

src/estimagic/optimization/subsolvers/_trsbox_quadratic.py
```diff
@@ -0,0 +1,624 @@
...
+    )
+    tangent = (
+        bound_on_tangent
+        * ((index_angle_greatest_reduction + 1) + 0.5 * raw_reduction)
+        / n_angles
+    )
+    return tangent
+
+
+def _calc_sum_sq(x):
```
janosg, line 622: You don't need the transpose here and it's questionable if this needs to be a function at al. In most cases just `x @ x` is shorter and clearer than calling this function.

src/estimagic/optimization/subsolvers/bounded_newton_quadratic.py
```diff
@@ -190,14 +203,36 @@ def compute_conjugate_gradient_step(
...
     else:
-        step_inactive = minimize_trust_conjugate_gradient(
-            gradient_inactive,
-            hessian_inactive,
-            trustregion_radius,
-            gtol_abs=gtol_abs_conjugate_gradient,
-            gtol_rel=gtol_rel_conjugate_gradient,
-        )
-        step_norm = np.linalg.norm(step_inactive)
+        if conjugate_gradient_method == "standard":
```
janosg, line 206: Maybe `standard` is not a good name because it sounds like this will be the default. I could live with `"cg"`, "`steihaug_toint"` and `"trsbox"` and think it also aligns with the names we use internally.

src/estimagic/optimization/subsolvers/bounded_newton_quadratic.py
```diff
@@ -216,14 +251,31 @@ def compute_conjugate_gradient_step(
...
+                if conjugate_gradient_method == "standard":
+                    step_inactive = minimize_trust_cg(
+                        gradient_inactive,
+                        hessian_inactive,
+                        trustregion_radius,
+                        gtol_abs=gtol_abs_conjugate_gradient,
+                        gtol_rel=gtol_rel_conjugate_gradient,
+                    )
+                    step_norm = np.linalg.norm(step_inactive)
+                elif conjugate_gradient_method == "steihaug-toint":
```
janosg, line 263: Here it's still `steihaug-toint`, above it was `steihaug_toint`. Make sure it's consistent everywhere and I prefer the `_` version.

src/estimagic/optimization/subsolvers/bounded_newton_quadratic.py
```diff
@@ -252,19 +304,19 @@ def compute_predicted_reduction_from_conjugate_gradient_step(
     active_bounds_info,
 ):
     """Compute predicted reduction induced by the Conjugate Gradient step."""
-    if active_bounds_info.all.size > 0:
+    if active_bounds_info.active.size > 0:
```
janosg, line 307: Was this a renaming or an actual bugfix?

tests/optimization/test_linear_subsolvers.py
```diff
@@ -1,14 +1,22 @@
...
-from estimagic.optimization.linear_subsolvers import improve_geomtery_trsbox_linear
-from estimagic.optimization.linear_subsolvers import minimize_trsbox_linear
+from estimagic.optimization.subsolvers.linear_subsolvers import (
+    improve_geomtery_trsbox_linear,
+)
+from estimagic.optimization.subsolvers.linear_subsolvers import minimize_trsbox_linear
 from numpy.testing import assert_array_almost_equal as aaae
 
 
+class LinearModel(NamedTuple):
```
janosg, line 15: Do you redefine the class here? If so, you can just import it!

tests/optimization/test_quadratic_subsolvers.py
```diff
@@ -1,21 +1,35 @@
...
+)
+from estimagic.optimization.subsolvers.quadratic_subsolvers import (
+    minimize_gqtpar_quadratic,
 )
-from estimagic.optimization.quadratic_subsolvers import minimize_bntr_quadratic
-from estimagic.optimization.quadratic_subsolvers import minimize_gqtpar_quadratic
 from numpy.testing import assert_array_almost_equal as aaae
 
 
+class MainModel(NamedTuple):
```
janosg, line 23: Also import. I don't think these classes should be defined in test files.

src/estimagic/optimization/subsolvers/_steihaug_toint_quadratic.py
segsell: Got rid of internal variable `reason` (which is called `message` in the function above).

src/estimagic/optimization/subsolvers/bounded_newton_quadratic.py
segsell: ``active_bounds_info`` is of type ``class(NamedTuple)``. I threw an error moving from ``namedtuple`` to ``NamedTuple``, as``all`` appears to be an internal keyword argument for ``class(NamedTuple)``.

janosg: ok. Thanks!

janosg (approved):
Thanks a lot! This is very important for pounders and tranquilo!

Issue #316, opened 2022-05-06
Improve estimation code


janosg:
### Problems

1. The estimation code does not yet work with general pytrees 
2. I always do the optimization outside of `estimate_ml` and most other users do the same. It just becomes too complicated to specify all relevant inputs for optimization and inference in one function. 
3. Numerical hessians are not yet used in estimation functions and closed form hessians raise an error (as far as I know)

### Proposed solution

1. Deprecate `estimate_ml` and `estimate_msm`
2. Replace them by `do_ml_inference` and `do_msm_inference` which do just the inference part of `estimate_ml` and `estimate_msm`. 
3. Implement a function to generate summary pytrees as described [here](https://estimagic.readthedocs.io/en/latest/development/eep-01-pytrees.html#format-1-everything-becomes-a-dataframe)
3. Add general pytree support to the estimation functions. Most of this will be trivial and the results are already calculated. Just the format needs to be converted.

hmgaudecker:
Well, you tend to have complicated models and know the internals.

Large majority of people will just use defaults no matter what.

Unless there are name clashes and defaults you want to set differently between optimization/inference, I would not depreciate the current stuff but leave it as a shorthand.

janosg:
Ok, I also heard from a few other people now that they like the one step estimation function so we keep that.

timmens:
Done with #330.

PR #330, opened 2022-05-17
Improve estimation code

timmens:
Tackles #316.

---

### TODO

- [x] Add proper testing for estimate_ml
- [x] Add proper testing for estimate_msm
- [x] Add pytree support for estimate_ml
- [x] Add pytree support for estimate_msm
- [x] Implement a function to generate summary pytrees as described [here](https://estimagic.readthedocs.io/en/latest/development/eep-01-pytrees.html#format-1-everything-becomes-a-dataframe)


src/estimagic/estimation/estimate_ml.py
```diff
@@ -56,20 +55,27 @@ def estimate_ml(
...
+        params (pytree): A pytree containing the estimated or start parameters of the
+            likelihood model. If the supplied parameters are estimated parameters, set
+            optimize_options to False. Pytrees can be a numpy array, a pandas Series, a
+            DataFrame with "value" column, a float and any kind of (nested) dictionary
+            or list containing these elements. See :ref:`params` for examples.
+        optimize_options (dict, str or False): Keyword arguments that govern the
+            numerical optimization. Valid entries are all arguments of
             :func:`~estimagic.optimization.optimize.minimize` except for criterion,
-            derivative, criterion_and_derivative and params. If you pass False as
+            criterion_and_derivative and params. If you pass False as
```
janosg, line 70: `criterion_and_derivative` should be allowed.

src/estimagic/estimation/estimate_msm.py
```diff
@@ -27,19 +27,19 @@ def estimate_msm(
...
     optimize_options,
+    *,
     lower_bounds=None,
     upper_bounds=None,
-    *,
     constraints=None,
     logging=False,
     log_options=None,
     simulate_moments_kwargs=None,
+    simulate_moments_and_jacobian=None,
```
janosg, line 37: We could drop `simulate_moments_and_jacobian`. JAX can only exploit synergies for gradients, not jacobians and I don't think any other framework can.

janosg (approved):

src/estimagic/estimation/estimate_ml.py
timmens: They are also passed to the function as ``loglike_and_derivative``. If we allow them in ``optimize_options`` then we need to update ``optimize_options`` instead of simply passing it to maximize.

src/estimagic/estimation/estimate_msm.py
timmens: Okay.

src/estimagic/estimation/estimate_ml.py
timmens: I misunderstood this. It is now resolved.

PR #335, opened 2022-05-27
Improve the result object in `maximize` and `minimize`

### LINKED ISSUE #323: Better result object for optimization

janosg:
closes #323


timmens (approved):
Very nice PR! The new results object makes everything much cleaner. Approved.

docs/source/explanations/optimization/why_optimization_is_hard.ipynb
```diff
@@ -2,7 +2,7 @@
  "cells": [
```
timmens, line 2: <p>You write the algorithm failed, but the output says <code>terminated succesfully</code>.</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/OpenSourceEconomics/estimagic/pull/335/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='12'/>

janosg: That's the point. We know it failed because we know the result but the algorithm thinks it was successful because some convergence criterion is fulfilled. Will try to make it clearer.

src/estimagic/optimization/optimize_result.py
```diff
@@ -0,0 +1,134 @@
...
+
+        return msg
+
+
+def _format_convergence_report(report, algorithm):
+    if report is not None:
+        report = pd.DataFrame.from_dict(report)
+    columns = ["one_step", "five_steps"]
+
+    table = report[columns].applymap(_format_float).astype(str)
```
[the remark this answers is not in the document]
janosg: Thanks!

src/estimagic/optimization/optimize_result.py
```diff
@@ -0,0 +1,134 @@
...
+    stars = pd.cut(
+        sr,
+        bins=[-np.inf, 1e-10, 1e-8, 1e-5, np.inf],
+        labels=["***", "** ", "*  ", "   "],
+    ).astype("str")
+
+    return stars
+
+
+def _format_float(number):
```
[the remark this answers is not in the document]
janosg: It was more complicated than necessary. I simplified and added a docstring.

docs/source/explanations/optimization/why_optimization_is_hard.ipynb
janosg: Ah, I see now. So actually whether the algorithm knows if it fails, depends on the seed. Fixed a seed now.

PR #336, opened 2022-05-28
Add support for pytrees to the dashboard

### LINKED ISSUE #313: Adjust dashboard to pytrees
### Problem 

The dashboard still assumes that the params are a DataFrame and does not work otherwise.

### How it should work

Params DataFrames (or ColumnDataSources in bokeh) are actually an ideal data structure for the dashboard. Instead of working with general pytrees here, we want to do the following:

- Improve the function that generates `params_groups`.
- Read the `flat_params_names` and `flat_params_groups` from the database 
- Read the external parameter vectors from the `optimization_iterations_table`
- Build a DataFrame/ColumnDataSource internally and leave the rest similar to how it is now. 

### Notes

- Now might be a good time to ditch bokeh and move to plotly
- No pytree functions should be used inside the Dashboard. If we need more flattened objects (e.g. flat start_params) we should write them into the log file during the optimization.

roecla:
Closes #313 

- [x] build a start_params DataFrame to be able to use existing dashboard code
- [x] adjust tests to create a fresh database with every run 
- [x] add test case of a criterion function using a mixed type dict in addition to a criterion function using an old-fashioned pandas DataFrame.
- [x] Improve grouping


janosg:
The grouping does not work well if params is just a numpy array. Each parameter gets its own group. I actually assume it is the same for every params format that is not nested. Rest looks very nice.

janosg (approved):
Thanks!

PR #338, opened 2022-05-29
Improve log reading

### LINKED ISSUE #317: Improve log reading
### Problem

- The functions in `read_log.py` still assume that `params` are a DataFrame 
- These functions are not well documented 
- Some tests of logging related functions rely on stored databases. This makes it very hard to adjust the logging code in general 

### Solution

- Rewrite all tests that require databases such that they generate their own databases 
- Adjust the functions in `read_log.py` to work for general pytrees. Functions that return DataFrames where rows are value columns of parameters should probably just return lists of params now. 
- Write documentation.

janosg:
closes #317


timmens (approved):
Very nice PR! I think the simple dataclass log-reader is a good compromise between reducing the number of function imports while not creating large classes.

PR #340, opened 2022-05-31
Add history collection for parallelizing optimizers.

janosg:
The standard way of collecting histories via a list partialled into the internal criterion function does not work for optimizers that evaluate the criterion in parallel. Instead we collect histories inside the batch evaluator. 

I found out that several of the pygmo optimizers do not actually support parallelization. Thus I removed the `n_cores` and `batch_evaluator` arguments and adjusted the documentation.


timmens (approved):
Approved when requested changes are implemented.

PR #343, opened 2022-06-03
Improve results returned by estimation functions

### LINKED ISSUE #337: Return better results object for `estimate_ml` and `estimate_msm`
### Current situation

- `estimate_ml` and `estimate_msm` return a dictionary with information
- The keys are either confusing or not 

- `estimate_ml` and `estimate_msm` should return a custom object (`EstimationResult`).
- It could be designed as follows:

**`estimate_ml`:**

- Attributes
1. params
2. jacobian
3. hessian

- Methods
1. summary
2. cov
3. se

- Properties
1. _summary (calls summary at default)
2. _cov (calls cov at default)
3. _se (...)

And similar for `estimate_msm`.

### Backwards compatibility
We could make the results object fully backwards compatible by implementing __getitem__ to make it behave like a dict. However, I think the userbase is currently small enough that we could simply make a breaking change.

janosg:
closes #337 

To-Do

- [x] Adjust estimate_ml to use new LikelihoodResult
- [x] Adjust estimate_msm to use new MomentsResult
- [x] Make sure full pytree support is implemented in MSM estimation
    - [x] empirical_moments can be a pytree (that must be aligned with output of `simulate_moments`)
    - [x] moments_cov can be a block pytree (adjust `get_moments_cov`)
    - [x] weights can just be "diagonal" or "optimal" for now
- [x] Update documentation notebooks
    - [x] estimate_ml
    - [x] estimate_msm


timmens (approved):
Very nice PR! Approved!

PR #346, opened 2022-06-05
Add option to use nonlinear constraints

timmens:


src/estimagic/parameters/nonlinear_constraints.py
```diff
@@ -0,0 +1,468 @@
...
+import pandas as pd
+from estimagic.differentiation.derivatives import first_derivative
+from estimagic.exceptions import InvalidFunctionError
+from estimagic.parameters.block_trees import block_tree_to_matrix
+from estimagic.parameters.tree_registry import get_registry
+from pybaum import tree_just_flatten
+from pybaum import tree_unflatten
+
+
+CONVERGENCE_ABSOLUTE_CONSTRAINT_TOLERANCE = 1e-5
```
janosg, line 13: I would move this to `algo_options.py` and call it `CONSTRAINTS_ABSOLUTE_TOLERANCE`. The convergence_ prefix is only for convergence criteria of an optimizer, not acceptance of parameter vectors due to constraints.

src/estimagic/parameters/nonlinear_constraints.py
```diff
@@ -0,0 +1,468 @@
...
+"""float: Allowed tolerance of the equality and inequality constraints for values to be
+considered 'feasible'.
+
+"""
+
+# ======================================================================================
+# Process nonlinear constraints
+# ======================================================================================
```
janosg, line 21: We don't need a heading to define a function of the same name.

src/estimagic/parameters/nonlinear_constraints.py
```diff
@@ -0,0 +1,468 @@
...
+    Returns:
+        list[dict]: List of processed constraints.
+
+    """
+    constraint_evals = [
+        _check_validity_and_return_evaluation(c, params, skip_checks)
+        for c in nonlinear_constraints
+    ]
+
+    _processor = partial(
```
janosg, line 59: I would not use partial just because a few arguments are constant if you use the function just once. Maybe use a regular loop instead of list comprehension then to keep it readable.

src/estimagic/parameters/nonlinear_constraints.py
```diff
@@ -0,0 +1,468 @@
...
+
+    # ==================================================================================
+    # Evaluate constraint function and selector
+    # ==================================================================================
+    selected = external_selector(params)
+
+    constraint_func = c["func"]
+
+    if constraint_eval is None:
+        constraint_eval = constraint_func(selected)
```
janosg, line 93: Run one test with `skip_checks=True` to get this line covered.

src/estimagic/parameters/nonlinear_constraints.py
```diff
@@ -0,0 +1,468 @@
...
+
+    # ==================================================================================
+    # Retrieve number of constraints
+    # ==================================================================================
+    if "n_constr" in c:
+        _n_constr = c["n_constr"]
+    else:
+        constraint_value = constraint_func(external_selector(params))
```
janosg, line 101: Isn't this evaluation redundant if `constraint_eval` is not `None`? Would a user ever want to specify `n_constr` if it only buys one less evaluation of the constraint? For least squares optimizers we chose not to support things like `n_residuals` because we thought that the gains are small and the error potential large.

src/estimagic/parameters/nonlinear_constraints.py
```diff
@@ -0,0 +1,468 @@
...
+    else:
+        # use finite-differences if no closed-form jacobian is defined
+        def jacobian(p):
+            return first_derivative(
+                constraint_func,
+                p,
+                **options,
+            )["derivative"]
+
+    def _jacobian_from_internal(x):
```
janosg, line 127: Let's call this `_internal_jacobian`; `_jacobian_from_internal` sounds like it takes some internal jacobian matrix and returns an external one.

src/estimagic/parameters/nonlinear_constraints.py
```diff
@@ -0,0 +1,468 @@
...
+        These can be significantly different from the external parameters, due to
+        estimagic's reparametrization features. In this function we compute the Jacobian
+        of the constraint at the internal parameters using information on the Jacobian
+        of the constraint at the selected external parameters.
+
+        """
+        params = converter.params_from_internal(x)
+        select = external_selector(params)
+        jac = jacobian(select)
+        jac_matrix = block_tree_to_matrix(jac, constraint_eval, selected)
```
janosg, line 141: Is `selected` a typo here and you mean `select` or is this on purpose using the variable defined in line 88?

src/estimagic/parameters/nonlinear_constraints.py
```diff
@@ -0,0 +1,468 @@
...
+        jac_sparse = _extend_jacobian(jac_matrix, selection_indices, n_params)
```
janosg, line 142: Also here, the `selection_indices` are defined quite far away. Is there away which makes it very hard to read this already complicated part of the code.

src/estimagic/parameters/nonlinear_constraints.py
```diff
@@ -0,0 +1,468 @@
...
+
+        # ==============================================================================
+        # Equality constraints
+        #
+        # We define the internal constraint function to be satisfied if it is equal
+        # to zero, by subtracting the fixed value.
+
+        _value = np.atleast_1d(np.array(c["value"], dtype=float))
+
+        def constraint_from_internal(x):
```
janosg, line 161: For same reasons as above I would call this `internal_func` or `internal_constraint_func`. It does not convert a constraint, it is a constraint.

src/estimagic/parameters/nonlinear_constraints.py
```diff
@@ -0,0 +1,468 @@
...
+
+    An equality constaint g(x) = 0 can be transformed to two inequality constraints
+    using (g(x), -g(x)) >= 0. Hence, the number of constraints doubles, and the
+    constraint functions itself as well as the derivative need to be updated.
+
+    """
+    if c["type"] == "eq":
+
+        def transform(x, func):
+            return np.concatenate((func(x), -func(x)), axis=0)
```
janosg, line 233: You can evaluate `func` once and then stack it with its negative. `func` can be slow!

src/estimagic/parameters/nonlinear_constraints.py
```diff
@@ -0,0 +1,468 @@
...
+        jacobian_from_internal = _compose_funcs(
+            _jacobian_from_internal, transformation["derivative"]
+        )
+
+        n_constr = 2 * _n_constr if transformation["name"] == "stack" else _n_constr
+
+    internal_constr = {
+        "n_constr": n_constr,
+        "type": _type,
+        "fun": constraint_from_internal,  # internal name for 'func'
```
janosg, line 203: Can we use `func` instead of `fun`?

src/estimagic/parameters/nonlinear_constraints.py
```diff
@@ -0,0 +1,468 @@
...
+
+def _extend_jacobian(jac_mat, selection_indices, n_params):
+    """Extend Jacobian on selected parameters to full params.
+
+    Jacobian of constraints is defined on a selection of the parameters, however, we
+    need the Jacobian on the full params. Since the Jacobian is trivially zero at the
+    non-selected params we can simply fill a zero matrix.
+
+    """
+    jac_sparse = np.zeros((jac_mat.shape[0], n_params))
```
janosg, line 291: Would call it `jac_extended` until we actually use sparse matrices.

src/estimagic/parameters/nonlinear_constraints.py
```diff
@@ -0,0 +1,468 @@
...
+    jac_sparse[:, selection_indices] = jac_mat
+    return jac_sparse
+
+
+def _get_selection_indices(params, selector):
+    """Get index of selected flat params and number of flat params."""
+    registry = get_registry(extended=True)
+    flat_params = tree_just_flatten(params, registry=registry)
```
janosg, line 299: Use `tree_flatten` to get a `treedef` and then use that `treedef` in `tree_unflatten`. Currently original pytrees can substitute for `treedefs` in pybaum but this will change in the future!

src/estimagic/parameters/nonlinear_constraints.py
```diff
@@ -0,0 +1,468 @@
...
+    if "selector" in c:
+        if not callable(c["selector"]):
+            raise ValueError(
+                f"'selector' entry needs to be callable in constraint {c}."
+            )
+        else:
+            try:
+                c["selector"](params)
+            except Exception:
+                raise InvalidFunctionError(
```
janosg, line 424: Here I am indifferent between `InvalidFunctionError` and `InvalidConstraintError`.

janosg (commented):
Thanks! Great PR and the tricky parts all look good. Comments mainly about style. 

We also need a description of the internal constraints interface in [internal_optimizers.rst](https://github.com/OpenSourceEconomics/estimagic/blob/main/docs/source/explanations/optimization/internal_optimizers.rst).

src/estimagic/parameters/nonlinear_constraints.py
```diff
@@ -0,0 +1,468 @@
...
+        constaint_eval: Evaluation of constraint at params, if skip_checks if False,
+            else None.
+
+    """
+    # ==================================================================================
+    # check functions
+    # ==================================================================================
+
+    if "func" not in c:
+        raise ValueError(
```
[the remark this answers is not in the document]
timmens: Those tests already exist in test_nonlinear_constraints.py::test_check_validity_nonlinear_constraint.

src/estimagic/parameters/nonlinear_constraints.py
timmens: Was not a typo but unnecessary. Fixed now.

src/estimagic/parameters/nonlinear_constraints.py
timmens: Very very good catch!

src/estimagic/parameters/nonlinear_constraints.py
timmens: For the internal interface I chose ``fun`` because most of the optimizers that can handle nonlinear constraints, expect the argument ``fun``.

janosg (approved):

PR #354, opened 2022-06-13
Throw better errors in get_bounds

timmens:


janosg (approved):
Thanks a lot. I really like the notebook!

PR #355, opened 2022-06-13
More tests for constraint processing

janosg:
- [x] Add test where linear constraint, fix and bounds overlap and fix bug in `drop_redundant_linear_constraints`.
- [x] Test that at least one parameter in a increasing constraint can have a lower bound
- [x] Test that too many bounds in an increasing constraint give an error. Increasing constraints cannot have more than one bound. 
- [x] Test that violated constraints give a good error
- [x] Test that constraints are inherited from pairwise equality constraints


src/estimagic/parameters/consolidate_constraints.py
```diff
@@ -644,10 +645,14 @@ def _check_consolidated_weights(weights, param_names):
...
 
     if n_constraints > n_params:
-        raise ValueError(msg_too_many + msg_general.format(relevant_names, weights))
+        raise InvalidConstraintError(
+            msg_too_many + msg_general.format(relevant_names, weights)
+        )
 
     if np.linalg.matrix_rank(weights) < n_constraints:
-        raise ValueError(msg_rank + msg_general.format(relevant_names, weights))
+        raise InvalidConstraintError(
```
timmens, line 653: This line is not covered by any tests. Could you create a simple test case to get this error? If its too much work you can merge.

timmens (approved):

PR #357, opened 2022-06-14
Add jax support

janosg:
- [x] Write jax aware `is_scalar` function
- [x] Use new `isscalar` function everywhere in `estimagic.parameters`
- [x] Release jax compatible pybaum version and add minimum version
    - environment.yml
    - pip requirements 
    - tox
    - meta.yaml when releasing
- [x] Make sure jax tests count into coverage; Possibly use same approach as in pybaum
- [x] Add more tests with jax


timmens (approved):
Great PR, and super good tests! It took a bit longer to review because I wanted to check the pybaum PR again, as we used the class `jaxlib.xla_extension.DeviceArray` there instead of `jax.numpy.ndarray`, but it seems that using the latter works just as well. Also thanks for updating CHANGES.rst.

janosg:
In pybaum we needed `type()`, here we just do `isinstance()`. I remember that I tried it out and what we did in pybaum was actually necessary there.

PR #359, opened 2022-06-17
Update InvalidParamsError message for constraints

timmens:
Print better error message for invalid starting parameters with constraints.


janosg (approved):
Can you also add a test case for this? Should be quick to do it for all constraints.

PR #361, opened 2022-06-20
Update estimation code

timmens:
- [ ] Refactor `flat_params` to `internal_params`
- [ ] Create new variable `free_params`
- [ ] Fix errors in estimation code (do not use converter etc.)
- [ ] Move ``bootstrap_from_outcomes`` code into ``bootstrap`` and extend functionality

### DIFF
--- a/src/estimagic/estimation/estimate_ml.py
+++ b/src/estimagic/estimation/estimate_ml.py
@@ -14,9 +14,11 @@ from estimagic.inference.ml_covs import cov_jacobian
 from estimagic.inference.ml_covs import cov_robust
 from estimagic.inference.ml_covs import cov_strata_robust
 from estimagic.inference.shared import calculate_ci
+from estimagic.inference.shared import calculate_free_estimates
 from estimagic.inference.shared import calculate_inference_quantities
 from estimagic.inference.shared import calculate_p_values
 from estimagic.inference.shared import check_is_optimized_and_derivative_case
+from estimagic.inference.shared import FreeParams
 from estimagic.inference.shared import get_derivative_case
 from estimagic.inference.shared import transform_covariance
 from estimagic.optimization.optimize import maximize
@@ -25,6 +27,7 @@ from estimagic.parameters.block_trees import block_tree_to_matrix
 from estimagic.parameters.block_trees import matrix_to_block_tree
 from estimagic.parameters.conversion import Converter
 from estimagic.parameters.conversion import get_converter
+from estimagic.parameters.space_conversion import InternalParams
 from estimagic.shared.check_option_dicts import check_numdiff_options
 from estimagic.shared.check_option_dicts import check_optimization_options
 from estimagic.utilities import to_pickle
@@ -229,8 +232,9 @@ def estimate_ml(
     # ==================================================================================
 
     if jac_case == "closed-form":
-        x = converter.params_to_internal(estimates)
-        int_jac = converter.derivative_to_internal(jacobian_eval, x)
+        int_jac = converter.derivative_to_internal(
+            jacobian_eval, internal_estimates.values
+        )
     elif jac_case == "numerical":
 
         def func(x):
@@ -324,6 +328,8 @@ def estimate_ml(
     # create a LikelihoodResult object
     # ==================================================================================
 
+    free_estimates = calculate_free_estimates(estimates, internal_estimates)
+
     res = LikelihoodResult(
         params=estimates,
         _converter=converter,
@@ -335,7 +341,8 @@ def estimate_ml(
         _internal_jacobian=int_jac,
         _internal_hessian=int_hess,
         _design_info=design_info,
-        _flat_params=internal_estimates,
+        _internal_estimates=internal_estimates,
+        _free_estimates=free_estimates,
         _has_constraints=constraints not in [None, []],
     )
 
@@ -345,7 +352,8 @@ def estimate_ml(
 @dataclass
 class LikelihoodResult:
     params: Any
-    _flat_params: Any
+    _internal_estimates: InternalParams
+    _free_estimates: FreeParams
     _converter: Converter
     _has_constraints: bool
     optimize_result: Union[OptimizeResult, None] = None
@@ -391,8 +399,6 @@ class LikelihoodResult:
         internal_jac = self._internal_jacobian
         internal_hess = self._internal_hessian
         design_info = self._design_info
-        converter = self._converter
-        flat_params = self._flat_params
 
         if method == "jacobian":
             int_cov = cov_jacobian(internal_jac)
@@ -414,9 +420,9 @@ class LikelihoodResult:
         np.random.seed(seed)
 
         free_cov = transform_covariance(
-            flat_params=flat_params,
+            internal_params=self._internal_estimates,
             internal_cov=int_cov,
-            converter=converter,
+            converter=self._converter,
             n_samples=n_samples,
             bounds_handling=bounds_handling,
         )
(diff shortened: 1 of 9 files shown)


src/estimagic/inference/bootstrap_helpers.py
```diff
@@ -1,32 +1,35 @@
...
+def check_inputs(
+    data=None, cluster_by=None, ci_method="percentile", ci_level=0.95, skipdata=False
+):
     """Check validity of inputs.
 
     Args:
         data (pd.DataFrame): original dataset.
         cluster_by (str): column name of variable to cluster by.
         ci_method (str): method of choice for confidence interval computation.
         alpha (float): significance level of choice.
```
timmens, line 13:
```suggestion
        alpha (float): significance level of choice.
        skipdata (bool): Whether to skip all checks on the data argument.
```

src/estimagic/estimation/estimate_ml.py
```diff
@@ -14,18 +15,22 @@
...
+from estimagic.inference.shared import calculate_estimation_summary
+from estimagic.inference.shared import calculate_free_estimates
 from estimagic.inference.shared import calculate_p_values
 from estimagic.inference.shared import check_is_optimized_and_derivative_case
-from estimagic.inference.shared import convert_flat_params_to_pytree
+from estimagic.inference.shared import FreeParams
 from estimagic.inference.shared import get_derivative_case
 from estimagic.inference.shared import transform_covariance
+from estimagic.inference.shared import transform_free_cov_to_cov
+from estimagic.inference.shared import transform_free_values_to_params_treedef
```
janosg, line 26: why `params_treedef` and not `params_tree`. In the long run, `treedef` in pybaum will be a custom object, not just the original pytree.

src/estimagic/estimation/estimate_ml.py
```diff
@@ -727,15 +729,15 @@ def p_values(
...
         )
 
-        free_values = self._flat_params.values[self._flat_params.free_mask]
-        free_se = np.sqrt(np.diagonal(free_cov))
-        free_p_values = calculate_p_values(free_values, free_se)
-
-        out = convert_flat_params_to_pytree(
-            free_p_values, self._flat_params, self._converter
+        free_p_values = calculate_p_values(
+            flat_values=self._free_estimates.values,
```
janosg, line 733: Maybe also rename to `free_values` and `free_standard_errors`

src/estimagic/inference/bootstrap.py
```diff
@@ -17,9 +18,10 @@
 
 
 def bootstrap(
-    data,
     outcome,
     *,
+    data=None,
+    existing_outcomes=None,
```
janosg, line 24: I think we have some notebooks in the documentation that use `bootstrap_from_outcomes`. @segsell can you check and adjust them? You should be most familiar because you just did the adjustment to the new results object.

src/estimagic/inference/bootstrap.py
```diff
@@ -51,88 +57,70 @@ def bootstrap(
...
+            data=data,
+            outcome=outcome,
+            cluster_by=cluster_by,
+            seed=seed,
+            n_draws=n_draws,
+            n_cores=n_cores,
+            error_handling=error_handling,
+            batch_evaluator=batch_evaluator,
+        )
+        existing_outcomes += bootstrap_outcomes
```
janosg, line 82: I think this would change `existing_outcomes` in place. Also, I suggest to give the combined outcomes a name that reflects that they are already combined. i.e. something like `all_outcomes = existing_outcomes + bootstrap_outcomes`

src/estimagic/inference/shared.py
```diff
@@ -100,16 +110,17 @@ def calculate_inference_quantities(estimates, internal_estimates, free_cov, ci_l
...
     registry = get_registry(extended=True)
 
-    df = pd.DataFrame(index=internal_estimates.names)
+    df = pd.DataFrame(index=free_estimates.all_names)
     df["value"] = tree_just_flatten(estimates, registry=registry)
-    df["free"] = internal_estimates.free_mask
+    df["free"] = free_estimates.free_mask
     df.loc[free_cov.index, "standard_error"] = np.sqrt(np.diag(free_cov))
 
     df["p_value"] = calculate_p_values(
```
janosg, line 126: I suggest we use a different approach to building the DataFrame. 

1. Add `estimation_result` as argument. This can be a LieklihoodResult, MomentsResult or BootstrapResult
2. Instead of calculating stuff, call the `ci`, `se`, `p_values`, ....  methods of `estimation_result` 
3. Call `tree_just_flatten` on each of them
4. Combine them into a DataFrame
5. Continue as before

That way we get better re-use between bootstrap and the others and also more code sharing between the different formats in which you can get CIs, p_values, ...

src/estimagic/inference/bootstrap.py
timmens: Very good catch! Much better now.

tests/estimation/test_estimate_msm.py
```diff
@@ -176,3 +176,24 @@ def test_to_pickle(tmp_path):
     )
 
     calculated.to_pickle(tmp_path / "bla.pkl")
+
+
+def test_caching():
```
janosg, line 181: To test caching we should use an identity check (via `is`), not almost equal with very high precision.

janosg (commented):
This is a huge step! Very nice changes. 

I made a few small comments that need to be addressed before merging. 

And there is one more thing: The name `flat_params` is still widely used for the second output of `get_converter`. Given our recent experience, I think this should be renamed to `internal_params`.

src/estimagic/inference/bootstrap.py
```diff
@@ -246,7 +235,11 @@ def p_values(self):
             Any: A pytree with the same structure as base_outcome containing p-values
                 for the parameter estimates.
         """
-        raise NotImplementedError("Bootstrapped p-values are not implemented yet.")
+        msg = (
+            "Bootstrap p-values are not implemented yet, due to missing p-values. You"
```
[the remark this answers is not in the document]
janosg: Done.

timmens:
The new test cases are great, thanks a lot! Is there anything more that needs to be done @janosg?

src/estimagic/inference/bootstrap.py
```diff
@@ -17,9 +19,10 @@
 
 
 def bootstrap(
-    data,
     outcome,
     *,
+    data=None,
+    existing_outcomes=None,
     outcome_kwargs=None,
     n_draws=1_000,
```
janosg, line 27: Currently the total number of draws is `n_draws + len(existing_outcomes)`, but it is not clearly documented in the docstring.

1. We should document the precise meaning of `n_draws` in the docstring
2. I think it would be if `n_draws` is the number of draws used. If existing outcomes is longer than `n_draws` only the first part of `existing_outcomes` should be used. Otherwise, `n_draws - len(existing_outcomes)` new draws should be sampled.

src/estimagic/estimation/estimate_ml.py
```diff
@@ -389,40 +399,28 @@ def _get_free_cov(
...
-            int_cov = cov_cluster_robust(
-                jac=internal_jac, hess=internal_hess, design_info=design_info
-            )
-        elif method == "strata_robust":
-            int_cov = cov_strata_robust(
-                jac=internal_jac, hess=internal_hess, design_info=design_info
-            )
+            msg = f"Invalid method: {method}. Valid methods are {self._valid_methods}."
+            raise ValueError(msg)
+        seed = self._seed if seed is None else seed
```
janosg, line 404: Instead of this, I think we should set a default seed for every method that needs a seed. If someone wants to pass `seed=None` this should be possible and actually use no seed instead of using the default seed. 

Of course, if seed is None, no result should be chached.

timmens: Use `estimagic.config.DEFAULT_SEED`

janosg: I think this line still needs to go! In general, `self._seed` should not exist.

janosg (approved):

PR #364, opened 2022-06-24
Use local random number generators

timmens:
**Old behavior:** Seeding was done by calling `np.random.seed` with a specific seed.

**Problem:** This call has side-effects and is no longer recommend to use by numpy

**Fix:** We use `np.random.default_rng` to create local random number generators when needed.

**Blogpost:**: https://albertcthomas.github.io/good-practices-random-number-generators/

**This PR:**
- [x] Create a utility function that creates a generator instance given user input
- [x] Pass random number generators instead of seeds through functions 
- [x] Replace `np.random.seed` by an appropriate use of `np.random.default_rng`
- [x] Adjust random behavior in `estimate_{ml, msm}` functions:
* Use `seed=None` as default
* Print warning if `seed=None` and output is random
- [x] Update documentation


src/estimagic/utilities.py
```diff
@@ -304,3 +304,23 @@ def isscalar(element):
...
+
+    seed (Union[None, int, numpy.random.Generator]): If seed is None or int the
+        numpy.random.default_rng is used seeded with seed. If seed is already a
+        Generator instance then that instance is used.
+
+    Returns:
+        numpy.random.Generator: The random number generator.
+
+    """
+    if isinstance(seed, np.random.Generator):
```
janosg, line 320: We need tests that:
- call `get_rng` with a `np.random.Generator`
- call `get_rng` with something invalid.

tests/inference/test_bootstrap_samples.py
```diff
@@ -1,6 +1,7 @@
 import numpy as np
 import pandas as pd
 import pytest
+from estimagic.config import DEFAULT_SEED
```
janosg, line 4: I think we should completely remove the default seed and use whatever you want in tests.

tests/inference/test_shared.py
```diff
@@ -16,6 +17,9 @@
 from pybaum import tree_equal
 
 
+RNG = np.random.default_rng(DEFAULT_SEED)
```
janosg, line 20: Why do you use a global RNG here and not a new one for each test case?

tests/optimization/test_tao_optimizers.py
```diff
@@ -79,10 +82,10 @@ def test_max_iters():
 
 
 def test_grtol():
-    np.random.seed(5474)
-    true_params = get_random_params(2, 0.3, 0.4, 0, 0.3)
+    rng = np.random.default_rng(5474)
```
janosg, line 85: Why not use `get_rng` here? Would be nice to be consistent.

janosg (commented):
Looks very nice.

tests/inference/test_shared.py
janosg: The same is done in a few other places

janosg (approved):

Issue #341, opened 2022-06-01
Add copy buttons to documentation


amageh:
### What would you like to enhance and why?

I think it would be nice to add copy buttons like [this](https://sphinx-copybutton.readthedocs.io/en/latest/) to the documentation so it is easier to copy code snippets etc.

### I would add them to:

- [Installation](https://estimagic.readthedocs.io/en/stable/getting_started/installation.html)
- [Algorithms guide](https://estimagic.readthedocs.io/en/stable/how_to_guides/optimization/how_to_specify_algorithm_and_algo_options.html) (i.e. to copy the name of algorithms)

Anyone feel free to add points in the documentation where this makes sense, I will open a PR to implement it.

janosg:
Thanks, Annica! This is a great idea.

For installation we should add a copy button once at the main installation instruction but then also for each optimizer that requires an optional dependency. 

Additional places:
- The command to start the dashboard 
- Self contained examples on how to use constraints that can then be adjusted by users

janosg:
Outdated.

PR #365, opened 2022-06-26
Add copy buttons for documentation

amageh:
Implements suggestions from issue #341. The package [sphinx-copybutton](https://sphinx-copybutton.readthedocs.io/en/latest/index.html#) adds buttons to all code blocks in the documentation, so I added code blocks for the algorithms and installation page.


janosg (commented):
Very nice. Thanks a lot!

Could you add quotes around the algorithms? i.e. `"scipy_lbfgsb"` instead of `scipy_lbfgsb`. I think that would make it more convenient if you really want to use the copied code.

PR #367, opened 2022-07-04
Fix pytest-linux environment

timmens:


janosg (approved):
Thanks! Good solution for now. In the long run we probably need to reach out to the maintainers of `petsc4py` to find a better fix.

PR #371, opened 2022-07-27
Prepare POUNDERS for larger refactoring

segsell:
Do some basic refactoring of POUNDERS to prepare for some more comprehensive changes.

To-Do:
- [x] Refactoring
- [x] Make sure benchmark results of the More & Wild benchmark set do not change
- [x] Rename variables


janosg (approved):
Thanks. We can merge this as soon as you made sure that the benchmark results do not change after this change.

PR #375, opened 2022-08-09
Add pounders fitting

segsell:
Add the POUNDERS fitting method to tranquilo.

### DIFF
--- a/src/estimagic/optimization/tranquilo/fit_models.py
+++ b/src/estimagic/optimization/tranquilo/fit_models.py
@@ -6,6 +6,7 @@ import numpy as np
 from estimagic.optimization.tranquilo.models import ModelInfo
 from estimagic.optimization.tranquilo.models import VectorModel
 from numba import njit
+from scipy.linalg import qr_multiply
 
 
 def get_fitter(fitter, user_options=None, model_info=None):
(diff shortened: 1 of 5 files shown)


tests/optimization/tranquilo/test_adjust_radius.py
```diff
@@ -70,7 +70,7 @@ def test_min_radius_is_not_violated(options):
         options=options,
     )
 
-    expected = 1e-6
+    expected = 7.5e-7
```
janosg, line 73: This does not test that min radius is not violated anymore because the min radius is not binding. We need to start with a smaller radius or not use default options to have a meaningful test.

src/estimagic/optimization/tranquilo/fit_models.py
```diff
@@ -233,16 +233,114 @@ def _fit_ridge(x, y, penalty):
...
+):
+    n_residuals = y.shape[1]
+
+    n_poly_features = n_params * (n_params + 1) // 2
+    offset = 0 if has_intercepts else 1
+
+    coef = np.empty((n_residuals, has_intercepts + n_params + n_poly_features))
+
+    # just-identified case
+    if n_samples == (n_params + 1):
```
janosg, line 297: define a variable `_is_just_identified`; use it here and in the loop.

src/estimagic/optimization/tranquilo/fit_models.py
```diff
@@ -233,16 +233,114 @@ def _fit_ridge(x, y, penalty):
...
+        omega = np.zeros(n_params)
```
janosg, line 298: I think we need better names for omega, beta and alpha. Rest can stay close to the paper, but these three seem to be fundamental to understand what's going on.

janosg (approved):

PR #379, opened 2022-08-23
Improve estimation table

### LINKED ISSUE #373: Two proposals for estimation_table
* estimagic version used, if any: `main` branch
* Python version, if any: 3.9.10
* Operating System: Windows

### What would you like to enhance and why? Is it related to an issue/problem?

Transitioning to the new estimagic tables worked out very well, thanks! But two issues appeared:

- I didn't find an easy way to add a row of strings (e.g. "Controls": ["Yes", "No"]) to the footer before calling `render_latex`.  I think it would be a good feature if the function would allow for these simple strings and wrap them as `f"\\multicolumn{{1}}{{c}}{{{sr[i]}}}"` automatically. Alternatively, I could add the multicolumn environments as the user myself. But this didn't work because an additional bracket was added by the function. I started to implement the former solution on the branch `improve_estimation_table` and can make it nicer in case you aggree with the general idea.
- The row showing observations is now also rounded (in my case to two significant digits), but I would like to not have it rounded -- preferably not only if it is called "observations" (maybe even for all integer rows). This feature was implemented in a previous version of estimagic, but seems to be missing now. What are your thoughts about it?

ChristianZimpelmann:
- Allow to easily add a row of strings like `"Controls": ["Yes", "No"])` to the footer before calling `render_latex`.
- Do not round all cells which contain integer values. Before this PR, e.g. the number of observations were also rounded to two significant digits.

Closes #373.


ChristianZimpelmann:
Two open questions remain (for now):

- as `_center_align_integers_and_non_numeric_strings` is currently set up, a string "Total 2.2 observations" in a cell, would be treated as a float (not center-aligned) and also the first part "Total " would be stripped of. I think we should adjust this, but wanted to discuss this first as I might miss a reason why you came up with this solution in the first place.

- I feel like the code block below can somehow be simplified, i.e. without looping over the columns. But didn't find a solution.
```code
    for c in df_formatted:
        df_formatted.loc[position_of_integers[c], c] = df_raw.loc[
            position_of_integers[c], c
        ]
```

hmgaudecker:
Thanks! While you are at it, can you please turn off the performance warnings, which are irrelevant for tables:

```
╭────────────────────────────────────────────────────────────────── Warnings ───────────────────────────────────────────────────────────────────╮
│ task_reg_invest_on_groups_or_params.py::task_reg_invest_on_groups                                                                             │
│ task_reg_invest_on_groups_or_params.py::task_reg_invest_on_groups                                                                             │
│ task_reg_invest_on_groups_or_params.py::task_reg_invest_on_groups                                                                             │
│ task_reg_invest_on_groups_or_params.py::task_reg_invest_on_groups                                                                             │
│ task_reg_invest_on_groups_or_params.py::task_reg_invest_on_groups                                                                             │
│ ... in 15 more locations.                                                                                                                     │
│     /x/estimagic/estimagic/src/estimagic/visualization/estimation_table.py:296: PerformanceWarning: indexing past lexsort depth may    │
│     impact performance.                                                                                                                       │
│       ci_in_body = body.loc[("",)][body.columns[0]].str.contains(";").any()                                                                   │
│                                                                                                                                               │
│ ♥ https://pytask-dev.rtdf.io/en/stable/how_to_guides/capture_warnings.html                                                                    │
╰───────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╯
```

hmgaudecker:
Thanks. Please make sure that when you call a test `test_render_latex()`, it actually calls the function `render_latex`. The test that is called this way should be `test_center_align_integers_and_non_numeric_strings`

While the bug we saw today seems fixed in that function, it has no effect on `render_latex` because the loop  in lines 349-350 of estimation_table is flawed:

```py
        for _, r in footer.iterrows():
            r = _center_align_integers_and_non_numeric_strings(r)
```

The (correct) result on the RHS is assigned to `r`, which just disappears in the next iteration.

I tried to fix with
```py
        for idx, row in footer.iterrows():
            footer[idx] = _center_align_integers_and_non_numeric_strings(row)
```
but this gives an error

```pytb
TypeError: incompatible index of inserted column with frame index
``` 
in my application.

ChristianZimpelmann:
Reminder: At some point, we should make sure that columns consisting of integer values only are right aligned (instead of center-aligned)

ChristianZimpelmann:
@janosg, I just merged main and updated the first comment in this PR.

The remaining comments are:

> - How should the interaction of treatment of integers with `add_trailing_zeros` be? Currently implemented: In body, trailing zeros are added to integers. In footer, no trailing zeros are added. I don't have a strong opinion here, as I am not planning to use `add_trailing_zeros=True`.

My suggestion: leave as is

> - Reminder: At some point, we should make sure that columns consisting of integer values only are right aligned (instead of center-aligned)

If I am not mistaken, this comments was brought forward by Hans-Martin. I suggest opening an issue and not implementing it here as thinking about comparisons over full columns will most likely need a deeper thought.

hmgaudecker:
> > How should the interaction of treatment of integers with `add_trailing_zeros` be? Currently implemented: In body, trailing zeros are added to integers. In footer, no trailing zeros are added. I don't have a strong opinion here, as I am not planning to use `add_trailing_zeros=True`.
> 
> My suggestion: leave as is

Given that the docstring reads: 

>  add_trailing_zeros (bool): If True, format floats such that they have same number of digits after the decimal point. Default True.

I would be very surprised if (under the default!) suddenly all integers showed up as floats. Or do you mean by interaction  "columns that contain both integers and floats" ? And by footer the number of observations or so? In that case I would not have a strong opinion, we don't really support tables containing all kinds of different numbers yet, anyhow.

> > Reminder: At some point, we should make sure that columns consisting of integer values only are right aligned (instead of center-aligned)
> 
> If I am not mistaken, this comments was brought forward by Hans-Martin. I suggest opening an issue and not implementing it here as thinking about comparisons over full columns will most likely need a deeper thought.

I guess I do not quite understand why we can easily center-align columns but not right-align them?

hmgaudecker:
> > Given that the docstring reads:
> > > add_trailing_zeros (bool): If True, format floats such that they have same number of digits after the decimal point. Default True.
> > 
> > 
> > I would be very surprised if (under the default!) suddenly all integers showed up as floats. Or do you mean by interaction "columns that contain both integers and floats" ? And by footer the number of observations or so? In that case I would not have a strong opinion, we don't really support tables containing all kinds of different numbers yet, anyhow.
> 
> No, I really mean that integers in any of the model outputs are shown as floats. (I guess usually those coefficients are just not integers). But this can easily be changed.
> 
> ```python
> def _get_models_multiindex():
>     df = pd.DataFrame(
>         data={}, columns=["value", "ci_lower", "ci_upper", "p_value"]
>     )
>     df.loc[0] = [0.23123, 0, 1, 1]
>     df.loc[1] = [1, 0, 1, 1]
>     df.loc[2] = [2, 0, 1, 1]
>     df.index = pd.MultiIndex.from_tuples(
>         [("p_1", "v_1"), ("p_1", "v_2"), ("p_2", "v_2")]
>     )
>     
>     info = {"n_obs": 400}
>     mod1 = {"params": df, "info": info, "name": "m1"}
>     mod2 = {"params": df, "info": info, "name": "m2"}
>     models = [mod1, mod2]
>     return models
> 
> models = _get_models_multiindex()
> out = et.estimation_table(
>         models, return_type="render_inputs", confidence_intervals=True
>     )
> out["body"]
> ```
> 
> leads to
> 	m1 	m2
> ('p_1', 'v_1') 	0.231$^{ }$ 	0.231$^{ }$
> ('p_1', '') 	(0.000;1.000) 	(0.000;1.000)
> ('p_1', 'v_2') 	1.000$^{ }$ 	1.000$^{ }$
> ('p_1', '') 	(0.000;1.000) 	(0.000;1.000)
> ('p_2', 'v_2') 	2.000$^{ }$ 	2.000$^{ }$
> ('p_2', '') 	(0.000;1.000) 	(0.000;1.000)

Ah, sorry, seems like I was mistaken in my assumptions.

When talking about integer cells, I was thinking about columns with (Pandas) dtype object, which may contain integer cells:

```ipython
In [8]: x = pd.Series(["a", 1, 0.01])

In [9]: type(x[1])
Out[9]: int
```

Seems like you are thinking about the case where we have all-float columns, some of which may happen to look like integers. I agree these cases should not be treated in any special way.

hmgaudecker:
> > I guess I do not quite understand why we can easily center-align columns but not right-align them?
> 
> Have I misunderstood your comment that you would like to check the whole column whether it only consists of integers? This is something we do not do at the moment. We currently do not center-align whole columns, but only check and align single cells. Switching from center-align to right-align for those specific cells wouldn't be a problem. But I am not sure whether this really is what we want in columns in which we have floats in the main part of the body and then n_observations at the end.

Apologies, I had not looked at the code and did not realise that this is fully internal, i.e., it cannot be controlled by the user. This is something that should change eventually (and we may prepare for it in the code by making two functions `_center_align_integers` and `_center_align_non_numeric_strings` out of the one function we have), but fine to leave as is for the moment IMHO.

ChristianZimpelmann:
Perfect. I opened #451.

Janos, if you agree, I think we can merge.

hmgaudecker:
With "We may prepare for it in the code ..." I meant "make the preparations now" :laughing: 

In the issue, that line will be misleading to future developers, I believe.

ChristianZimpelmann:
Alright, that makes sense. I deleted it in the issue. 

I think it isn`t worth right now to split up the function before we know exactly which changes we are planning to make.

The failing tests seem to be unrelated to this PR.

janosg:
Hi @ChristianZimpelmann, thanks a lot for finishing this up. A few questions:

> "Total 2.2 observations" in a cell, would be treated as a float (not center-aligned) and also the first part "Total " would be stripped of.

Is this still the case? I don't think there ever was an intentional choice to have it and it is just a bug and we should fix it before merging. 

> Allow to easily add a row of strings like "Controls": ["Yes", "No"]) to the footer before calling render_latex

Is the workflow for this documented, ideally with an example? I think this is a very important feature and it should be documented before we merge. 

> Thanks! While you are at it, can you please turn off the performance warnings, which are irrelevant for tables

Is this done? Should not take long and should be done before we merge.

> Thanks. Please make sure that when you call a test test_render_latex(), it actually calls the function render_latex. The test that is called this way should be test_center_align_integers_and_non_numeric_strings ... [see comment by HM]

Is the bug mentioned in this comment by HM fixed?

> I had not looked at the code and did not realise that this is fully internal, i.e., it cannot be controlled by the user. This is something that should change eventually

While I agree that the default visualization of integer only columns should change (see #451), I am always hesitant to add yet another argument to give more control to the user. The whole philosophy of `estimation_table` is to give you something great by default and if you really need absolutely fine control, let you modify the generated DataFrame before it is rendered to latex or html. 

Sorry that I ask for more changes, but I think we need to make sure that no bug that was uncovered during this PR slips through the review.

janosg:
I think to really improve `estimation_table` it would be good to have a set of tables that need to look good out of the box. This could be a separate repo that uses pytask to create the tables and compiles them into one pdf / html document. 

Do we have a volunteer to set this up / contribute examples? I think it should be a mix of real world examples and hand crafted examples that contain elements that do not look good in the current version. 

Once we have that, I would be more motivated to spend a day on cleaning up the code and making sure everything looks nice. 

@timmens could you look into the failing tests on Monday? I agree that they are not related to the current PR but it looks like they fall in your area of expertise.

hmgaudecker:
> The whole philosophy of estimation_table is to give you something great by default and if you really need absolutely fine control, let you modify the generated DataFrame before it is rendered to latex or html.

I guess I tend to forget because so far, I have only used it in projects where core developers of this feature were involved and things ended up being fixed in PRs like this one :-)

Would also be great to document how that would work in practice. *(not that I checked the website right now...)*

ChristianZimpelmann:
Thanks!

> > "Total 2.2 observations" in a cell, would be treated as a float (not center-aligned) and also the first part "Total " would be stripped of.
> 
> Is this still the case? I don't think there ever was an intentional choice to have it and it is just a bug and we should fix it before merging.

No, this was fixed in this PR.

 
> > Allow to easily add a row of strings like "Controls": ["Yes", "No"]) to the footer before calling render_latex
> 
> Is the workflow for this documented, ideally with an example? I think this is a very important feature and it should be documented before we merge.

I am afraid it is not. Would you document it in a docstring or the Notebook?
> 
> > Thanks! While you are at it, can you please turn off the performance warnings, which are irrelevant for tables
> 
> Is this done? Should not take long and should be done before we merge.

I don't see where it is turned off. But I don't get any performance warning when running `pytest .\tests\visualization\test_estimation_table.py`. Or when would you expect those warnings?

> > Thanks. Please make sure that when you call a test test_render_latex(), it actually calls the function render_latex. The test that is called this way should be test_center_align_integers_and_non_numeric_strings ... [see comment by HM]
> 
> Is the bug mentioned in this comment by HM fixed?

Yes!

> > I had not looked at the code and did not realise that this is fully internal, i.e., it cannot be controlled by the user. This is something that should change eventually
> 
> While I agree that the default visualization of integer only columns should change (see #451), I am always hesitant to add yet another argument to give more control to the user. The whole philosophy of `estimation_table` is to give you something great by default and if you really need absolutely fine control, let you modify the generated DataFrame before it is rendered to latex or html.
> 
> Sorry that I ask for more changes, but I think we need to make sure that no bug that was uncovered during this PR slips through the review.

> Do we have a volunteer to set this up / contribute examples? I think it should be a mix of real world examples and hand crafted examples that contain elements that do not look good in the current version.

I could make a first draft of it next week.

janosg:
The documentation should be [here](https://estimagic.readthedocs.io/en/stable/how_to_guides/miscellaneous/how_to_generate_publication_quality_tables.html). Nobody reads the docstring. 

The Performance warnings are probably only triggered when you have params with a MultiIndex. But it would not hurt to ignore these warnings preemptively for the entire estimation table function.

Thanks a lot for the PR and for setting up the repo!

@hmgaudecker I would say it's documented [here](https://estimagic.readthedocs.io/en/stable/how_to_guides/miscellaneous/how_to_generate_publication_quality_tables.html) but I am very open to any suggestions to improve this documentation.

src/estimagic/visualization/estimation_table.py
```diff
@@ -136,91 +138,94 @@ def estimation_table(
...
-            show_footer=show_footer,
-            append_notes=append_notes,
-            notes_label=notes_label,
-            significance_levels=significance_levels,
-            custom_notes=custom_notes,
-            siunitx_warning=siunitx_warning,
-            show_index_names=show_index_names,
-            show_col_names=show_col_names,
-            escape_special_characters=escape_special_characters,
+    with warnings.catch_warnings():
```
ChristianZimpelmann, line 141: > But it would not hurt to ignore these warnings preemptively for the entire estimation table function.

@janosg, is this the implementation you had in mind?

janosg: I would have renamed the current `estimation_table` to `_estimation_table` and defined a new `estimation_table` that calls `_estimation_table` in this catch warnings context. Otherwise we get a lot of indentation.

janosg (approved):
Thanks again, @ChristianZimpelmann! I think we can merge!

src/estimagic/visualization/estimation_table.py
hmgaudecker: Or decorate it? https://numpy.org/devdocs/reference/generated/numpy.testing.suppress_warnings.html

janosg: Very good idea! We should do that!

ChristianZimpelmann: sounds good!

ChristianZimpelmann: Thanks!

ChristianZimpelmann: I implemented it using a decorator in the last [commit](https://github.com/OpenSourceEconomics/estimagic/pull/379/commits/a32f69c2c37855d2a63b000402e44e249b3beffa).

Do you have any more suggestions? Otherwise, we can merge.

janosg: From my side all good!

ChristianZimpelmann:
One test is still (or again?) failing.. unrelated to changes in this PR as far as I understand.

Should I just ignore it?

janosg:
We are working on these random failures. Yes, it's unrelated. I'll merge.

PR #388, opened 2022-10-11
Improve trust region sampling in tranquilo

timmens:
In this PR we want to improve the trust region sampling of Tranquilo. In particular, we implement

### Sphere sampling
- Naive (random points on the sphere)
- Optimal (points on the sphere placed to maximize the minimum distance between two points)

### Cube sampling
- Naive (random points on the cube)
- Optimal (points on the sphere placed to maximize the minimum distance between two points)

If binding bounds are present we map the shape (sphere or cube) into the feasible trustregion, which is defined as the intersection of the bounds and the trustregion.


src/estimagic/optimization/tranquilo/sample_points.py
```diff
@@ -110,47 +117,207 @@ def _naive_sampler(
...
         size=(n_points, n_params),
     )
+    return points
+
+
+def _hull_sampler(
+    trustregion,
+    target_size,
+    rng,
+    ord,  # noqa: A002
```
janosg, line 139: We should rename `ord` to `order` to not shadow the builtin `ord`.

src/estimagic/optimization/tranquilo/sample_points.py
```diff
@@ -110,47 +117,207 @@ def _naive_sampler(
...
 
-    if bounds is not None and (bounds.lower is not None or bounds.upper is not None):
-        bounds = _get_effective_bounds(trustregion, bounds)
-        points = np.clip(
-            points,
-            a_min=bounds.lower,
-            a_max=bounds.upper,
-        )
+    if algo_options is None:
+        algo_options = {"stopping_max_iterations": 3}
```
janosg, line 219: If algo options are given, we should check if they contain max iterations and else set our default here.

src/estimagic/optimization/tranquilo/sample_points.py
```diff
@@ -110,47 +117,207 @@ def _naive_sampler(
...
+
+    if existing_xs is not None:
+        # map existing points into unit space for easier optimization
+        existing_xs = (existing_xs - trustregion.center) / trustregion.radius
+
+    # start params
+    x0 = rng.normal(size=(n_points, n_params))
+    x0 = _project_onto_unit_hull(x0, ord=ord)
+
+    res = em.maximize(
```
janosg, line 231: I would use scipy minimize to save the overhead. Since you use a 2d numpy array as params we don't even take the fast path and treat params as pytree.

src/estimagic/optimization/tranquilo/sample_points.py
```diff
@@ -110,47 +117,207 @@ def _naive_sampler(
...
+        slc = slice(0, -n_existing_pairs) if n_existing_pairs else slice(None)
+    else:
+        sample = x
+        slc = slice(None)
+
+    dist = pdist(sample) ** 2  # pairwise squared distances
+    dist = dist[slc]  # drop distances between existing points as they should not
+    # influence the optimization
+
+    crit_value = -logsumexp(-dist)  # smooth minimum
```
janosg, line 284: This should have a scaling factor we can set and test in benchmarks.

src/estimagic/optimization/tranquilo/sample_points.py
```diff
@@ -110,47 +117,207 @@ def _naive_sampler(
...
+            x vector at which the criterion function has already been evaluated, that
+            satisfies lower_bounds <= existing_xs <= upper_bounds.
+        bounds (Bounds or None): NamedTuple.
 
+    """
+    n_points = _get_effective_n_points(target_size, existing_xs=existing_xs)
+    n_params = len(trustregion.center)
+    effective_bounds = _get_effective_bounds(trustregion, bounds=bounds)
+
+    points = rng.normal(size=(n_points, n_params))
```
janosg, line 167: I think a uniform might be better for box sampling. Would make it flexible for now so we can try it out in a benchmark.

src/estimagic/optimization/tranquilo/sample_points.py
```diff
@@ -110,47 +117,207 @@ def _naive_sampler(
...
+    Args:
+        x (np.ndarray): 2d array of points to be projects. Each value is in [-1, 1].
+        ord (int): Type of norm to use when scaling the sampled points. For 2 we project
+            onto the hull of a sphere, for np.inf onto the hull of a cube.
+
+    Returns:
+        np.ndarray: The projected points.
+
+    """
+    norm = np.linalg.norm(x, axis=1, ord=ord).reshape(-1, 1)
```
janosg, line 316: Is the non-differentiability of the infinity norm a problem for the optimizer? Have you tried a high finite order, e.g. 100?

src/estimagic/optimization/tranquilo/sample_points.py
timmens: Very good point. For now, I'm sticking with estimagic but will use 1d-arrays in the optimization to use the fast path. From here on it will be easy to switch to `scipy.minimize` if we decide to do that at some point.

src/estimagic/optimization/tranquilo/sample_points.py
timmens: I chose the name `lse_scale` for logsumexp scale. Any better ideas for the argument name?

src/estimagic/optimization/tranquilo/sample_points.py
timmens: It did not seem that non-differentiability was a great problem here I tried a high finite order and it worked as well as the infinity norm. Should we replace the infinity-norm with some large number for ord? Or should we implement a general sampler for an arbitrary order?

src/estimagic/optimization/tranquilo/sample_points.py
janosg: The name should indicate the direction. It is the opposite of a smoothing factor. I would probably call it `hardness` and describe in the docstring that this refers to how closely the soft minimum approximates a hard minimum. Nobody who uses this even knows about what logsumexp is and why we use it here. 

And as always when we use that trick, link to this blogpost for further explanation: https://www.johndcook.com/blog/2010/01/20/how-to-compute-the-soft-maximum/

src/estimagic/optimization/tranquilo/sample_points.py
janosg: I would choose an arbitrary order. Only question is then when we switch the default random sampling from normal to uniform. Probably normal for order <= 3 and uniform above that.

janosg (approved):
Very nice!

PR #389, opened 2022-10-12
Add pounders filtering to tranquilo

segsell:
- [x] Fix bug in Pounders fitting
- [x] Make Tranquilo more flexible
- [x] Add pounders filtering
- [x] Refactor
- [x] Write more tests


janosg (approved):
Thanks! I ran some benchmarks on the different fitting methods and settled on `fit_pounders_experimental`. The differences are small but it is better in some cases.

PR #391, opened 2022-10-19
Speeding up subsolvers

mpetrosian:
- [x] render  bntr subsolver numba-compatible. This includes jitting the following functions:
    - [x] `take_preliminary_gradient_descent_step_and_check_for_solution`
    - [x] `compute_conjugate_gradient_step`
    - [x] `minimize_trust_trsbox`
    - [x] `minimize_trust_cg`
    - [x] `minimize_trust_stcg`


src/estimagic/optimization/subsolvers/_trsbox_quadratic_fast.py
```diff
@@ -0,0 +1,655 @@
...
+                step_len_constr = (
+                    upper_bounds[i] - x_candidate[i]
+                ) / gradient_projected[i]
+            else:
+                step_len_constr = (
+                    lower_bounds[i] - x_candidate[i]
+                ) / gradient_projected[i]
+            if step_len_constr < step_len:
+                step_len = step_len_constr
+                index_bound_active = np.array([i])
```
janosg, line 466: Two questions:
1. is `index_bound_active` always defined? As far as I see it is only defined inside the if condition but who guarantees that this if condition is always true at least once?
2. Why is this an array of fixed length 1 instead of just an int?

src/estimagic/optimization/subsolvers/_trsbox_quadratic_fast.py
```diff
@@ -0,0 +1,655 @@
...
+    of ``tangent`` in the interval [0, ``bound_on_tangent``], where ``tangent`` is
+    the tangent of half the angle to the trust-region boundary.
+    """
+    previous_reduction = None
+    next_reduction = None
+
+    max_reduction = 0
+    index_angle_greatest_reduction = -1
+    old_reduction = 0
+    n_angles = int(17 * bound_on_tangent + 3.1)
```
janosg, line 545: Don't we need `np.int64` instead of just int? @timmens is this similar to the case you had?

timmens: I think there is a difference between the `int()` method and passing an unsupported dtype to, for example, a `numpy.ndarray`. But maybe someone should take a look at the detailed log.

Otherwise, this could be done with `np.floor(...).astype(np.int64)`.

janosg:
Ran another set of benchmarks. If anything, the faster subsolvers also perform better.

janosg (approved):

PR #395, opened 2022-11-01
Refactor tranquilo code

janosg:
- [x] Improve construction of test cases in `test_tranquilo.py`.
- [x] Make sure we have one function to calculate the necessary number of parameters/degrees of freedom for linear and quadratic models. Should be defined in `models.py`.
- [x] Remove unused arguments in the aggregator interface. 
- [x] Tests for input processing functions in tranquilo.py.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -6,7 +6,7 @@ repos:
     -   id: debug-statements
     -   id: end-of-file-fixer
 -   repo: https://github.com/asottile/reorder_python_imports
-    rev: v3.8.5
+    rev: v3.9.0
     hooks:
     -   id: reorder-python-imports
         types: [python]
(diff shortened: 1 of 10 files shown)


timmens:
1. `subsolvers/quadratic_subsolvers.py`, `subsolvers/bounded_newton_quadratic.py` and `models.py` contain functions to evaluate models. However, they use different arguments. In `subsolvers/linear_subsolvers.py` a linear model is evaluated, but we cannot use the other functions straight aways since they assume that the input has at least the attribute `square_terms`.
2.  Function `_polynomial_features` in `fit_models.py` is very similar to `_square_features` in `filter_points.py`. We could think about if we can combine these two or call one from the other.

timmens:
We will tackle problems 1. and 2. in separate PRs.

segsell (approved):
Very nice refactoring PR! The test modules look way cleaner now. And great job finding out that ```np.linalg.lstsq``` is more robust than ```np.linalg.sovle``` in ```powell_fitting```. Thank you!!

segsell:
The powell fitting benchmarks now look slightly worse switch from ```np.linalg.solve``` to ```np.linalg.lstsq```. But this is hardly noticeable in therelevant case of ```2 * n + 1``` points.

PR #399, opened 2022-11-10
Add poisedness check for sample quality

segsell:
Add function to calculate the lambda poisedness of a sample.

Note that we move to ```black``` python version ```3.10``` in ```.pre-commit-config.yaml```.


janosg (approved):
Very nice! Thanks!

PR #404, opened 2022-11-13
Speedup gqtpar

mpetrosian:
Add a new subsolver, gqtpar_fast, where we apply numba-jitting to most functions.


janosg (approved):
Thanks! Great speedup! I ran benchmarks and the fast and slow gqtpar are essentially identical.

PR #411, opened 2022-11-19
Fix bug in multistart with least-squares optimizers

### LINKED ISSUE #410: Error in multistart with `ls` optimizers
Hi! this is Mateo from last year's DSE Summer School. I'm starting to use estimagic and it is just, well, magical! Thank you!

I've run into this small bug.

### Bug description

Least squares algorithms appear not to work with multistarts. I think the issue is that some updating step of multistarts is looking for the best criterion *value* in a list where the least-squares algo has stored the vector of residuals instead.

### To Reproduce

```
import estimagic as em
import numpy as np

def loss(params):

    resid = np.array([params['x'], params['y']])
    out = {
        "root_contributions": resid,
        "contributions": np.square(resid),
        "value": np.sum(np.square(resid)),
    }
    return out

est = em.minimize(
    criterion=loss,
    params={'x': -1.0, 'y': 1.0},
    lower_bounds={'x': -10, 'y': -10},
    upper_bounds={'x': 10, 'y': 10},
    algorithm='scipy_ls_trf',
    multistart=True,
    multistart_options={
        "n_samples": 3,
        "share_optimizations": 1.0
    }
)

print(est)
```

### Expected behavior

I'd expect the above code to perform three starts.

### Screenshots/Error messages

Instead, it raises the following error after the first start

```
Traceback (most recent call last):
  File "/home/mvg/em_fail.py", line 16, in <module>
    est = em.minimize(
  File "/usr/local/Anaconda3/lib/python3.9/site-packages/estimagic/optimization/optimize.py", line 403, in minimize
    return _optimize(
  File "/usr/local/Anaconda3/lib/python3.9/site-packages/estimagic/optimization/optimize.py", line 766, in _optimize
    raw_res = run_multistart_optimization(
  File "/usr/local/Anaconda3/lib/python3.9/site-packages/estimagic/optimization/tiktak.py", line 168, in run_multistart_optimization
    state, is_converged = update_convergence_state(
  File "/usr/local/Anaconda3/lib/python3.9/site-packages/estimagic/optimization/tiktak.py", line 471, in update_convergence_state
    if valid_new_y[best_index] <= best_y:
ValueError: The truth value of an array with more than one element is ambiguous. Use a.any() or a.all()
```

### System

 - OS: Ubuntu 18.04

janosg:
resolves #410

### DIFF
--- a/src/estimagic/optimization/tiktak.py
+++ b/src/estimagic/optimization/tiktak.py
@@ -461,14 +461,19 @@ def update_convergence_state(
     best_y = current_state["best_y"]
     best_res = current_state["best_res"]
 
+    # get indices of local optimizations that did not fail
     valid_indices = [i for i, res in enumerate(results) if not isinstance(res, str)]
 
+    # If all local optimizations failed, return early so we don't have to worry about
+    # index errors later.
     if not valid_indices:
         return current_state, False
 
+    # ==================================================================================
+    # reduce eveything to valid optimizations
+    # ==================================================================================
     valid_results = [results[i] for i in valid_indices]
     valid_starts = [starts[i] for i in valid_indices]
-
     valid_new_x = [res["solution_x"] for res in valid_results]
     valid_new_y = []
 
@@ -485,6 +490,9 @@ def update_convergence_state(
                 )
             )
 
+    # ==================================================================================
+    # accept new best point if we find a new lowest function value
+    # ==================================================================================
     best_index = np.argmin(valid_new_y)
     if valid_new_y[best_index] <= best_y:
         best_x = valid_new_x[best_index]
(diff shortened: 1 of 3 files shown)


src/estimagic/optimization/tiktak.py
```diff
@@ -456,17 +461,38 @@ def update_convergence_state(current_state, starts, results, convergence_criteri
...
+    for res in valid_results:
+        if np.isscalar(res["solution_criterion"]):
+            valid_new_y.append(res["solution_criterion"])
+        else:
+            valid_new_y.append(
+                aggregate_func_output_to_value(
+                    f_eval=res["solution_criterion"],
+                    primary_key=primary_key,
+                )
+            )
```
timmens, line 491: Why do we need to if-clause? Doesn't `aggregate_func_output_to_value` handle the scalar case as well?

janosg: No, it does not. Usually this function is called, when we exactly know what we will get (e.g. on the function value that gets passed to a least squares optimizer). Here we are not sure how much post-processing has already been done in by the internal least squares optimizer.

timmens (approved):
Great that we've found and solved the bug. Approved.

PR #415, opened 2022-11-30
Update README and pre-commit config

### LINKED ISSUE #413: Use modern pre-commit hooks and implement own
There are some interesting new pre-commit hooks that should be considered for the estimagic development, e.g. [refurbed](https://github.com/dosisod/refurb). Further, we should consider writing our own hook that checks that functions are called with keyword arguments and refactors them automatically. 

## Tasks

- [ ] Write a custom hook that checks the usage of args and kwargs in functions
- [ ] Delete hooks that we don't use
- [ ] Check if there are new hooks

timmens:
In this PR I added estimagics monthly download numbers as a badge to our README. I further went through our pre-commit hooks and looked for new ones. One new hook that stands out is [refurbed](https://github.com/dosisod/refurb). It proposes ways to replace code sections with more pythonic versions; however, it is not ready yet. It is too slow, sometimes proposes wrong things and ignoring arguments does not work well with pre-commit. My recommendation is to use the tool via the console once in a while and adopt some of the proposed changes but not integrate it into our pre-commit hook suite.

#### Changes
I recommend looking at the `.pre-commit-config.yaml` changes in split view. I deleted no hooks but added `{pyupgrade, name-tests-test, check-toml, forbid-submodules}`.

#### Tasks
- [x] Add estimagic download numbers to README
- [x] Close #413


timmens:
@janosg 

The new hook `name-tests-test` fails because
```console
tests/visualization/helpers_test_estimation_table.py does not match pattern "test_.*\.py"
```
We can either rewrite (this would mean adding the three functions defined in `helpers_test_estimation_table.py` to the beginning of `test_estimation_table.py`) or ignore this case. I think I'd prefer a rewrite. What do you think?

janosg:
Yes, I prefer the rewrite as well. Can you do it?

janosg (approved):
Thanks a lot!

PR #425, opened 2023-01-17
Improve handling of notebooks in documentation

### LINKED ISSUE #422: Improve handling of notebooks in documentation
### What would you like to enhance and why? Is it related to an issue/problem?

We would like to execute most notebooks during each sphinx build, such that it becomes impossible to get outdated notebooks. Notebooks with very long runtime (e.g. [this](https://github.com/OpenSourceEconomics/estimagic/blob/main/docs/source/explanations/inference/bootstrap_montecarlo_comparison.ipynb)) or notebooks with optional dependencies need to be excluded.

See the discussion #420 for background.

### Tasks

- [ ] Extend the `rtd_environment.yml` file such that estimagic can actually run on readthedocs. Currently it does for example not contain numpy or scipy. 
- [ ] Add a pre-commit hook that deletes the output of all notebooks we want to execute during the sphinx build
- [ ] Set `nbsphinx_execute = "never"` to `nbsphinx_execute = "auto"` in `conf.py

### Potential alternatives

It is possible that this is still too demanding for readthedocs hardware. Last time we tried, we failed in the environment creation. In that case we could use a local pre-commit hook to run all notebooks that have changed.

baharcos:
This PR changes the handling of notebooks in documentation from nbsphinx to [myst-nb](https://myst-nb.readthedocs.io/en/latest/).

+ up-to-date/maintained better
+ option to force the execution of notebooks without needing to delete the outputs
+ option to exclude some notebooks from the forced execution
+ adjustable maximum cell execution time

Current caveats:
- the download feature of the notebooks is lost but potentially can be reintroduced
- it has to be manually checked whether the notebook cells run without errors

Jax is used as the example case: 
[developer notes](https://jax.readthedocs.io/en/latest/developer.html#)
[readthedocs.yml](https://github.com/google/jax/blob/main/.readthedocs.yml)
[configuration](https://github.com/google/jax/blob/main/docs/conf.py)
[requirements](https://github.com/google/jax/blob/main/docs/requirements.txt)


baharcos:
The latex rendering is not working properly in this notebook: https://estimagic--425.org.readthedocs.build/en/425/how_to_guides/miscellaneous/how_to_generate_publication_quality_tables.html and I couldnt fix. @janosg Do you maybe know how to have 
from IPython.core.display import HTML, Latex 
and  
from IPython.display import IFrame in the rtd_environment? (ipython is already in the rtd_environment) If not I will just exclude this notebook from forced execution while building.

janosg (approved):
Thanks!

PR #427, opened 2023-01-21
Format yaml files

### LINKED ISSUE #426: Styling of yaml files
### Current situation

Consistent styling is helpful for many reasons. Also applies to yaml files, particularly when comparing settings across projects.

Styling of yaml files is ad-hoc.
 
### Desired Situation

Style yaml files automatically.

### Proposed implementation

Add yamlfix and yamllint hooks as described [here](https://github.com/OpenSourceEconomics/estimagic/pull/420#issuecomment-1382756450) , fix settings so that it they work in estimagic.

Based on [this](https://github.com/OpenSourceEconomics/estimagic/pull/420#issuecomment-1382756450) conversation, I had been under the impression that these were added in #420, but they did not appear. 

Adding them is not fully trivial. Reasons AFAICT:

- [ ] Fix settings of yamlfix so that yamllint does not complain when test files are changed.
- [ ] Align the output of `.envs/update_envs.py` with what yamlfix will do (e.g., six spaces for the pip-installed packages, align settings yamlfix/yamllint for inline comments)

timmens:
Closes #426.


timmens:
In order for `yamllint` to work I had to edit the configuration from
```yaml
hyphens: enable
```
to
```yaml
hypens:
   max-spaces-after: 4
```
which allows yaml files with nested lists like:
```yaml
history_x:
  -   - 0.15
       - 0.008
       - 0.01
  -   - 0.25
       - 0.008
       - 0.01
```
In principle, it might be possible to align the settings of yamlfix and yamllint here, but I could not find the yamlfix option which governs this behaviour.

.yamllint.yml
```diff
@@ -0,0 +1,38 @@
...
+  comments-indentation:
+    level: warning
+  document-end: disable
+  document-start:
+    level: warning
+  empty-lines: enable
+  empty-values: disable
+  float-values: disable
+  hyphens:
+    max-spaces-after: 4
```
hmgaudecker, line 23:
```suggestion
    max-spaces-after: 3
```
I think this should be enough. My guess of what the parser is doing in these cases is that it will take 1 space after the hyphen of the outer list and two spaces for the inner list.

.pre-commit-config.yaml
```diff
@@ -1,10 +1,14 @@
+---
```
hmgaudecker, line 1: @janosg @tobiasraabe -- can you please double-check that the order of the hooks makes sense for you here and if possible also for other projects? If that is consistent, diffing will be much easier now that the formatting is consistent, too.

environment.yml
```diff
@@ -1,3 +1,4 @@
+---
```
hmgaudecker, line 1: I would be in favour of just sorting alphabetically, again, for easily seeing differences across projects. Thoughts?

hmgaudecker (commented):
Looks great to me, thanks a lot!!! 

I would love to see common settings for the hooks across projects so one can easily spot differences. In particular, what would be good IMHO:
- Same order of hooks (if possible)
- Same settings for (basic) py / yaml / md linting and formatting
- Same sorting of environment files (alphabetically?)

Please append to the list if I am missing anything. I think estimagic is a good place to settle that because it's environment is close to being a superset of the others' environments.

@janosg @tobiasraabe I explicitly tagged you re the order, but it would be great if you could have a look at the other settings, too.


Partly related: yamlfix is based on [ruyaml](https://ruyaml.readthedocs.io/en/latest/) -- loading/dumping based on that would add another dependency but probably prevent some pre-commit surprises (and all of them if it could be configured according to the section in pyproject.toml). No strong opinion here, just wanted to toss it into the round.

janosg:
In principle I like the standardization across projects, but we have tried that multiple times in the past and it always failed. Back then the idea was that there is one tiny project (I think it was called the OSE packaging guide) that always uses the latest version of pre-commits, CI and other infrastructure and each project copies from there. Typically nobody copied and thus at some point we stopped updating. 

I think the only way we can implement something like that would be to automate it. I think Tobi does that to some extent for the many pytask repos. But probably after automatic updates we would still have to fix problems manually and I am not sure if people are willing to do that. There are also some things we probably cannot standardize. For example I would love to enable some more ruff rules in estimagic but fixing all issues in such a large codebase takes time. In small projects we can be stricter.

RE sorting: I think I prefer it if environment files are sorted by category (e.g. runtime dependencies, test dependencies, ...) but I don't have a strong opinion. I do like estimagic's new way of auto-generating variations of the development environment for testing, but probably most other packages don't need that. 

If we start another real effort to standardize the whole infrastructure across OSE packages, we should probably discuss that in a larger group. Definitely out of scope for this PR.

janosg:
BTW: The failing link will be fixed when you merge the main in this branch.

timmens:
My only addition to Janos' comment is that we already sort alphabetically in each category, except for the first and last line.

hmgaudecker:
> In principle I like the standardization across projects, but we have tried that multiple times in the past and it always failed. Back then the idea was that there is one tiny project (I think it was called the OSE packaging guide) that always uses the latest version of pre-commits, CI and other infrastructure and each project copies from there. Typically nobody copied and thus at some point we stopped updating.

Yeah, fair enough, I remember that and I don't think we will be able to go fully automatic.

> RE sorting: I think I prefer it if environment files are sorted by category (e.g. runtime dependencies, test dependencies, ...) but I don't have a strong opinion. I do like estimagic's new way of auto-generating variations of the development environment for testing, but probably most other packages don't need that.

Exactly. Different projects just have different needs. What I like to do to compare is to diff files from different projects. Three things were annoying in the past on that one:

1. Whitespace/formatting *(solved by having the same formatters/linters and settings for them)*
2. Different versions of hooks *(solved by pre-commit CI)*
3. Different order of the same things where order does not matter / is not crucial *(could be solved by agreeing on a rule and starting with it in a project that has the largest numbers of hooks / settings)*

Because of the last point, I think this is the right place to start.

hmgaudecker (approved):
Looks good to me! 

*Disclaimer: Did not check all 281 files.*

janosg (approved):

PR #437, opened 2023-02-14
Replace fuzzywuzzy

aidatak97:
Rewrote the propose_alternatives function using difflib
Added some simple unit tests for it
Removed fuzzywuzzy from environment.yml and setup.cfg


janosg (approved):
Thanks!

PR #443, opened 2023-03-07
Tranquilo cleanup

timmens:
In this PR, we clean up certain components of tranquilo:

1. We create a new module `bounds.py` for the `Bounds` dataclass.
1. We create a new module `region.py` for the `Region` dataclass
    1. The region class is now aware of its shape, which is a sphere if bounds aren't binding and a cube otherwise
    1. The region class now provides methods `map_to_unit` and `map_from_unit` for transitioning between the unit space used in points sampling, fitting, and solving the subproblem, and the regular trustregion space.
1. The sampling code now adaptively samples from the correct shape using the new shape information of the trustregion.
1. We replace the class `ModelInfo` with the string `model_type` which can only be either `linear` or `quadratic`.

### DIFF
--- a/src/estimagic/optimization/tranquilo/models.py
+++ b/src/estimagic/optimization/tranquilo/models.py
@@ -1,5 +1,5 @@
 from dataclasses import dataclass, replace
-from typing import NamedTuple, Union
+from typing import Union
 
 import numpy as np
 from numba import njit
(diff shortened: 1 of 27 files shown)


src/estimagic/optimization/tranquilo/models.py
```diff
@@ -308,25 +303,12 @@ def _predict_scalar(model: ScalarModel, centered_x: np.ndarray) -> np.ndarray:
...
-        if info.has_interactions:
-            out += n_interactions(dim)
-    elif isinstance(info_or_name, str) and info_or_name in (
-        "linear",
-        "quadratic",
-        "diagonal",
-    ):
-        name = info_or_name
-        if name == "quadratic":
+    if model_type in ("linear", "quadratic"):
```
timmens, line 309: Unnecessary check

janosg (approved):
A few small comments but I already approve.

PR #444, opened 2023-03-10
Add functionality to Region class

timmens:
In this PR, I extend the functionality of the Region class. I create modules region.py and bounds.py and add tests for their private functions and classes.

The biggest change happened to the `Region` class. Which is now a `dataclass` object with **attributes** (center, radius, bounds); **properties** (shape, cube_bounds, cube_center) and **methods** (map_to_unit, map_from_unit). The properties are dynamically updated based on the center, radius and the user-provided bounds. If any of the bounds is binding we switch from a spherical trustregion shape to a cube one. The methods `map_to_unit` and `map_from_unit` map points from the trustregion to the unit space and back. They are attached to this class because they are used in multiple modules and only depend on the information stored in the trustregion.


src/estimagic/optimization/tranquilo/region.py
```diff
@@ -0,0 +1,96 @@
...
+    def shape(self) -> str:
+        any_bounds_binding = _any_bounds_binding(
+            bounds=self.bounds, center=self.center, radius=self.radius
+        )
+        return "cube" if any_bounds_binding else "sphere"
+
+    @property
+    def cube_bounds(self) -> Bounds:
+        if self.shape == "sphere":
+            radius = self.radius
```
janosg, line 29: I think we should do one of the two if `shape == "sphere"`
1. raise an Attribute error for cube_bounds
2. return cube_bounds but also use a cube_radius

Simply returning bounds constructed from a sphere radius is not a good option in my opinion. I am tending to option 2.

src/estimagic/optimization/tranquilo/region.py
```diff
@@ -0,0 +1,96 @@
...
+            radius = get_radius_of_cube_with_volume_of_sphere(
+                self.radius, len(self.center), scaling_factor=1.0
+            )
+        cube_bounds = _get_cube_bounds(
+            center=self.center, radius=radius, bounds=self.bounds
+        )
+        return cube_bounds
+
+    @property
+    def cube_center(self) -> np.ndarray:
```
janosg, line 40: Same thing here. Either raise an error if `shape == "sphere"` or return whatever center is implied by the cube bounds.

src/estimagic/optimization/tranquilo/region.py
```diff
@@ -0,0 +1,96 @@
...
+        if self.shape == "sphere":
+            _center = self.center
+        else:
+            cube_bounds = self.cube_bounds
+            _center = (cube_bounds.lower + cube_bounds.upper) / 2
+        return _center
+
+    def map_to_unit(self, x: np.ndarray) -> np.ndarray:
+        """Map points inside the trustregion to the unit sphere or cube."""
+        return _map_to_unit(self.cube_bounds, x=x)
```
janosg, line 50: If we define cube bounds as suggested above, we will need the if condition here. If `shape == "sphere"` -> scale by radius, if `shape == "cube"` -> scale by cube bounds.

src/estimagic/optimization/tranquilo/sample_points.py
```diff
@@ -100,28 +95,29 @@ def _box_sampler(
...
-        bounds (Bounds or None): NamedTuple.
 
     """
     n_params = len(trustregion.center)
-    effective_bounds = _get_effective_bounds(trustregion, bounds=bounds)
 
     points = rng.uniform(
-        low=effective_bounds.lower,
-        high=effective_bounds.upper,
+        low=trustregion.cube_bounds.lower,
```
janosg, line 109: should probably evaluate `cube_bounds` once and then use its attributes to save repeated calculation.

src/estimagic/optimization/tranquilo/tranquilo.py
```diff
@@ -158,14 +157,14 @@ def _tranquilo(
 
     bounds = Bounds(lower=lower_bounds, upper=upper_bounds)
 
-    if trustregion_shape is None:
-        trustregion_shape = "sphere" if not bounds.has_any else "cube"
-
     if sampler is None:
-        sampler = f"optimal_{trustregion_shape}"
+        sampler = "optimal_sphere"
```
janosg, line 161: In the long run this should just be `"optimal"`. Will you do that in the next PR?

src/estimagic/optimization/tranquilo/tranquilo.py
```diff
@@ -158,14 +157,14 @@ def _tranquilo(
...
 
     if subsolver is None:
-        subsolver = "bntr_fast" if trustregion_shape == "cube" else "gqtpar_fast"
+        if _any_finite(bounds.lower, bounds.upper):
```
janosg, line 164: Can you use `bounds.has_any`?

tests/optimization/tranquilo/test_region.py
```diff
@@ -0,0 +1,75 @@
...
+
+def test_region_binding_bounds():
+    bounds = Bounds(lower=-np.ones(2), upper=0.5 * np.ones(2))
+    region = Region(center=np.zeros(2), radius=1, bounds=bounds)
+    assert region.shape == "cube"
+    assert region.radius == 1
+    assert region.bounds is bounds
+    # shrinkage because cube radius is smaller than (spherical) radius
+    assert np.all(bounds.lower - region.cube_bounds.lower < 0)
+    assert_array_equal(region.cube_bounds.upper, bounds.upper)
```
janosg, line 75: Nice simple tests!

src/estimagic/visualization/deviation_plot.py
```diff
@@ -70,7 +70,9 @@ def deviation_plot(
         .reset_index()
     )
     average_deviations = (
-        deviations.groupby(["algorithm", "n_evaluations"]).mean()[outcome].reset_index()
+        deviations.groupby(["algorithm", "n_evaluations"])
+        .mean(numeric_only=True)[outcome]
+        .reset_index()
```
timmens, line 75: Response to deprecation warning.

PR #445, opened 2023-03-15
Tranquilo Refactoring

timmens:
- [x] Replace np.squeeze with operation that is more precise, unless an axis argument is given
- [x] Clip sampled points at bounds
- [x] Adaptively sample points given trustregion shape


janosg (approved):
We can merge after:
- Test coverage is higher
- We confirmed that the benchmark is unchanged

PR #446, opened 2023-03-17
Improve argument processing of tranquilo

janosg:
## Problems

1. The default configuration of tranquilo is quite complex. This is easiest shown with an example: The default `target_sample_size` depends on the `model_type`; The default `model_type` depends on the `functype`. But the `model_type` can also be overwritten by the user. If so, we need to check that the user specified a valid model type before we can create the default `target_sample_size`. I.e. everything is entangled and the order of processing stuff has to be chosen very carefully. 
2. The default values are not collected in one place. 
3. The namespace of the main tranquilo function is cluttered with stuff that was only needed to initialize the main components; It is hard to tell which variables will get used later on and which can be easily changed.

## Solution 

- All default values should be defined in `options.py`; If default values depend on other arguments, we implement `getter_functions`, e.g. `get_default_radius_options(x)`
- Nothing else is in `options.py`; In particular, all the logic related to combining default options and user options is in `process_arguments.py`. 
- The outer `tranquilo` function only calls `process_arguments` and then calls `_internal_tranquilo` with the processed arguments. This removes all namespace cluttering. A small trick to avoid long lists of arguments, the outer `tranquilo` functions only uses `*args, **kwargs` as arguments and `functools.wraps(process_arguments`) to get the right signature and docstring.

## Alternatives

We could use dags to figure out the order of everything, but we currently do not have dags as a dependency and we don't need it anywhere else. In any case, the current PR gets us closer to a situation where we could use dags. 

## To-Do

- [x] Move the actual logic from tranquilo into `_internal_tranquilo`
- [x] Rename `_new_tranquilo` to `_tranquilo`
- [ ] Run all tests and a benchmark
- [x] Make `RadiusOptions.initial_radius` a mandatory argument without default value
- [x] Make all ConvOptions mandatory 
- [x] Make all StopOptions mandatory
- [x] Write some tests for the functions in process_arguments and options.py
- [x] Put all component default options into `options.py`. Example:
  - [x] Define a class `SubsolverOptions` that contains everything the `default_options` dict in `solve_subproblem` contains 
  - [x] Adjust `get_component` sucht that the user_options can be NamedTuples on which we call `_asdict`
  - [x] Instantiate the new class in `solve_subproblem` instead of defining a dict there
- [x] Improve error message in default aggregator and move the entire compatibility check from `aggregate_models` into `get_default_aggregator` if possible. 
- [x] make default model fitter dependent on model type and sample size instead of functype
- [ ] Improve error handling and type checking inside process_arguments

## Discuss
- Check up on differences in benchmarks introduced by commit  [9d68fab](https://github.com/OpenSourceEconomics/estimagic/pull/446/commits/9d68fabed31e3a9450d5d492083578e733945f96)
- Double-check if we want to divide `AcceptanceOptions` into individual options in `acceptance_decision.py`.

### DIFF
--- a/src/estimagic/optimization/tranquilo/get_component.py
+++ b/src/estimagic/optimization/tranquilo/get_component.py
@@ -4,6 +4,7 @@ import warnings
 from functools import partial
 
 from estimagic.utilities import propose_alternatives
+from estimagic.optimization.tranquilo.options import update_option_bundle
 
 
 def get_component(
@@ -27,8 +28,8 @@ def get_component(
         component_name (str): Name of the component. Used in error messages. Examples
             would be "subsolver" or "model".
         func_dict (dict): Dict with function names as keys and functions as values.
-        default_options (NamedTuple, Dict or None): Default options as a dict or
-            NamedTuple. The default options will be updated by the user options.
+        default_options (NamedTuple): Default options as a dict or NamedTuple. The
+            default options will be updated by the user options.
         user_options (NamedTuple, Dict or None): User options as a dict or NamedTuple.
             The default options will be updated by the user options.
         redundant_option_handling (str): How to handle redundant options. Can be
@@ -128,8 +129,8 @@ def _get_valid_options(
     """Get the options that are valid for the function.
 
     Args:
-        default_options (NamedTuple, Dict or None): Default options as a dict or
-            NamedTuple. The default options will be updated by the user options.
+        default_options (NamedTuple): Default options as a dict or NamedTuple. The
+            default options will be updated by the user options.
         user_options (NamedTuple, Dict or None): User options as a dict or NamedTuple.
             The default options will be updated by the user options.
         signature (list): List of arguments that are present in the signature.
@@ -142,16 +143,8 @@ def _get_valid_options(
         dict: Valid options.
 
     """
-    _default_options = {} if default_options is None else default_options
-    _user_options = {} if user_options is None else user_options
-
-    if not isinstance(_user_options, dict):
-        _user_options = user_options._asdict()
-
-    if not isinstance(_default_options, dict):
-        _default_options = default_options._asdict()
-
-    _options = {**_default_options, **_user_options}
+    _options = update_option_bundle(default_options, user_options=user_options)
+    _options = _options._asdict()
 
     _valid_options = {k: v for k, v in _options.items() if k in signature}
     _redundant_options = {k: v for k, v in _options.items() if k not in signature}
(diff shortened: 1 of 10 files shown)


src/estimagic/optimization/tranquilo/get_component.py
```diff
@@ -143,6 +145,12 @@ def _get_valid_options(
     _default_options = {} if default_options is None else default_options
     _user_options = {} if user_options is None else user_options
 
+    if not isinstance(_user_options, dict):
+        _user_options = user_options._asdict()
+
+    if not isinstance(_default_options, dict):
+        _default_options = default_options._asdict()
+
```
janosg, line 153: I think we should use `update_option_bundle` function here to get the automatic type checks and conversions for free. This might require moving it to `options.py` to avoid circular imports.

src/estimagic/optimization/tranquilo/options.py
```diff
@@ -1,11 +1,74 @@
...
+        aggregator = "least_squares_linear"
+    elif functype == "likelihood" and model_type == "linear":
+        aggregator = "information_equality_linear"
+    else:
+        allowed_combinations = {
+            "scalar": "quadratic",
+            "least_squares": "linear",
+            "likelihood": "linear",
+        }
+        raise ValueError(
```
janosg, line 62: `NotImplementedError` would probably be better.

src/estimagic/optimization/tranquilo/solve_subproblem.py
```diff
@@ -85,20 +86,7 @@ def get_subsolver(solver, user_options=None, bounds=None):
...
-        "gtol_rel": 1e-8,
-        "gtol_scaled": 0,
-        "gtol_abs_conjugate_gradient": 1e-8,
-        "gtol_rel_conjugate_gradient": 1e-6,
-        "k_easy": 0.1,
-        "k_hard": 0.2,
-    }
-
-    all_options = {**default_options, **user_options}
+    all_options = SubsolverOptions()._replace(**user_options)._asdict()
```
janosg, line 89: Also use `update_option_bundle` to get type conversion

janosg (commented):
Thanks. Three minor points. After they are fixed I'll run another benchmark and then we can merge.

src/estimagic/optimization/tranquilo/solve_subproblem.py
timmens: This will yield a circular import. This would, however, be fixed in the upcoming subsolver PR, when we adapt the subsolver code to the `get_component` framework.

timmens (approved):
Great PR. This is an important step forward to disentangle the actual algorithm development of tranquilo and its options management.

PR #449, opened 2023-03-22
Fix bugs in tranquilo

timmens:
- [x] Make sure that centering and scaling is done via `map_to_unit` and `map_from_unit`
- [x] Adjust `model.py` to the case of sphere and cube trustregions; focus on `move_model`.
- [x] Add `effective_center` and `effective_radius` to `Region`
- [x] Add `residualize` option
- [x] Fix `get_default_model_fitter`
- [x] Use `state.x` instead of `state.trustregion.center` when appropriate

### Notes for reviewer:

- [x] Check `filter_points.py` line 96-97:
  ```python
  order = 2 if state.trustregion.shape == "sphere" else np.inf
  dists = np.linalg.norm(xs - state.x, axis=1, ord=order)
  ```
- [x] Check `new_history.py` line 207-213:
  ```python
  out = _find_indices_in_region(xs, center=region.center, radius=region.radius)

  if region.shape == "sphere":
      mask = np.linalg.norm(xs[out] - region.center, axis=1) <= region.radius
      out = out[mask]

  return out
  ```


src/estimagic/optimization/tranquilo/options.py
```diff
@@ -27,13 +27,18 @@ def get_default_sample_size(model_type, x):
...
+    if model_type == "linear" or sample_size >= n_params:
+        fitter = "ols"
     else:
         fitter = "tranquilo"
     return fitter
 
 
+def get_default_residualize(model_type, sample_size, x):
+    n_params = n_free_params(dim=len(x), model_type=model_type)
+    return not (model_type == "linear" or sample_size >= n_params)
```
janosg, line 39: Should we merge these two functions to express that `residualize` is by default used iff the tranquilo fitter is used?

src/estimagic/optimization/tranquilo/models.py
janosg, line 1: - `VectorModel` and `ScalarModel` should be frozen
- `subtract_model` can be deleted

src/estimagic/optimization/tranquilo/region.py
janosg, line 1: We should probably do all the calculations in `__post_init__` and store the results in private attributes. The properties would then  only do error handling and look up the private attributes.

src/estimagic/optimization/tranquilo/models.py
janosg: Maybe `move_model` and the lower level functions `_scale_model` and `_shift_model` should take and return coefficient arrays instead of models, so we don't need to make sure that the `region`/mapping information of the intermediate models is set correctly. Should also be faster and more readable and names can be aligned with the math notation in our docstrings.

janosg (commented):
Thanks a lot! Very nice. 

A last general comment: We should probably replace the suffix `_centered` (e.g. for x, samples, ...) by `_unit`.

src/estimagic/optimization/tranquilo/options.py
janosg: Or make `default_residualize` dependent on fitter

src/estimagic/optimization/tranquilo/new_history.py
```diff
@@ -202,20 +201,10 @@ def get_x_indices_in_region(self, region):
...
-            center = self.get_xs(x_indices=center)
-
         xs = self.get_xs()
 
-        out = _find_indices_in_trust_region(xs, center=center, radius=radius)
-
-        if shape == "sphere":
-            mask = np.linalg.norm(xs[out] - region.center, axis=1) <= radius
-            out = out[mask]
+        mask = np.linalg.norm(xs - region.center, axis=1) <= region.radius
```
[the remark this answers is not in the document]
timmens: Under what conditions did the state occur where we used the early return before?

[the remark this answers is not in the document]
janosg: Ah, sorry, I might have been confused: The early returned handled an empty history, not an empty mask.

src/estimagic/optimization/tranquilo/region.py
```diff
@@ -16,31 +16,19 @@ class Region:
...
+        cube_center = _get_cube_center(cube_bounds)
+        effective_center = _get_effective_center(shape, self.center, cube_center)
+        effective_radius = _get_effective_radius(shape, self.radius, cube_bounds)
+
+        # cannot use standard __setattr__ because it is frozen
+        super().__setattr__("shape", shape)
+        super().__setattr__("cube_bounds", cube_bounds)
+        super().__setattr__("cube_center", cube_center)
+        super().__setattr__("effective_center", effective_center)
+        super().__setattr__("effective_radius", effective_radius)
```
janosg, line 31: I think `cube_bounds` and `cube_center` need to be private attributes that are accessed via corresponding properties. Currently we lost the error handling (i.e. `AttributeErrors if shape == "sphere"`)

janosg (approved):
Very nice!

src/estimagic/optimization/tranquilo/models.py
```diff
@@ -166,86 +134,79 @@ def move_model(model, new_region):
...
+        out = _shift_vector_model(out, shift=shift)
+
+    # scale region to scaling of new_region
+    scale_from_unit = new_region.effective_radius
+    out = _scale_model(out, scale=scale_from_unit)
     return out
 
 
-def _scale_model(model, old_radius, new_radius):
+def _scale_model(model, scale):
```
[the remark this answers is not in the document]
janosg: Done.

src/estimagic/optimization/tranquilo/models.py
```diff
@@ -166,86 +134,79 @@ def move_model(model, new_region):
         Union[ScalarModel, VectorModel]: The moved model.
 
     """
-    old_region = model.region
-    out = _scale_model(model, old_radius=old_region.radius, new_radius=1.0)
+
+    # scale region to unit space
```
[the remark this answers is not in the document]
janosg: Done.

PR #450, opened 2023-03-29
Report all issues rather than stop at first one.

Kaniyaki:
I created an empty list called "report". If constraints are satisfied, an error message is appended to the "report". 
I updated all the conditions, and in the end, a dictionary is transformed into a string. 
If that string is not empty, the error message is raised. 
I used the "print" statement and "assert False" to check if errors are raised.

@janosg should I keep the "Invalid model specification:" message when the error is raised? All unit tests pass, so this should be good.


janosg:
Hi @Kaniyaki, looks great. 

I would replace "Invalid model specification" by "Violated constraint at start params". Estimagic is a very general library and not everything you optimize could be called a model. 

Currently github action fails because test coverage over your changes is low. This is nothing you introduced, but could you fix it? I.e. call the function with invalid inputs and use `with pytest.raises` to make sure the correct error is raised?

janosg (approved):
Sorry that this got lost for so long. Great work!

PR #452, opened 2023-03-31
Optimize code to replace the df

Kaniyaki:


Kaniyaki:
Optimized the "check_fixes_and_bounds" function, which originally had a runtime of 5.61 ms. 
The main problem was that it used DataFrame. 
Instead of it, I used a dictionary and added the "_iloc"  function as a substitute for DataFrame.iloc. Current runtime is 55 µs, which is a significant improvement. 
In the end, I made sure that all tests passed, and they did.
Below, you can see the comparison for the runtime using %timeit and %lprun. 
timeit_comparisson
[lprun_comparisson.pdf](https://github.com/OpenSourceEconomics/estimagic/files/11129878/lprun_comparisson.pdf)

janosg (approved):
Again, sorry for the delay. Great work, just polished it a bit.

PR #455, opened 2023-04-06
adjust `update_envs` to make sure that created files have LF line endings

ChristianZimpelmann:
- On Windows, the environment files created by `update_envs` have CRLF line endings.
- In particular, this leads to problems with the `mixed-line-ending` hook.

- This PR makes sure that LF line endings are used by opening the file in binary mode. 
- I am of course open for other solutions.

### DIFF
--- a/.envs/update_envs.py
+++ b/.envs/update_envs.py
@@ -42,9 +42,10 @@ def main():
     # write environments
     for name, env in zip(["linux", "others"], [test_env_linux, test_env_others]):
         # Specify newline to avoid wrong line endings on Windows.
-        # See: https://stackoverflow.com/a/23434608
-        with Path(f".envs/testenv-{name}.yml").open("w", newline="\n") as file:
-            file.write("\n".join(env) + "\n")
+        # See: https://stackoverflow.com/a/69869641
+        Path(f".envs/testenv-{name}.yml").write_text(
+            "\n".join(env) + "\n", newline="\n"
+        )
 
 
 if __name__ == "__main__":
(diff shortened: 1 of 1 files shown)


janosg:
Hi @ChristianZimpelmann, can you describe the problem you get with the mixed-line-ending hook? I thought the whole point of that hook was to fix such problems.

ChristianZimpelmann:
> Hi @ChristianZimpelmann, can you describe the problem you get with the mixed-line-ending hook? I thought the whole point of that hook was to fix such problems.

The problem is that `mixed-line-ending` and `update-environment-files` fail alternatingly and alternatingly change the line endings from LF to CRLF and viceversa. Somehow `mixed-line-ending` does not check the files created by `update-environment-files` during the same pre-commit run, but only after I try to commit again.

janosg:
Ah, probably because they are not staged. 

I was just worried how robust this approach is. Do you have a source for this trick? Then we could just add a comment and the source to explain why we open a yaml file in binary mode.

ChristianZimpelmann:
Good point!

I found a much better solution now, but nevertheless added a comment including [a source for the solution](https://stackoverflow.com/a/23434608).

timmens:
Our development environment is already using Python 3.10, so we could also use this:

```python
Path('my.txt').write_text('hello\nworld', newline='\n')
```

Source: https://stackoverflow.com/a/69869641

janosg (approved):
We can merge this from my side (as is or with Tim's suggestion). Thanks a lot!

ChristianZimpelmann:
Thanks, @timmens for the reference! That seems like the best solution.

PR #460, opened 2023-04-23
Improve processing of noisy benchmark results

janosg:
## Problem

Currently, the benchmark plots are based on the actual history of criterion values the optimizer generated. In the noisy case, this means that optimizers who get lucky draws are rewarded. 

## Solution

We re-evaluate a noise free criterion on the history of parameters the optimizer produced. The re-calculation is done in parallel to save time.


src/estimagic/benchmarking/run_benchmark.py
```diff
@@ -154,40 +168,52 @@ def _get_kwargs_list_and_names(problems, opt_options):
...
+    Returns:
+        dict: Processed result.
+
+    """
+    _registry = get_registry(extended=True)
+    _criterion = problem["noise_free_criterion"]
+    _start_x = problem["inputs"]["params"]
+    _start_crit_value = _criterion(_start_x)
+    if isinstance(_start_crit_value, np.ndarray):
+        _start_crit_value = (_start_crit_value**2).sum()
```
timmens, line 187: If `_start_crit_value` is a `np.ndarray` we are always in the least-squares case?

src/estimagic/benchmarking/process_benchmark_results.py
```diff
@@ -45,267 +42,146 @@ def create_convergence_histories(
...
-        converged = y_converged
-    elif stopping_criterion == "x":
-        converged = x_converged
-    elif stopping_criterion == "x_and_y":
-        converged = y_converged & x_converged
-    elif stopping_criterion == "x_or_y":
-        converged = y_converged | x_converged
+def _aggregate_idxs_with_or(x, y):
+    if x is None or y is None:
+        out = None
```
timmens, line 184: This should be `and` and then the else clause needs to be two elif clauses.

timmens (approved):
Conditional approve.

src/estimagic/benchmarking/process_benchmark_results.py
janosg: Done.

PR #462, opened 2023-04-28
Benchmark reports

segsell:
Add benchmark reports:

- [x] convergence report
- [x] rank report
- [x] traceback report


segsell:
Two questions:

1) Do we want the traceback report to be part of the ```benchmark_reports``` module in estimagic?
2) The convergence and rank reports contain similar information. Both have as columns the optimizers we compare against each other and as indexes the name of the benchmark problem. Moreover, both have as values the strings "failed" and "error". In case of success, the convergence report prints "success", whereas the rank report shows the rank of the optimizer (where 0 means the optimizer was the fastest in the competition). A second difference is that the convergence report additionally contains the column ```dimensionality```, which denotes the dimension of the respective benchmark problems.
So, since both reports are similar, do we want to include the ```dimensionality``` information also in the rank report? So that the user, if she/he wants to access all the information in one report, only needs to call the rank report.

See these two screenshots below. The first one refers to the convergence report, the second to the rank report.

Screenshot from 2023-04-28 09-47-46

Screenshot from 2023-04-28 09-48-13

src/estimagic/benchmarking/benchmark_reports.py
```diff
@@ -0,0 +1,171 @@
...
+        y_precision=y_precision,
+    )
+    scenarios = list({algo[1] for algo in results.keys()})
+
+    success_info = converged_info.replace({True: "success", False: "failed"})
+    for key, value in results.items():
+        if isinstance(value["solution"], str):
+            success_info.at[key] = "error"
+
+    solution_times = histories.groupby(["problem", "algorithm"])[runtime_measure].max()
```
janosg, line 121: This is duplicated with code in `profile_plot._create_solution_times`, but I think it forgets to set the solution time to infinity for unsolved problems which would potentially lead to very wrong results. We should probably make `_create_solution_times` public and use it here.

src/estimagic/benchmarking/benchmark_reports.py
```diff
@@ -0,0 +1,171 @@
...
+
+    if normalize_runtime:
+        solution_times = solution_times.unstack()
+        solution_times = solution_times.divide(solution_times.min(axis=1), axis=0)
+        solution_times = solution_times.stack(dropna=False)
+        solution_times.name = runtime_measure
```
janosg, line 127: Since the normalization should not matter for the rank, I think this can go.

tests/benchmarking/test_benchmark_reports.py
janosg, line 1: These tests are high cost (we actually run benchmarks to get inputs) and low reward (we only do asserts on shapes and column names. I think it would be better to use a few very simple hand-crafted benchmark results where we actually know the exact result. Two problems and two optimizers would be enough. Then we could have one failure, one unsolved but not failed and and two successful cases. The histories can also all be unrealistically short.

janosg (commented):
Thanks for the PR. I made a few comments. In addition to those, we also need the new functionality to be documented with examples [here](https://github.com/OpenSourceEconomics/estimagic/blob/main/docs/source/how_to_guides/optimization/how_to_benchmark_optimization_algorithms.ipynb)

src/estimagic/benchmarking/benchmark_reports.py
```diff
@@ -0,0 +1,171 @@
...
+        stopping_criterion=stopping_criterion,
+        x_precision=x_precision,
+        y_precision=y_precision,
+    )
+    scenarios = list({algo[1] for algo in results.keys()})
+
+    success_info = converged_info.replace({True: "success", False: "failed"})
+    for key, value in results.items():
+        if isinstance(value["solution"], str):
+            success_info.at[key] = "error"
```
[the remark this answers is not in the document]
segsell: Thank you! This clearly has to go in a private function.

src/estimagic/benchmarking/benchmark_reports.py
segsell: I can do that. From a DRY perspective, I understand the virtue of using ```profile_plot._create_solution_times``` here. Two notes, however, why I decided not to call it directly:

1) We need the ```solution_times``` DataFrame to be in a different format. Calling the ```profile_plot._create_solution_times``` function would require to apply ```solution_times.stack().reset_index()``` 
to the resulting DataFrame.

2) What ```profile_plot._create_solution_times``` does with ```solution_times[~converged_info] = np.inf```, is done [here](https://github.com/OpenSourceEconomics/estimagic/blob/benchmark_reports/src/estimagic/benchmarking/benchmark_reports.py#L139). ```success_info``` contains one of the strings "success", "failed", "error" as values for each benchmark problem. Hence, problems that did not converge are not assigned an integer rank, but the string "failed" or "error". For comparison, in  ```profile_plot._create_solution_times``` ```np.inf``` refers to either "failed" or "error".

src/estimagic/benchmarking/benchmark_reports.py
segsell: Thanks! Yes, we don't actually need it here.

src/estimagic/benchmarking/benchmark_reports.py
```diff
@@ -0,0 +1,171 @@
...
+
+    return convergence_report
+
+
+def rank_report(
+    problems,
+    results,
+    *,
+    runtime_measure="n_evaluations",
+    normalize_runtime=False,
```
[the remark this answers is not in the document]
segsell: Yes, absolutely!

tests/benchmarking/test_benchmark_reports.py
segsell: You're right, the tests are too expensive. Will do as you suggested!

janosg (approved):

PR #464, opened 2023-05-20
Make tranquilo an optional dependency

janosg:
- [ ] Remove all files related to tranquilo
- [ ] Add importing from tranquilo package
- [ ] Remove numba as estimagic depedency
    - [ ] Make sure it is not used anywhere else in estimagic 
- [ ] Add Python 3.11 to CI 
- [ ] Make a release


src/estimagic/algorithms.py
```diff
@@ -0,0 +1,46 @@
...
+    nag_optimizers,
+    nlopt_optimizers,
+    pygmo_optimizers,
+    scipy_optimizers,
+    simopt_optimizers,
+    tao_optimizers,
+    bhhh,
+    neldermead,
+    pounders,
+    tranquilo,
```
timmens, line 30: Why not stick to the import order? It seems less error-prone when we update this.

```suggestion
    bhhh,
    cyipopt_optimizers,
    fides_optimizers,
    nag_optimizers,
    neldermead,
    nlopt_optimizers,
    pounders,
    pygmo_optimizers,
    scipy_optimizers,
    simopt_optimizers,
    tao_optimizers,
    tranquilo,
```

timmens (approved):
Great PR, approved! I only have two minor remarks.

pyproject.toml
```diff
@@ -91,6 +91,7 @@ filterwarnings = [
     "ignore:Widget._widget_types is deprecated",
     "ignore:Widget.widget_types is deprecated",
     "ignore:Widget.widgets is deprecated",
+    "ignore:Parallelization together with",
```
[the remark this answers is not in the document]
janosg: From tranquilo.

src/estimagic/algorithms.py
janosg: Will do that next time.

PR #467, opened 2023-08-09
Add estimagic logo to documentation page

mpetrosian:


docs/source/index.md
timmens, line 1: Two things:

1. Is there a reason why you don't close the `img` tag? Like ``. It renders nonetheless, but it seems slightly weird to me.
2. Have you tried centering the image, and what do you think (example image attached)? I think a simple way would be to write
    ```html
    <div id="logo" style="text-align: center;">
    ...
    </div>
  
<img width="1259" alt="Screenshot 2023-08-10 at 10 44 08" src="https://github.com/OpenSourceEconomics/estimagic/assets/39708667/2b042e6a-e350-401e-aa26-40ebb2ec0c82">

timmens:
Okay, I'm just realizing that the `img` needs to be closed, since, otherwise the logo div spans much further than it should. On the attached image you see that the text below is also centered, this is because the centering in the logo div also applies to that section because the img tag is not closed.

docs/source/index.md
mpetrosian: 1. Thanks for the catch! Fixed!
2.  I think, you're right; the center-aligned logo looks better!

timmens (approved):
Looks great, thank you!

timmens (approved):
I still don't understand why the mock import does not work for tranquilo, but at least the autodoc works now. I've also left the `tranquilo_scalar` change as it is, though I think it is not necessary to fix the bug.

janosg:
I changed the name inside `mark_minimizer` back so we don't get a breaking change.

janosg (approved):

PR #477, opened 2024-02-07
Identity Weighting Matrix

sidd3888:
Added the identity weighting matrix to the MSM estimation:
- First change made is to the `get_weighting_matrix()` function to allow for the identity matrix
- Second change permits `"identity"` as a valid option for weights in `estimate_msm()`

### DIFF
--- a/src/estimagic/estimation/estimate_msm.py
+++ b/src/estimagic/estimation/estimate_msm.py
@@ -149,7 +149,7 @@ def estimate_msm(
     # Check and process inputs
     # ==================================================================================
 
-    if weights not in ["diagonal", "optimal"]:
+    if weights not in ["diagonal", "optimal", "identity"]:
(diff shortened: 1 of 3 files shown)


janosg:
Hi @sidd3888, thank you very much for your Pull Request. The changes are great. There is just one more thing: It would be great to also have a small test case for the `get_weighting_matrix` with the "identity" option. 

Let me know whether you want to add this test or I should do it. If you want to add the test, it should be in [this file](https://github.com/OpenSourceEconomics/estimagic/blob/main/tests/estimation/test_msm_weighting.py)

sidd3888:
If I understand the code there correctly, this would require nothing more than just adding "identity" as a possible option for the `get_weighting_matrix` test right? I can add that to this PR right away.

janosg:
Yes. Thank you so much!

janosg (approved):

PR #482, opened 2024-03-03
Take care of deprecation warnings from dependencies

### LINKED ISSUE #470: Replace `DateFrame.applymap` with `DataFrame.map` (in estimation_table)
pandas 2.1. deprecated DataFrame.applymap
> New in version 2.1.0: DataFrame.applymap was deprecated and renamed to DataFrame.map. 

The current `estimation_table` uses `applymap` in several occasions and produces many DeprecationWarnings.

- replace with `.map`
- as long as we still support pandas 2.0, we should probably ignore the warning

timmens:
In this PR, we take care of deprecation warnings from estimagic's dependencies.

In particular, we

- [x] Call `df.map()` instead of `df.applymap()` if pandas version 2.1.0 or greater is installed
- [x] Add new module `compat.py`, in which we implement functions with compatibility issues
- [x] Ignore deprecation warning about MemoizeJac, which is raised in cyipopt. It seems that there are working on this as well (see [here](https://github.com/scipy/scipy/issues/17572))
- [x] Use preferred `np.prod` instead of  `np.product`, which is deprecated with numpy version 1.25.0 (see release notes [here](https://numpy.org/devdocs/release/1.25.0-notes.html#deprecations). In earlier versions of numpy `np.prod` was already implemented and `np.product` would simply call `np.prod`; therefore we can just swap `np.product` for `np.prod`. (There were some subtle differences, but they do not apply to our use cases; see [here](https://stackoverflow.com/questions/49863633/numpy-product-vs-numpy-prod-vs-ndarray-prod))
- [x] Use `pd.Series(list).unique()` instead of `pd.unique(list)`, as the latter ist deprecated. Also use `pd.Series(list).value_counts()` instead of `pd.value_counts(list)`.
- [x] Remove `maxiter` option from `scipy_truncated_newton` algorithm, since it is not a valid option.
- [x] Use `DataFrame.replace()` properly for categorical columns. The old implementation raised the following errors:
   ```console
   estimagic/tests/inference/test_bootstrap.py:70: FutureWarning: Downcasting behavior in `replace` is deprecated and 
   will be removed in a future version. To retain the old behavior, explicitly call `result.infer_objects(copy=False)`. To opt-in 
   to the future behavior, set `pd.set_option('future.no_silent_downcasting', True)`
   df = raw.replace({"time": replacements})

   estimagic/tests/inference/test_bootstrap.py:70: FutureWarning: The behavior of Series.replace (and 
   DataFrame.replace) with CategoricalDtype is deprecated. In a future version, replace will only be used for cases that 
   preserve the categories. To change the categories, use ser.cat.rename_categories instead.
   df = raw.replace({"time": replacements})
   ```
- [x]  Remove usage of DataFrameGroupBy.grouper, due to following warning:
  ```console
   estimagic/src/estimagic/parameters/consolidate_constraints.py:595: FutureWarning: DataFrameGroupBy.grouper is deprecated and will be removed in a future version of pandas.
    weights["dupl_group"] = weights.groupby(list(weights.columns)).grouper.group_info[0]
  ```
  **Current solution:**
  ```python
  weights["dupl_group"] = weights.groupby(list(weights.columns)).ngroup()
  ```
- [x] Set dtype to float before overwriting certain values with np.inf, due to the following warning:
   ```console
   estimagic/src/estimagic/visualization/profile_plot.py:169: Setting an item of incompatible dtype is deprecated and will raise an error in a future version of pandas. Value 'inf' has dtype incompatible with int64,  please explicitly cast to a compatible dtype first.
   solution_times[~converged_info] = np.inf
   ```
  **Current solution:**
  - Cast data frame to float for all columns, then assign np.inf.
  - Previously all columns where we assigned np.inf where casted to float anyways, resulting in a data frame with some float and some int columns. Now they are all float.
- [x] Conversion of an array with ndim > 0 to a scalar is deprecated. Specifically, this occurs in
  - tranquilo (see issue [#15](https://github.com/OpenSourceEconomics/tranquilo/issues/15) in tranquilo)
  - simopt strong optimize (**todo**: what to do about simopt)
  **Current solution:** Ignore the warning.
- [x] For CI run certain tests with pandas version < 2.0.0
  - [x] Select tests we want to run
  - [x] Create new CI task
  - [x] Add new environment to .envs
- [x] Add Python 3.12 to linux test matrix
--- 

Closes #470.


src/estimagic/parameters/parameter_groups.py
```diff
@@ -35,10 +35,10 @@ def get_params_groups_and_short_names(params, free_mask, max_group_size=8):
         names.append(name)
 
     # if every parameter has its own group, they should all actually be in one group
-    if len(pd.unique(groups)) == len(groups):
+    if len(pd.Series(groups).unique()) == len(groups):
```
hmgaudecker, line 38:
```suggestion
    if len(set(groups)) == len(groups):
```

src/estimagic/compat.py
```diff
@@ -0,0 +1,33 @@
...
+
+Contains wrapper functions to handle compatibility issues between different versions of
+external libraries.
+
+"""
+
+from estimagic.config import IS_PANDAS_VERSION_NEWER_OR_EQUAL_TO_2_1_0
+
+
+def pandas_df_map(df, func, na_action=None, **kwargs):
```
hmgaudecker, line 11:
```suggestion
def pd_df_map(df, func, na_action=None, **kwargs):
```
Since `import pandas as pd` is the convention used throughout, that seems easier to parse.

src/estimagic/visualization/profile_plot.py
```diff
@@ -160,13 +160,12 @@ def create_solution_times(df, runtime_measure, converged_info, return_tidy=True)
...
             desired precision. If the desired precision was not achieved the value is
-            set to np.inf (for n_evaluations) or 7000 days (for walltime since there
-            no infinite value is allowed).
+            set to np.inf.
 
     """
     solution_times = df.groupby(["problem", "algorithm"])[runtime_measure].max()
-    solution_times = solution_times.unstack()
-    solution_times[~converged_info] = np.inf
+    solution_times = solution_times.unstack().astype(float)
```
hmgaudecker, line 167: So what is it if not a float already? And does the unit matter or do we just not care here?

hmgaudecker (approved):
Looks great to me -- just a couple of small remarks!

src/estimagic/visualization/profile_plot.py
timmens: It is an int. For the runtime measures `n_batches` and `n_evaluations` an integer makes more sense. However, I do not think that converting to a float will change any subsequent calculations. But I haven't written this code, so I am not entirely sure.

src/estimagic/parameters/parameter_groups.py
timmens: Nice one!

src/estimagic/visualization/profile_plot.py
hmgaudecker: Ah, then I would just put it in the line below, makes it clearer that it is done so that inf is supported.

hmgaudecker:
While you are at it, I have a tiny feature request: Please add a newline at the end of `estimation_table.render_latex()` (and possibly other formats). This seems to be the standard convention throughout. It also seems such  a small change to not merit its own set of issue & PR.

Core reason: In one project, the output ends up under VC and there are tons of fake changes until the pre-commit hooks run, which fix this issue.

.github/workflows/main.yml
```diff
@@ -72,6 +73,33 @@ jobs:
         run: |
           micromamba activate estimagic
           pytest -m "not slow and not jax"
+  run-tests-with-old-pandas:
```
timmens, line 76: Add a comment on why we do this.

PR #483, opened 2024-03-04
Small bug in bootstrap keywords

alanlujan91:
"cluster" is not a bootstrap option

### DIFF
--- a/tests/estimation/test_msm_weighting.py
+++ b/tests/estimation/test_msm_weighting.py
@@ -95,18 +95,19 @@ def test_get_moments_cov_passes_bootstrap_kwargs_to_bootstrap():
     rng = get_rng(1234)
     data = rng.normal(scale=[10, 5, 1], size=(100, 3))
     data = pd.DataFrame(data=data)
+    data["cluster"] = np.random.choice([1, 2, 3], size=100)
 
     def calc_moments(data, keys):
         means = data.mean()
         means.index = keys
         return means.to_dict()
 
-    moment_kwargs = {"keys": ["a", "b", "c"]}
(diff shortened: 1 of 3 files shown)


tests/estimation/test_msm_weighting.py
```diff
@@ -95,18 +95,19 @@ def test_get_moments_cov_passes_bootstrap_kwargs_to_bootstrap():
...
-    moment_kwargs = {"keys": ["a", "b", "c"]}
+    moment_kwargs = {"keys": ["a", "b", "c", "cluster"]}
 
     with pytest.raises(ValueError, match="a must be a positive integer unless no"):
         get_moments_cov(
             data=data,
             calculate_moments=calc_moments,
             moment_kwargs=moment_kwargs,
-            bootstrap_kwargs={"n_draws": -1},
+            bootstrap_kwargs={"n_draws": -1, "cluster_by": "cluster"},
```
timmens, line 112: I would prefer not to add the cluster_by option test to this block. Here we want to check that the bootstrap kwargs are actually passed to the bootstrap function. I think for the cluster_by argument we would rather want to check that an error is thrown if an invalid option is passed. Maybe you could add something like this below:

```python
with pytest.raises(ValueError, match="Invalid bootstrap_kwargs: {'cluster'}"):
    get_moments_cov(
        data=data,
        calculate_moments=calc_moments,
        moment_kwargs=moment_kwargs,
        bootstrap_kwargs={"cluster": "cluster"},
    )
```

timmens (changes requested):
Hey Alan! Thanks for finding this bug and directly providing a fix.

I have only a minor remark about the test, but other than that, it looks great :tada:

timmens (approved):
Thanks a lot!

PR #480, opened 2024-02-28
Fix API section in docs

timmens:
In this PR, I fix the API section in the documentation.

In particular, I restrict plotly to version 5.18 or lower.

When using 5.19, autodoc returned the following error, which is raised in plotly-code that was newly introduced in version 5.19:

```console
WARNING: autodoc: failed to import function 'convergence_plot' from module 'estimagic'; the following exception was raised:
Traceback (most recent call last):
  File "/home/tim/miniforge3/envs/estimagic-docs/lib/python3.9/site-packages/sphinx/ext/autodoc/importer.py", line 62, in import_module
    return importlib.import_module(modname)
  File "/home/tim/miniforge3/envs/estimagic-docs/lib/python3.9/importlib/__init__.py", line 127, in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
  File "<frozen importlib._bootstrap>", line 1030, in _gcd_import
  File "<frozen importlib._bootstrap>", line 1007, in _find_and_load
  File "<frozen importlib._bootstrap>", line 986, in _find_and_load_unlocked
  File "<frozen importlib._bootstrap>", line 680, in _load_unlocked
  File "<frozen importlib._bootstrap_external>", line 850, in exec_module
  File "<frozen importlib._bootstrap>", line 228, in _call_with_frames_removed
  File "/home/tim/miniforge3/envs/estimagic-docs/lib/python3.9/site-packages/estimagic/__init__.py", line 2, in <module>
    from estimagic.benchmarking.get_benchmark_problems import get_benchmark_problems
  File "/home/tim/miniforge3/envs/estimagic-docs/lib/python3.9/site-packages/estimagic/benchmarking/get_benchmark_problems.py", line 5, in <module>
    from estimagic.benchmarking.cartis_roberts import CARTIS_ROBERTS_PROBLEMS
  File "/home/tim/miniforge3/envs/estimagic-docs/lib/python3.9/site-packages/estimagic/benchmarking/cartis_roberts.py", line 21, in <module>
    from estimagic.config import IS_NUMBA_INSTALLED
  File "/home/tim/miniforge3/envs/estimagic-docs/lib/python3.9/site-packages/estimagic/config.py", line 3, in <module>
    import plotly.express as px
  File "/home/tim/miniforge3/envs/estimagic-docs/lib/python3.9/site-packages/plotly/express/__init__.py", line 14, in <module>
    from ._imshow import imshow
[...]

PR #487, opened 2024-03-08
Fix documentation

timmens:
In this PR, we fix the API section of the documentation.

### What happened?

In our documentation, we use [mock imports](https://www.sphinx-doc.org/en/master/usage/extensions/autodoc.html#confval-autodoc_mock_imports). From the sphinx docs:
> This is useful when some external dependencies are not met at build time and break the building process. You may only specify the root package of the dependencies themselves and omit the sub-modules:

However,  _mocked_ modules cannot be used during build time. This was not a problem until a few weeks ago when `plotly` started checking `pandas` version numbers (mocked for our documentation), which requires the execution of `pandas` code --- namely `pandas.__version__` --- during build time. My unawareness of what was actually happening led me to restrict the `plotly` version (see #480).

In #482, I added code to estimagic that also checked the `pandas` version numbers, leading to the same problem as before.

### Fix

Remove `pandas` from the list of mock imports.

### DIFF
--- a/docs/rtd_environment.yml
+++ b/docs/rtd_environment.yml
@@ -4,7 +4,7 @@ channels:
   - conda-forge
   - nodefaults
 dependencies:
-  - python=3.9
+  - python=3.10
   - pip
   - setuptools_scm
   - toml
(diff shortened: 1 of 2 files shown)


hmgaudecker (approved):
Helpful write-up of what happened!

PR #495, opened 2024-05-02
Enhancement Proposals for Typing and Alignment

janosg:
# Two new enhancement proposals

This PR adds two new enhancement proposals. The two originated as one proposal and were only split after the joint proposal was already accepted. 

## EEP-02 Static typing

This enhancement proposal explains the adoption of static typing in estimagic. The goal
is to reap a number of benefits:

- Users will benefit from IDE tools such as easier discoverability of options and
  autocompletion.
- Developers and users will find code easier to read due to type hints.
- The codebase will become more robust due to static type checking and use of stricter
  types in internal functions.

Achieving these goals requires more than adding type hints. estimagic is currently
mostly [stringly typed](https://wiki.c2.com/?StringlyTyped). For example, optimization
algorithms are selected via strings. Another example are
[constraints](https://estimagic.readthedocs.io/en/latest/how_to_guides/optimization/how_to_specify_constraints.html),
which are dictionaries with a fixed set of required keys.

This enhancement proposal outlines how we can accommodate the changes needed to reap the
benefits of static typing without breaking users' code in too many places.

## EEP-03 Alignment with scipy

Since the typing proposal will already introduce some breaking changes and deprecations, we can use the same deprecation cycle to  better align some names in estimagic with SciPy and other optimization libraries. This will
make it even easier for scipy users to switch to estimagic. The goals are:

- If we can make code written for SciPy run with estimagic, we should do so
- If we cannot make it run, the user should get a helpful error message that explains
  how the code needs to be adjusted.

We will achieve this by renaming arguments that are compatible with what SciPy' minimize expects to the SciPy names and adding aliases for all other arguments of `scipy.optimize.minimize`. Some of the aliases are purely meant to give better error messages for users coming from scipy, others add actual functionality.


docs/source/development/eep-02-typing.md
```diff
@@ -0,0 +1,445 @@
...
+[option dictionaries](https://estimagic.readthedocs.io/en/latest/how_to_guides/optimization/how_to_specify_algorithm_and_algo_options.html),
+etc.).
+
+This enhancement proposal outlines how we can accomodate the changes needed to reap the
+benefits of static typing without breaking users' code in too many places.
+
+## Motivation and ressources
+
+- [Writing Python like it's Rust](https://kobzol.github.io/rust/python/2023/05/20/writing-python-like-its-rust.html).
+  A very good blogpost that the drawbacks of "stringly-typed" Python code and shows how
```
timmens, line 42:
```suggestion
  A very good blogpost that summarizes the drawbacks of "stringly-typed" Python code and shows how
```

docs/source/development/eep-02-typing.md
```diff
@@ -0,0 +1,445 @@
...
+```python
+def sphere(params: np.ndarray) -> float:
+    return params @ params
+
+
+def dict_sphere(params: dict) -> float:
+    return params["a"] ** 2 + params["b"] ** 2
+
+
+def least_squares_sphere(params: np.ndarray) -> dict[str:Any]:
```
timmens, line 85:
```suggestion
def least_squares_sphere(params: np.ndarray) -> dict[str, Any]:
```

docs/source/development/eep-02-typing.md
```diff
@@ -0,0 +1,445 @@
...
+The following changes apply to all functions that are directly related to optimization,
+i.e. `maximize`, `minimize`, `slice_plot`, `criterion_plot`, `params_plot`,
+`count_free_params`, `check_constraints` and `OptimizeResult`.
+
+### The `criterion` function
+
+#### Current situation
+
+A function that takes params (a pytree) as first argument and returns a scalar (if only
+scalar algorithms will be used) or a dictionary that contains at the entries "value" (a
```
timmens, line 67:
```suggestion
scalar algorithms will be used) or a dictionary that contains the entries "value" (a
```

docs/source/development/eep-02-typing.md
```diff
@@ -0,0 +1,445 @@
...
+- We do not need to restrict the type of additional arguments of the criterion function.
+- The simplest form of our criterion functions is also compatible with scipy.optimize
+
+**Problems**
+
+- Newcomers find it hard to specify least-squares problems
+- Internally we can make almost no assumptions about the output of a criterion function,
+  making the code very complex and brittle
+- The best typing information we could get for the output of the criterion function is
+  `float | dict[str: Any]` which is not very useful.
```
timmens, line 104:
```suggestion
  `float | dict[str, Any]` which is not very useful.
```

docs/source/development/eep-02-typing.md
```diff
@@ -0,0 +1,445 @@
...
+
+#### Proposal
+
+`params` and additional keyword arguments of the criterion function stay unchanged. The
+output of the criterion function becomes `float | CriterionValue`. There are decorators
```
timmens, line 109:
```suggestion
output of the criterion function becomes `float | CriterionValue`. They are decorators
```

docs/source/development/eep-02-typing.md
```diff
@@ -0,0 +1,445 @@
...
+```python
+def least_squares_sphere(params: np.ndarray) -> em.CriterionValue:
+    out = CriterionValue(
+        residuals=params, info={"p_mean": params.mean, "p_std": params.std()}
+    )
+    return out
+```
+
+We can exploit this deprecation cycle to rename `root_contributions` to `residuals`
+which is more in line with the literater.
```
timmens, line 125:
```suggestion
which is more in line with the literature.
```

docs/source/development/eep-02-typing.md
janosg: I think that one was correct.

docs/source/development/eep-02-typing.md
```diff
@@ -0,0 +1,1474 @@
...
+- For likelihood functions, `criterion` returns a dictionary that contains at least the
+  entry `"contributions"` which can be any pytree that can be flattened into a numpy
+  array of floats.
+- For least-squares problems, `criterion` returns a dictionary that contains at least
+  the entry `"root_contributions"` which can be any pytree that can be flattened into a
+  numpy array of floats.
+- In any case the returned dictionary can have an arbitrary number of additional keys.
+  The corresponding information will be stored in the log database (if logging is used).
+
+A few simple examples of valid criterion functions are:
```
hmgaudecker, line 96: Can you add a link to the docs / tutorials? IIRC, this is being used there, else may be explain a little? I found it especially confusing since you open above with *the same criterion function can work for scalar, least-squares and likelihood optimizers*, so I was trying to map into that until realizing it has nothing to do with each other.

docs/source/development/eep-02-typing.md
```diff
@@ -0,0 +1,1474 @@
...
+def dict_sphere(params: dict) -> float:
+    return params["a"] ** 2 + params["b"] ** 2
+
+
+def least_squares_sphere(params: np.ndarray) -> dict[str, Any]:
+    out = {"root_contributions": params, "p_mean": params.mean(), "p_std": params.std()}
+    return out
+```
+
+**Things we want to keep**
```
hmgaudecker, line 112: Always start with the thing to keep, then add the reason.

docs/source/development/eep-02-typing.md
```diff
@@ -0,0 +1,1474 @@
...
+### The objective function
+
+#### Current situation
+
+The objective or criterion function is the function being optimized. Currently, it is
+called `criterion` in estimagic.
+
+The `criterion` function maps a pytree of parameters into a criterion value.
+
+An important feature of estimagic is that the same criterion function can work for
```
hmgaudecker, line 78:
```suggestion
The same criterion function can work for
```
Leave benefits for the part below.

docs/source/development/eep-02-typing.md
```diff
@@ -0,0 +1,1474 @@
...
+
+
+def least_squares_sphere(params: np.ndarray) -> dict[str, Any]:
+    out = {"root_contributions": params, "p_mean": params.mean(), "p_std": params.std()}
+    return out
+```
+
+**Things we want to keep**
+
+- The fact that `params` can be arbitrary pytrees makes estimagic flexible and popular.
```
hmgaudecker, line 114:
```suggestion
- Allow arbitrary pytrees for the `params` argument. This makes estimagic flexible and popular.
```

docs/source/development/eep-02-typing.md
```diff
@@ -0,0 +1,1474 @@
...
+  We do not need to restrict this type in any way because flattening the pytree gives us
```
hmgaudecker, line 115: I would not explain this here, this is the goals section?

docs/source/development/eep-02-typing.md
```diff
@@ -0,0 +1,1474 @@
...
+  a very precise type no matter how complicated the tree was.
+- Using the same criterion function for scalar, likelihood and least-squares optimizers
```
hmgaudecker, line 117:
```suggestion
- Allow using the same criterion function for scalar, likelihood and least-squares optimizers. This feature
```

docs/source/development/eep-02-typing.md
```diff
@@ -0,0 +1,1474 @@
...
+  makes it easy to try out and compare very different algorithms with minimal code
+  changes.
+- We do not need to restrict the type of additional arguments of the criterion function.
```
hmgaudecker, line 120:
```suggestion
- No restrictions on the type of additional arguments of the criterion function.
```
Reason?

docs/source/development/eep-02-typing.md
```diff
@@ -0,0 +1,1474 @@
...
+- The simplest form of our criterion functions is also compatible with scipy.optimize
```
hmgaudecker, line 121:
```suggestion
- Maintain compatibility with scipy.optimize when the criterion function returns a scalar.
```
Not sure whether this is correct, but in any case be explicit.

docs/source/development/eep-02-typing.md
```diff
@@ -0,0 +1,1474 @@
...
+
+**Problems**
+
+- Most users of estimagic find it hard to write criterion functions that return the
+  correct dictionary. Therefore, they don't use the logging feature and we often get
+  questions about specifying least-squares problems correctly.
+- Internally we can make almost no assumptions about the output of a criterion function,
+  making the code that processes the criterion output very complex and full of if
+  conditions.
+- The best typing information we could get for the output of the criterion function is
```
hmgaudecker, line 131: Not sure this qualifies as a problem under the status quo (no static typing). Seems more about process / implementation, i.e., a problem with the code base conditioning on wanting to make the move.

docs/source/development/eep-02-typing.md
```diff
@@ -0,0 +1,1474 @@
...
+  `float | dict[str, Any]`, which is not very useful.
+- We only know if the specified criterion function is compatible with the selected
```
hmgaudecker, line 133:
```suggestion
- We only know whether the specified criterion function is compatible with the selected
```

docs/source/development/eep-02-typing.md
```diff
@@ -0,0 +1,1474 @@
...
+  optimizer after we evaluate it once. This means that errors for users are raised very
```
hmgaudecker, line 134:
```suggestion
  optimizer after we evaluate it once. This means that users see errors only very
```

docs/source/development/eep-02-typing.md
```diff
@@ -0,0 +1,1474 @@
...
+  late.
+- While optional, in least-squares problems it is possible that a user specifies
```
hmgaudecker, line 136: I don't get this point. Seems like a consequence of wanting to use the same criterion in multiple optimizers, which you cite as a strength above? So what precisely is the problem / what do you mean with "compatible"?

docs/source/development/eep-02-typing.md
```diff
@@ -0,0 +1,1474 @@
...
+  `root_contributions`, `contributions` and `value` even though any of them could be
+  constructed out of the `root_contributions`. This means we need to check that all of
+  them are compatible.
+- The name `criterion` is not used very widely in the Python optimization ecosystem.
```
hmgaudecker, line 140: *So what? If it was the standard term in all of applied math and only Python was off, that would be fine, I guess. So issue is that it is only used in econ?*

Having read on, I would suggest mentioning explicitly that the transition from scipy.optimize should be as simple as possible.

docs/source/development/eep-02-typing.md
```diff
@@ -0,0 +1,1474 @@
...
+
+The output of the criterion function becomes `float | FunctionValue`. There are
+decorators that help the user to write valid criterion functions without making an
+explicit instance of `FunctionValue`.
+
+The first two previous examples remain valid. The third one will be deprecated and
+should be replaced by:
+
+```python
+@em.mark.least_squares
```
hmgaudecker, line 152:
```suggestion
import estimagic as em
from estimagic import FunctionValue


@em.mark.least_squares
```

docs/source/development/eep-02-typing.md
```diff
@@ -0,0 +1,1474 @@
...
+  `root_contributions`, `contributions` and `value` even though any of them could be
+  constructed out of the `root_contributions`. This means we need to check that all of
+  them are compatible.
+- The name `criterion` is not used very widely in the Python optimization ecosystem.
+
+#### Proposal
+
+The output of the criterion function becomes `float | FunctionValue`. There are
+decorators that help the user to write valid criterion functions without making an
+explicit instance of `FunctionValue`.
```
hmgaudecker, line 146: Confusing without explaining at some point here what `FunctionValue` will look like.

docs/source/development/eep-02-typing.md
```diff
@@ -0,0 +1,1474 @@
...
+a pytree or vector. Since the decorator was used, we know how to interpret the output.
+
+```python
+@em.mark.least_squares
+def decorated_sphere(params: np.ndarray) -> np.ndarray:
+    return params
+```
+
+In this last syntax, the criterion function is implemented the same way as in existing
+least-squares optimizers (e.g. DFO-LS), which will make it very easy for new users of
```
hmgaudecker, line 177: I think this is less about the optimization algorithm than its (most popular) Python implementation?

docs/source/development/eep-02-typing.md
```diff
@@ -0,0 +1,1474 @@
...
+estimagic. Similarly, `em.mark.likelihood` will be available for creating criterion
+functions that are compatible with the BHHH optimizer.
```
hmgaudecker, line 179: Same.

docs/source/development/eep-02-typing.md
```diff
@@ -0,0 +1,1474 @@
...
+```
+
+We can exploit this deprecation cycle to rename `root_contributions` to `residuals`
+which is more in line with the literature. Similarly, `contributions` can be renamed to
+`loglikes`.
+
+Since there is no need to modify instances of `FunctionValue`, it should be immutable.
+
+If a user only wants to express the least-squares structure of the problem without
+logging any additional information, they can only return the least-squares residuals as
```
hmgaudecker, line 167: The logging bit is incomprehensible without additional info.

docs/source/development/eep-02-typing.md
```diff
@@ -0,0 +1,1474 @@
...
+```
+
+In this last syntax, the criterion function is implemented the same way as in existing
+least-squares optimizers (e.g. DFO-LS), which will make it very easy for new users of
+estimagic. Similarly, `em.mark.likelihood` will be available for creating criterion
+functions that are compatible with the BHHH optimizer.
+
+For completeness, we can implement an `em.mark.scalar` decorator, but this will be
+optional, i.e. if none of the decorators was used we'll assume that the problem is a
+scalar problem.
```
hmgaudecker, line 183: This is why the first two examples will continue to work, no? Explain as the first thing, rather than at the end.

docs/source/development/eep-02-typing.md
```diff
@@ -0,0 +1,1474 @@
...
+
+```{note}
+In principle, we could make the usage of the decorator optional if a `FunctionValue`
+instance is returned. However, then we still would need one criterion evaluation until
+we know whether the criterion function is compatible with the selected optimizer.
+```
+
+```{note}
+A more modern alternative to a `mark` decorator would be to use `typing.Annotated` to
```
hmgaudecker, line 192: Difficult to have both like in pytask?

hmgaudecker (commented):
Looks great! Gotta stop now, only skimmed most of the rest, seems very sensible to me!

docs/source/development/eep-02-typing.md
```diff
@@ -0,0 +1,1474 @@
...
+- `lower_bounds` and `upper_bounds` are replaced by `bounds` (as in optimization)
+- `numdiff_options` and `optimize_options` become dataclasses
+- `logging` and `log_options` get aligned with our proposal for optimization
+
+In the long run we plan a general overhaul of `MSM` estimation that provides better
+access to currently internal objects such as the MSM objective function.
+
+(design-philosophy)=
+
+## Design Philosophy
```
hmgaudecker, line 1325: I think this should go upfront, i.e., after "Motivation and resources". *(makes sense you wrote it last, but for a reader it is better to have upfront)*

docs/source/development/eep-02-typing.md
```diff
@@ -0,0 +1,1474 @@
...
+
+Potentially, we could also offer a `.All` attribute that returns a list of all currently
+selected algorithms. That way a user could for example loop over all `Bounded` and
+`GradientBased` `LeastSquares` algorithms and compare them in a criterion plot.
+
+These categories match nicely with our
+[algorithm selection tutorials](https://effective-programming-practices.vercel.app/scientific_computing/optimization_algorithms/objectives_materials.html).
+
+We can use
+[`dataclasses.make_dataclass`](https://docs.python.org/3/library/dataclasses.html#dataclasses.make_dataclass)
```
schroedk, line 429: Hey estimagic team, nice proposal. I had a discussion with @janosg about this part. Using `make_dataclass` will not give static type support, as the type is build dynamically at runtime.

This feature is a great idea, but my guess is it will take some effort. I would tackle it the following way:

1. For every category, add a class attribute to the `Algorithm` base class (this will just be a stub)
    ```python
     class Algorithm(ABC):
         is_gradient_based: bool
         ....
    ```
2. Use `__init__subclass__` to check that a subclass sets the attribute (also validate it), e.g.:
    ```python
     class Algorithm(ABC):
         is_gradient_based: bool
         def __init_subclass__(cls, **kwargs):
                if not hasattr(cls, "is_gradient_based"):
                    raise TypeError(...)
        
                if not isinstance(cls.is_gradient_based, bool):
                    raise ValueError(...)
                
                super().__init_subclass__(**kwargs)
    ```
3. Write a script, which search for subclasses of `Algorithm` and based on the class attributes builds a static data structure as code representing the wanted tree (I mean a script which generates the corresponding python code and writes it to a file). Instead of searching, one could also register the subclasses in `__init__subclas__`
4. Add a pre-commit hook, which calls the file generating script and validates the output.
5. Add tests to ensure the generated code is valid python code (e.g. correct indentation, etc.)

docs/source/development/eep-02-typing.md
```diff
@@ -0,0 +1,1474 @@
...
+   options supported by the selected algorithm. `Algorithm`s are immutable.
+1. Given an instance of an `Algorithm`, a user can easily create a modified copy of that
+   instance by using the `with_option` method.
+1. We can provide additional methods `with_stopping` and `with_convergence` that call
+   `with_option` internally but provide two additional features:
+   1. They validate that the option is indeed a stopping/convergence criterion.
+   1. They allow to omit the `convergence_` or `stopping_` at the beginning of the
+      option name and can thus reduce repetition in the option names.
+1. As before, the user can pass a global set of options to `maximize` or `minimize`. We
+   continue to support option dictionaries but also allow `AlgorithmOption` objects that
```
schroedk, line 563: It is a very good idea to have this for backwards compatibility. However, I would not show this in the documentation and rather give the user a more biased explanation for configuration (in favor of configuring the algorithms directly).

docs/source/development/eep-02-typing.md
```diff
@@ -0,0 +1,1474 @@
...
+overarching goals of the proposal are the following:
+
+- Better discoverability and autocomplete for users of estimagic.
+- Better readability of code due to type hints.
+- More robust code due to static type checking and use of stricter types in internal
+  functions.
+
+Achieving these goals requires more than adding type hints. estimagic is currently
+mostly [stringly typed](https://wiki.c2.com/?StringlyTyped) and full of dictionaries
+with a fixed set of required keys (e.g.
```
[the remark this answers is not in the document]
janosg: True, this is less a distinction than a list of examples. I made that clearer now.

docs/source/development/eep-02-typing.md
janosg: Moved it to the front, thanks!

docs/source/development/eep-02-typing.md
janosg: Our docs are actually not great about this. Need to change that soon. But I will explain it a bit better in the text.

docs/source/development/eep-02-typing.md
janosg: Agree. Was also redundant.

docs/source/development/eep-02-typing.md
janosg: The thing is that in all cases it would be sufficient to specify the output at the least aggregated level. For a leas-squares problem, only the `"root_contributions"` are necessary. estimagic can then calculate the `"contributions"` and `"value"` from the `"root_contributions"`. 

However users can also provide all three dictionary entries. If they do that, we check that their calculations coincide with ours. This is what I meant by compatible but I should have called it consistent.

The problem does not come from having criterion functions that are compatible with multiple classes of optimizers but from using a dictionary as signaling device for the output type. 

This is why the preferred way of specifying least-squares problems will be:

```python
@em.mark.least_squares 
def ls_sphere(params):
    return params
```

docs/source/development/eep-02-typing.md
janosg: Changed it.

docs/source/development/eep-02-typing.md
janosg: Agree, I completely rewrote that section.

docs/source/development/eep-02-typing.md
janosg: Agree, I hope it is clearer now.

docs/source/development/eep-02-typing.md
janosg: `DFO-LS` is an algorithm and a package. But I used `scipy.optimize.least_squares` as example now.

docs/source/development/eep-02-typing.md
janosg: Rewritten.

docs/source/development/eep-02-typing.md
janosg: I changed the proposal. It will become possible to signal the function type purely with type hints. However, we don't need `typing.Annotated` because we simply introduce more fine grained return types (e.g. `LeastSquaresFunctionValue`. With this, the type checker can do more than if we used `typing.Annotated`.

docs/source/development/eep-02-typing.md
```diff
@@ -0,0 +1,1479 @@
...
+  configuration and return types. This takes the best of both worlds: we get the safety
+  and static analysis that (in Python) can only be achieved using objects but the
+  beginner friendliness and freedom provided by functions. Example: Having a `minimize`
+  function, it is very easy to add the possibility of running minimizations with
+  multiple algorithms in parallel and returning the best value. Having a `.solve` method
+  on an algorithm object would require a whole new interface for this.
+
+## Type checkers and their configuration
+
+We choose mypy as static type checker and run it as part of our continuous integration.
```
timmens, line 1428: We could also think about running mypy using pre-commit hooks if it's not too slow. This is what JAX is doing (see their [.pre-commit-config.yaml](https://github.com/google/jax/blob/main/.pre-commit-config.yaml).

janosg:
@hmgaudecker, @timmens I completely rewrote the sections **The objective function** and **Custom derivatives**. You might want to have a look.

docs/source/development/eep-02-typing.md
janosg: Done.

docs/source/development/eep-02-typing.md
```diff
@@ -0,0 +1,1479 @@
...
+- Users cannot select algorithms without reading the documentation.
+
+#### Difficulties
+
+The usual solution to selecting algorithms in an autocomplete friendly way is an `Enum`.
+However, there are two difficulties that make this solution suboptimal:
+
+1. The set of available algorithms depends on the packages a user has installed. Almost
+   all algorithms come from optional dependencies and very few users install all
+   optional dependencies.
```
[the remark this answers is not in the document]
janosg: Yes. My proposal only provided autocomplete in jupyter notebooks which is not enough. As @schroedk pointed out, this problem is unsolvable in general.

docs/source/development/eep-02-typing.md
janosg: Hi @schroedk, welcome to estimagic and thanks for pointing this out. This is completely correct, my approach only worked in jupyter notebooks, which is not good enough. 

I think your proposal contains two separate points:
1. An approach to use a pre-commit hook to achieve static analysis and autocomplete for algorithm selection as well as filtering algorithms by category. That approach will also work in all other places where I suggested to use `make_dataclass`.
2. An alternative to the `mark.minimizer` decorator via class variables that are validated via `__init_subclass__`. 

I am completely convinced about the first point and will adjust my proposal. 

I will also add the second point as alternative to the decorator approach so we can have a discussion about the pros and cons of each idea. My main concern is that most people who have contributed algorithms to estimagic so far do not know the difference between class and instance variables but are very familiar with using decorators. It will be interesting to hear opinions on this.

docs/source/development/eep-02-typing.md
```diff
@@ -0,0 +1,1479 @@
...
+from dataclasses import dataclass
+from typing import Callable, Tuple
+import estimagic as em
+
+
+@dataclass(frozen=True)
+class ScalarProblemFunctions:
+    fun: Callable[[NDArray[float]], float]
+    jac: Callable[[NDArray[float]], NDArray[float]]
+    fun_and_jac: Callable[[NDArray[float]], Tuple[float, NDArray[float]]]
```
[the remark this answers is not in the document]
janosg: This is the internal problem representation, where  everything is numpy based.  It is actually a goal to be as strict as possible here because then internal algorithms can make strong assumptions and have to handle fewer cases. 

The only reason to deviate from that would be if we want to support end-to-end jax compatible (and differentiable) optimizers like the ones from jaxopt (which is now part of optax). However that would require huge changes, as everything we do inside the internal criterion function would have to be jax compatible. 

I wouldn't say it's impossible but some features (e.g. logging) would be hard or impossible and of course the user's problems would also have to be fully jax compatible (e.g. no DataFrames as params).

docs/source/development/eep-02-typing.md
```diff
@@ -0,0 +1,1479 @@
...
+  optimizer after we evaluate it once. This means that users see errors only very late.
+- While optional, in least-squares problems it is possible that a user specifies
+  `root_contributions`, `contributions` and `value` even though any of them could be
+  constructed out of the `root_contributions`. This means we need to check that all of
+  them are compatible.
+- The name `criterion` is not used very widely in the Python optimization ecosystem.
+
+#### Proposal
+
+The output of the criterion function becomes `float | FunctionValue`. There are
```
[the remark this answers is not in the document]
janosg: It's actually `float | PyTree[float] | FunctionValue`.

docs/source/development/eep-02-typing.md
```diff
@@ -0,0 +1,1479 @@
...
+scalar problem.
+
+```{note}
+In principle, we could make the usage of the decorator optional if a `FunctionValue`
+instance is returned. However, then we still would need one criterion evaluation until
+we know whether the criterion function is compatible with the selected optimizer.
+```
+
+```{note}
+A more modern alternative to a `mark` decorator would be to use `typing.Annotated` to
```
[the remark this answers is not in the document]
janosg: Proposal is updated. There is a solution based on type hints but it is not using Annotated.

docs/source/development/eep-02-typing.md
```diff
@@ -0,0 +1,1479 @@
...
+
+- A dict with required keys is brittle
+- Autodiff needs to be handled completely outside of estimagic
+- The names `criterion`, `derivative` and `criterion_and_derivative` are not aligned
+  with scipy and very long.
+
+#### Proposal
+
+1. We keep the three arguments but rename them to `fun`, `jac` and `fun_and_jac`.
+1. `jac` or `fun_and_jac` can also be a string `"jax"` or a more autocomplete friendly
```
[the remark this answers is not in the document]
janosg: Yes, exactly. The updated proposal contains an explicit example now.

docs/source/development/eep-02-typing.md
janosg: Agree and added a note.

docs/source/development/eep-02-typing.md
janosg: Added it.

docs/source/development/eep-02-typing.md
```diff
@@ -0,0 +1,1479 @@
...
+    # ...
+    algorithm=em.algorithms.scipy_lbfgsb,
+    algo_options=options,
+    # ...
+)
+```
+
+```{note}
+In my currently planned implementation, autocomplete will not work reliably for the
+copy constructors (`with_option`, `with_stopping` and `with_convergence). The main
```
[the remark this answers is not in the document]
janosg: Fixed manually.

docs/source/development/eep-02-typing.md
```diff
@@ -0,0 +1,1479 @@
...
+cannot remove our current error handling just because many of these errors could now be
+caught by static analysis.
+
+We can investigate using `jaxtyping`'s pytest hooks to enable runtime typecheckers like
+beartype during testing but it is not a priority for now.
+
+## Changes in documentation
+
+- No type hints in docstrings anymore
+- Only show new recommended ways of doing things, not deprecated ones
```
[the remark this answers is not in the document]
janosg: Added a comment with link.

janosg:
I incorporated all comments and think they made the proposal much better. 

So far there are no concerns, objections or contentious aspects. 

A list of all breaking changes and deprecations can be found [here](https://github.com/OpenSourceEconomics/estimagic/blob/14c92cb028cae8083a390bd5ac0d2d972cbb1830/docs/source/development/eep-02-typing.md#summary-of-breaking-changes). 

Now everyone has one more week to comment on the proposal. If you like the proposal, give this comment a 👍 or approve this PR if you have the rights to do so.

timmens (approved):
I think this will be a big step forward for `estimagic` and I am excited to see the new interfaces in action. Very nice proposal! :tada:

docs/source/development/eep-02-typing.md
```diff
@@ -0,0 +1,1766 @@
...
+
+
+def least_squares_sphere(params: np.ndarray) -> LeastSquaresFunctionValue:
+    out = LeastSquaresFunctionValue(
+        value=params, info={"p_mean": params.mean, "p_std": params.std()}
+    )
+    return out
+```
+
+I think we can support this version on top of the decorator approach but I would not
```
ChristianZimpelmann, line 309: Writing style seems off

docs/source/development/eep-02-typing.md
```diff
@@ -0,0 +1,1766 @@
...
+The **second example** also returns a float but uses a different format for the
+parameters:
+
+```python
+def dict_sphere(params: dict) -> float:
+    return params["a"] ** 2 + params["b"] ** 2
+```
+
+If the user wants the criterion function to be compatible with specialized optimizers
+for least-squares problems, the criterion function needs to return a dictionary. The
```
ChristianZimpelmann, line 146: I don't fully understand what the specialized optimizers require. Do I understand correctly that `"p_mean"` and `"p_std"` are not required, but just the least-squares residuals in `root_contributions`. Why does the criterion function needs to return a dictionary then?

ChristianZimpelmann (commented):
First comments after having read about 1/3

docs/source/development/eep-02-typing.md
```diff
@@ -0,0 +1,1766 @@
...
+| `convergence_scaled_gradient_tolerance`    | `convergence_gtol_scaled` |            |
+
+### Things we do not want to align
+
+- We do not want to rename `algorithm` to `method` because our algorithm names are
+  different from scipy, so people who switch over from scipy need to adjust their code
+  anyways.
+- We do not want to rename `algo_options` to `options` for the same reason.
+- I am not sure if `params` should be renamed to `x0`. It would align estimagic more
+  with scipy, but `params` is just easier to pronounce and use as a word than `x0`.
```
ChristianZimpelmann, line 1719: I agree

docs/source/development/eep-02-typing.md
```diff
@@ -0,0 +1,1766 @@
...
+    # ...
+)
+```
+
+```{note}
+In my currently planned implementation, autocomplete will not work reliably for the
+copy constructors (`with_option`, `with_stopping` and `with_convergence`). The main
+reason is that most editors do not play well with `functools.wraps` or any other means
+of dynamic signature creation. For more details, see the discussions about the
+[Internal Algorithm Interface](algorithm-interface).
```
ChristianZimpelmann, line 870: Hyperlink doesn't seem to work in this environment.

ChristianZimpelmann (approved):
Sounds like an important step forward!

I would actually prefer some of the old names over the new ones (derived from scipy / NlOpt) -- they are much more straight forward (e.g., if estimagic keeps `params` instead of `x0`, the meaning of `xtol_abs` isn't easy to understand. But I also see the advantage of aligning the labels closer to those packages and I am sure you have thought a lot about it.

docs/source/development/eep-02-typing.md
hmgaudecker: Syntax should be `[Internal Algorithm Interface](#algorithm-interface)`, I think.

hmgaudecker:
> I would actually prefer some of the old names over the new ones (derived from scipy / NlOpt) -- they are much more straight forward (e.g., if estimagic keeps `params` instead of `x0`, the meaning of `xtol_abs` isn't easy to understand. But I also see the advantage of aligning the labels closer to those packages and I am sure you have thought a lot about it.

Agreed. In the end, we are talking about a trade-off re user bases here:

1. Clarity, drawing in new users
2. Similarity with existing packages, drawing in switchers

So far, we mostly targeted the first group. I am sure the second group is larger when it comes to scipy, so aligning there might be useful. I am less convinced when it comes to NLOpt.

As a middle ground, can't we keep the old names and allow the other ones as aliases?

docs/source/development/eep-02-typing.md
hmgaudecker: Same here. Furthermore, to me `x` / `x0` gives the impression of a vector, not an arbitrary pytree.

amageh (approved):
Great work, thanks to everyone who added to this! 👏 

Some of my thoughts:
- I am not a big fan of some of the renaming changes (especially criterion --> fun) but I agree that alignment with other packages is advantageous and therefore I think it makes sense to do it.
- The proposed idea for algorithms sounds very cool, looking forward to seeing this implemented! I also really like the proposed changes for bounds and constraints. 

Congrats on this comprehensive (and well-written) EEP!

mpetrosian (approved):
Very cool proposed changes and well written proposal!

Fascinated by the pre-commit hook solution for building `Algorithm` classes!
Just minor points:
- Can you add examples for the usages of the jax-like decorators for numerical differentiation?
- In the renaming table, instead of `stopping_maxfun`, you probably meant to write `stopping_maxfev`.
- There is an inconsistency between the naming conventions used in the examples and the proposed names in the renaming table, but I wouldn't rewrite the examples for this. 
Looking forward to working on and using the new estimagic!

docs/source/development/eep-02-typing.md
janosg: Agree, thanks. I changed it.

docs/source/development/eep-02-typing.md
janosg: Thanks! I converted the example to two examples to make it clearer. But this partially just validates my feeling that our current way of specifying least-squares problems is very hard to describe.

ChristianZimpelmann: Thanks! Much clearer now 👍

docs/source/development/eep-02-typing.md
janosg: Thanks, then we have a decision. I clarified it in the text.

docs/source/development/eep-02-typing.md
janosg: It works on [readthedcos](https://estimagic--495.org.readthedocs.build/en/495/development/eep-02-typing.html#id11) 

Which environment are you using?

ChristianZimpelmann: Just the [preview here on GitHub](https://github.com/OpenSourceEconomics/estimagic/blob/14c92cb028cae8083a390bd5ac0d2d972cbb1830/docs/source/development/eep-02-typing.md#proposal-4)

janosg:
Thanks for all of your comments. It is very nice to see that we have such an active community and all of the comments really helped to make this proposal better!

So far the only topic that needs more discussion is the renaming. As @hmgaudecker  pointed out, this is a trade-off between absolute clarity on the one side and alignment with the ecosystem and brevity on the other side. 

I am for the renaming but don't have a super strong opinion here. 

I personally think that `fun` is not a very good name but I do prefer `jac` over `derivative`. 

I also find that the NlOpt inspired options are actually more readable than our very long name because ours are so long that they are just overwhelming the eye  (e.g. [here](https://github.com/OpenSourceEconomics/estimagic/blob/f430f807cebcb3ce71265a24fe6936db1b821faf/src/estimagic/optimization/scipy_optimizers.py#L357))

I am against having aliases. In the long run this is just too much maintenance and potential for confusion. We would do it in the deprecation phase. 

I am opening two polls in [Zulip](https://ose.zulipchat.com/#narrow/stream/221432-estimagic/topic/EEP-02.3A.20Static.20typing) so we can have a vote about this.

hmgaudecker:
> I also find that the NlOpt inspired options are actually more readable than our very long name because ours are so long that they are just overwhelming the eye (e.g. [here](https://github.com/OpenSourceEconomics/estimagic/blob/f430f807cebcb3ce71265a24fe6936db1b821faf/src/estimagic/optimization/scipy_optimizers.py#L357))

To bring something else to the table, what about a custom `convergence` object?

janosg:
> > I also find that the NlOpt inspired options are actually more readable than our very long name because ours are so long that they are just overwhelming the eye (e.g. [here](https://github.com/OpenSourceEconomics/estimagic/blob/f430f807cebcb3ce71265a24fe6936db1b821faf/src/estimagic/optimization/scipy_optimizers.py#L357))
> 
> To bring something else to the table, what about a custom `convergence` object?

Custom convergence objects would be nice. A super nice example of custom convergence objects that can be combined with boolean operations is in [pydvl](https://pydvl.org/stable/api/pydvl/value/stopping/). Another example is in [Mystic](https://mystic.readthedocs.io/en/latest/mystic.html#module-mystic.termination)

If I understand your proposal correctly, this would mean that we handle convergence for all optimizers inside estimagic instead of expecting individual optimizers to do it. 

The problem is that this is very hard to implement for wrapped optimizers. The only thing we could do for all optimizers is check convergece after every function evaluation. However, typically optimizers check convergence after every iteration. The difference is that one iteration typically entails multiple function evaluations and not all of them are expected to lead to a (substantial) improvement in the function value (exploration vs. exploitation). This a slow progress over a few function evaluations could lead to spurious convergence. 

On a side not: This is the reason why an optimizer architecture where each optimizer just proposes a next step and then all objective evaluations are done in one main loop for all optimizers is hard to do (despite it's conceptual appeal). 

Of course, we could only consider iterations that lead to an actual improvement in function values, but it does not eliminate all problems. 

We can tackle this at some point but I see it as independent of the current proposal. After all, we will need a naming scheme for those objects.

hmgaudecker:
I think I was just thinking of grouping related options in one object, instead of repeating "convergence" in every string!

janosg:
> I think I was just thinking of grouping related options in one object, instead of repeating "convergence" in every string!

Ah, we have the `algorithm.with_stopping` and `algorithm.with_convergence` copy constructors for that. They are described [here](https://estimagic--495.org.readthedocs.build/en/495/development/eep-02-typing.html#id11)

So far I did not propose to generally split the `algo_options` into three because this would mean that we become more different from scipy, but we could discuss that.

hmgaudecker (approved):
Great stuff!

PR #500, opened 2024-07-10
Cleanup

janosg:
In this PR we shed some weight before we start the renaming and implementation of enhancement proposals. 

## Remove the dashboard

Remove the bokeh based dashboard. We have not promoted or used it for years and want to replace it with something more useful. Whether that is another dashboard, a tensorboard integration or a realtime convergence plot in a notebook is open. 

Users who have used the dashboard can get similar functionality by using `criterion_plot` and `params_plot` on a log file inside a jupyter notebook. 

## Remove the simopt optimizers 

We added the simopt optimizers for noisy objective function about a year ago but could not get our wrappers to solve simple optimization problems reliably and precisely. We therefore remove them again.

Users with noisy objective functions can use `tranquilo`, `nag_bobyqa` or `nag_dfols` instead. 

## Remove the `derivative_plot`

The `derivative_plot` visualizes the internal of Richardson extrapolation based numerical derivatives. Since we remove Richardson extrapolation, we need to remove the plot as well.


README.md
```diff
@@ -28,8 +28,8 @@ provides functionality to perform statistical inference on estimated parameters.
 - The complete history of parameters and function evaluations can be saved in a database
   for maximum reproducibility.
 - Painless and efficient multistart optimization.
-- The progress of the optimization is displayed in real time via an interactive
-  dashboard.
+- The progress of the optimization is displayed in real time in `criterion_plots` and
+  `params_plots`.
```
timmens, line 33: Not entirely convinced by this. You also say something similar elsewhere (see other comments).

My main problem with this is twofold:

(1) You do not mention that this requires logging (but to be fair, it was not mentioned beforehand), and
(2) It's not displayed in real-time unless you refresh the plotting manually (or am I missing something)

janosg: I think for now it's the best thing we have so I wanted to point ex dashboard users to it.

timmens: Makes sense. I do not have a strong opinion on this, but I would rather write something like this:

> The progress of the optimization can be displayed in real time using criterion and
> parameter plots.

I also just realized that you write `criterion_plots` instead of `criterion_plot`. This is slightly confusing since there is no `criterion_plots` function.

janosg: Done.

docs/source/index.md
```diff
@@ -187,7 +187,7 @@ algorithms
   for maximum reproducibility. See [How to use logging]
 - Painless and efficient multistart optimization. See [How to do multistart]
 - The progress of the optimization is displayed in real time via an interactive
-  dashboard. See {ref}`dashboard`.
+  `criterion_plot` and `params_plot`.
```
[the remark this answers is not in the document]
janosg: Done.

README.md
```diff
@@ -28,8 +28,8 @@ provides functionality to perform statistical inference on estimated parameters.
 - The complete history of parameters and function evaluations can be saved in a database
   for maximum reproducibility.
 - Painless and efficient multistart optimization.
-- The progress of the optimization is displayed in real time via an interactive
-  dashboard.
+- The progress of the optimization can be displayed in in `criterion_plot` and
```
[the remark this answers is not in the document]
janosg: Actually not because it sounds too much like auto-updating plots.

PR #502, opened 2024-07-10
Rename estimagic to optimagic

janosg:
This PR renames estimagic to optimagic 

We keep the name estimagic as internal namespace for everything related to estimation. 

Examples:

```python
from optimagic import minimize 
from estimagic import estimate_ml
```

Importing optimization related functions from estimagic is deprecated and will be removed in version 0.6.0.


timmens (approved):
I think optimagic will benefit greatly from the renaming! Very nice PR as well!

src/optimagic/__init__.py
```diff
@@ -0,0 +1,49 @@
...
+    "traceback_report",
+    "slice_plot",
+    "criterion_plot",
+    "params_plot",
+    "count_free_params",
+    "check_constraints",
+    "OptimizeLogReader",
+    "OptimizeResult",
+    "LikelihoodResult",
+    "MomentsResult",
```
timmens, line 47: These `estimagic` results should not be here, and are not imported anyways.

PR #506, opened 2024-07-19
New bounds

janosg:
This PR implements the changes for specifying bound constraints on parameters from the typing proposal and aligns the bounds argument with scipy.


src/optimagic/optimization/optimize.py
```diff
@@ -1,3 +1,16 @@
+"""Public functions for optimization.
+
+This module defines the public functions `maximize` and `minimize` that will be called
+by users.
+
+Internally, `maximize` and `minimize` just pass all arguments through to `_optimize`
+and add a `direction` argument.
+
+`optimize` processes the input arguments and converts them to stricter types. It then
+calls `_internal_optimize` which does the actual optimization.
```
janosg, line 10: This is already outdated.

src/optimagic/parameters/bounds.py
```diff
@@ -0,0 +1,242 @@
...
+    specified (in the case where params is a DataFrame with bounds as a column).
+
+    Args:
+        params: The parameter pytree.
+        bounds: The lower and upper bounds.
+        registry: pybaum registry.
+
+    Returns:
+        Consolidated and flattened lower_bounds.
+        Consolidated and flattened upper_bounds.
```
janosg, line 98: @timmens check the docstring

src/optimagic/deprecations.py
```diff
@@ -95,3 +96,33 @@ def replace_and_warn_about_deprecated_algo_options(algo_options):
...
+    upper_bounds,
+    bounds,
+    soft_lower_bounds=None,
+    soft_upper_bounds=None,
+):
+    old_bounds = {
+        "lower": lower_bounds,
+        "upper": upper_bounds,
+        "soft_lower": soft_lower_bounds,
+        "soft_upper": soft_upper_bounds,
```
timmens, line 112: Should be `"lower_bounds"`, `"upper_bounds"`, `"soft_lower_bounds"`, ..., right?

src/optimagic/deprecations.py
```diff
@@ -95,3 +96,33 @@ def replace_and_warn_about_deprecated_algo_options(algo_options):
...
+    if old_present:
+        msg = (
+            f"Specifying bounds via the arguments {', '.join(old_present)} is "
+            "deprecated and will be removed in optimagic version 0.6.0 and later. "
+            "Please use the `bounds` argument instead."
+        )
+        warnings.warn(msg, FutureWarning)
+
+    if bounds is None and old_present:
+        bounds = Bounds(**old_bounds)
```
timmens, line 126: What about the case where both the new and old bounds are set. In the current implementation we use the new bounds, but do not throw a warning other than the FutureWarning above.

src/optimagic/parameters/bounds.py
```diff
@@ -0,0 +1,242 @@
...
+from optimagic.parameters.tree_registry import get_registry
+from dataclasses import dataclass
+from optimagic.typing import PyTree, PyTreeRegistry
+from scipy.optimize import Bounds as ScipyBounds
+from typing import Sequence
+from numpy.typing import NDArray
+from typing import Any
+
+
+@dataclass
```
timmens, line 17: Not frozen with replace method?

timmens (changes requested):
Very nice PR. Only a few minor questions.

src/optimagic/deprecations.py
janosg: No, but I need to adjust for it in the error message.

src/optimagic/deprecations.py
janosg: As discussed, this behavior is safe enough.

janosg: Same.

src/optimagic/parameters/bounds.py
janosg: Should be frozen, will change it.

timmens (approved):

PR #507, opened 2024-07-21
Update infrastructure

timmens:
- [x] Use hatch for packaging
- [x] Add infrastructure to release both estimagic and optimagic
- [x] Use mypy as pre-commit
- [x] Replace black and nbqa by ruff
- [x] Fix ruff warnings
- [x] Add token `PYPI_API_TOKEN_OPTIMAGIC` as secret to repo

### Open Questions:

1. Should we rename the old token secret from `PYPI_API_TOKEN` to `PYPI_API_TOKEN_ESTIMAGIC`. In this case we would need to create a new one.
    -> No
1. Find a better place than the new folder `.scripts` to store the python file `update-name-in-pyproject.py`. Proposal: Combine contents of `.scripts` folder with `.envs` folder in new folder called `.ci`.
    -> We will use .tools for everything now

### mypy discussion

For now, I have opted to keep both the mypy pre-commit and the mypy GitHub Action task. Given tools like [`pre-commit CI`](https://pre-commit.ci/), I see an advantage in not using the development environment for the hooks. However, since we are not entirely sure that this mypy call is doing the same thing that the mypy call from our dev environment is doing, I believe the GitHub Action task is necessary. Combined, we should (1) be able to gain some experience with pre-commit mypy, (2) ideally get quick errors before committing, and (3) be confident that all errors should be caught, at least after pushing.


janosg:
> 1. Should we rename the old token secret from `PYPI_API_TOKEN` to `PYPI_API_TOKEN_ESTIMAGIC`. In this case we would need to create a new one.

No, having `PYPI_API_TOKEN` and `PYPI_API_TOKEN_OPTIMAGIC` is clear enough and we don't need to spend time on the renaming. 

> Find a better place than the new folder .scripts to store the python file update-name-in-pyproject.py. Proposal: Combine contents of .scripts folder with .envs folder in new folder called .ci.

Yes, we should combine all pre-commit hooks in one folder but the resulting environment files should still stay in `.envs`. `.ci` is not a good name because we will soon have a lot more pre-commit hooks that are actually code generation tools and not related to quality assurance (see here for [details](https://estimagic.org/en/latest/development/eep-02-typing.html#step-2-achieving-autocomplete-without-too-much-typing)). We might also have related pytest hooks. I am almost indifferent between the names `.scripts` and `.tools` but would not get more specific.

.github/workflows/publish-to-pypi.yml
```diff
@@ -24,6 +24,37 @@ jobs:
...
+      - name: Install pypa/build
+        run: >-
+          python -m
+          pip install
+          build
+          toml
+          --user
+      - name: Set name entry in pyproject.toml to estimagic
+        run: >-
+          python .scripts/update-name-in-pyproject.py
```
janosg, line 50: After starting to work on the enhancement proposals I see the advantage of having `optimagic` and `estimagic` in one package. I think we should completely abandon plans to split at some time in the future. NumFocus is against it anyways. Then we also don't have to make double releases for very long.

I would therefore say, that we should add a `FutureWarning` to `estimagic.__init__.py` and `optimagic.__init__.py` when we release the `estimagic` package. The future warning could say something like: 

```txt
"estimagic has been renamed to optimagic. Please uninstall estimagic and install optimagic 
instead. Don't worry, your estimagic imports will still work if you install optimagic and simple
warnings will help you to adjust them for future releases.  If you use pip, the commands are
[please add uninstall and reinstall commands]. If you use conda or mamba, the commands
are [...]."
```

docs/source/development/ep-02-typing.md
```diff
@@ -1222,7 +1222,6 @@ class ScipyNelderMead(Algorithm):
     def _solve_internal_problem(
         self, problem: InternalProblem, x0: NDArray[float]
     ) -> InternalOptimizeResult:
-
```
janosg, line 1225: Do our new tools automatically format code snippets in the documentation?

environment.yml
```diff
@@ -33,15 +32,13 @@ dependencies:
   - sphinx-panels  # docs
   - sphinxcontrib-bibtex  # docs
   - seaborn  # dev, tests
-  - mypy  # dev, tests
+  - mypy>=1.10,<1.11.0  # dev, tests
```
janosg, line 35: Why the pin and do we need a reminder to remove it at some point?

pyproject.toml
```diff
@@ -1,16 +1,99 @@
+# ======================================================================================
+# Project metadata
+# ======================================================================================
+[project]
+name = "optimagic"
+description = "Tools to solve difficult numerical optimization problems."
+requires-python = ">=3.8"
```
janosg, line 7: >=3.10; We dropped other versions pretty aggressively.

pyproject.toml
```diff
@@ -1,16 +1,99 @@
...
+[tool.hatch.metadata]
+allow-direct-references = true
 
+
+# ======================================================================================
+# Ruff configuration
+# ======================================================================================
 [tool.ruff]
-target-version = "py37"
+target-version = "py38"
```
janosg, line 93: We could also move this up to 3.10.

janosg (commented):
Looks very good. Thanks. Just minor comments.

timmens:
Warning looks like this:

image

janosg (approved):
Thanks, very nice PR!

PR #508, opened 2024-07-30
Update scaling argument handling

timmens:
In this PR, I update the scaling argument handling.

- [x] Throw deprecation warnings if `scaling_options` argument is used
- [x] Allow to pass `ScalingOptions` to `scaling` argument
- [x] Pre-process scaling input and throw errors if it fails
- [x] Update documentation
- [x] Inline code from `check_arguments.py`
- [x] Update scale converter

> [!NOTE]
> I chose to stick with the `ScalingOptionsDict`, as it gives you autocomplete in the editor. See below. 

image


src/optimagic/exceptions.py
```diff
@@ -47,6 +47,10 @@ class InvalidBoundsError(OptimagicError):
     """Exception for invalid user provided bounds."""
 
 
+class InvalidScalingOptionsError(OptimagicError):
```
janosg, line 50: I would rename this to `InvalidScalingError` so it covers a slightly wider range of errors.

src/optimagic/parameters/scale_conversion.py
```diff
@@ -2,24 +2,29 @@
 from typing import Callable, NamedTuple
 
 import numpy as np
+from numpy.typing import NDArray
 
+from optimagic.parameters.scaling import ScalingOptions
 from optimagic.parameters.space_conversion import InternalParams
 
 
+class ScaleConverter(NamedTuple):
```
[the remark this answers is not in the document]
janosg: Also, let's get rid of NamedTuples and use frozen dataclasses instead. Not just here but in all places we touch during the current sprint.

docs/source/how_to/how_to_scaling.md
janosg, line 1: 1. I think after the "What does well scaled mean" section we should add a "TLDR" section with a code snippet that just sets `scaling=True` and one sentence explaining that this uses the start values heuristic described below, that it works for all problems but there might be better ones. 

2. We should mostly use the word "objective function" instead of "criterion"

janosg (commented):
Very nice but gave me some additional ideas we need to discuss.

tests/optimagic/optimization/test_with_scaling.py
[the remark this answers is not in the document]
timmens: Choose `test_with_scaling.py` but add check that optimization returns correct result.

src/optimagic/parameters/scaling.py
```diff
@@ -0,0 +1,78 @@
...
+
+
+@dataclass(frozen=True)
+class ScalingOptions:
+    method: Literal["start_values", "bounds"] = "start_values"
+    clipping_value: float = 0.1
+    magnitude: float = 1.0
+
+
+class ScalingOptionsDict(TypedDict):
```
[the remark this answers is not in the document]
janosg: I'm fully convinced.

src/optimagic/parameters/scale_conversion.py
```diff
@@ -1,30 +1,65 @@
...
+        internal_params: NamedTuple of internal and possibly reparametrized but not yet
+            scaled parameter values and bounds.
+        scaling: Scaling options. If None, no scaling is performed.
 
     Returns:
-        ScaleConverter: NamedTuple with methods to convert between scaled and unscaled
+        ScaleConverter: Dataclass with methods to convert between scaled and unscaled
             internal parameters and derivatives.
-        InternalParams: NamedTuple with entries:
+        InternalParams: Dataclass with entries:
```
janosg, line 62: We don't need to list all the attributes of `InternalParams` as it should be documented by that class.

src/optimagic/parameters/scaling.py
```diff
@@ -0,0 +1,78 @@
+from dataclasses import dataclass
+from typing import Literal, TypedDict
+
+from typing_extensions import NotRequired
+
+from optimagic.exceptions import InvalidScalingError
+
+
+@dataclass(frozen=True)
+class ScalingOptions:
```
janosg, line 10: We should start adding docstrings to user facing Options classes.

janosg (approved):

PR #510, opened 2024-07-31
Activate isort rule in ruff and bump mypy version

timmens:
- Activate isort rule in ruff (rule "I")
- Bump mypy version to 1.11
  This entails a behavior change compared to 1.10; see diff [here](https://github.com/OpenSourceEconomics/optimagic/pull/510/commits/5f4108c66f3439c84549e4fc37c5c4d5175ada31).


janosg (approved):
Thanks!

PR #513, opened 2024-08-02
Update multistart option handling

timmens:
In this PR, I update the multistart options handing.

- [x] Add public `MultistartOptions` data class
- [x] Add internal `MultistartInfo` data class
- [x] Add a lot of type and value checks to the public data class
- [x] Add deprecations and exceptions
- [x] Only use `MultistartOptions` instead of dictionaries internally
- [x] Add deprecation test
- [x] Change default `"sampling_method"` to `"random"`. 
- [x] Validate that RNG can be used as seeds.


src/optimagic/optimization/multistart.py
```diff
@@ -12,20 +12,388 @@
...
+            because they do not have the correct type.
+
+    """
+
+    n_samples: int | None = None
+    n_optimizations: int | None = None
+    sampling_distribution: Literal["uniform", "triangular"] = "uniform"
+    sampling_method: MultistartSamplingMethod = "random"
+    sample: Sequence[PyTree] | None = None
+    mixing_weight_method: Literal["tiktak", "linear"] = "tiktak"
```
janosg, line 94: Since this is re-used in the TypedDict below, the literal type should maybe be assigned to a variable.

src/optimagic/optimization/multistart.py
```diff
@@ -12,20 +12,388 @@
...
+            "be 'raise' or 'continue'."
+        )
+
+
+# Internal Options
+# ======================================================================================
+
+
+@dataclass
+class InternalMultistartOptions:
```
janosg, line 296: As additional protection this should maybe have a `__post_init__` that does some checks. Overhead is not a concern for things that happen just once in an entire multistart optimization.

src/optimagic/optimization/multistart.py
janosg, line 1: The first ~400 lines should be extracted into a separate file called `multistart_options.py` as it is all related to option handling. If necessary, we can have a multistart directory.

src/optimagic/optimization/multistart.py
```diff
@@ -39,26 +407,26 @@ def run_multistart_optimization(
...
             upper=upper_sampling_bounds,
             # -1 because we add start parameters
-            n_samples=options["n_samples"] - 1,
-            sampling_distribution=options["sampling_distribution"],
-            sampling_method=options["sampling_method"],
-            seed=options["seed"],
+            n_samples=options.n_samples - 1,
+            sampling_distribution=options.sampling_distribution,
+            sampling_method=options.sampling_method,
+            seed=options.seed,
```
janosg, line 429: We should not need that many individual arguments. We can either pass in the entire InternalMultistartOptions or if you think this is not transparent enough, have more nested InternalMultistartOptions such that we could pass in something like `internal_options.sampling`.

src/optimagic/optimization/multistart.py
```diff
@@ -12,20 +12,388 @@
...
+            are "raise" and "continue".
+
+    Raises:
+        InvalidMultistartError: If the multistart options cannot be processed, e.g.
+            because they do not have the correct type.
+
+    """
+
+    n_samples: int | None = None
+    n_optimizations: int | None = None
```
janosg, line 90: `n_optimizations` should be renamed to `stopping_maxopt` to align it with our new naming scheme and make clear that this number might not be reached.

src/optimagic/optimization/multistart.py
```diff
@@ -12,20 +12,388 @@
...
+    sampling_distribution: Literal["uniform", "triangular"] = "uniform"
+    sampling_method: MultistartSamplingMethod = "random"
+    sample: Sequence[PyTree] | None = None
+    mixing_weight_method: Literal["tiktak", "linear"] = "tiktak"
+    mixing_weight_bounds: tuple[float, float] = (0.1, 0.995)
+    convergence_relative_params_tolerance: float = 0.01
```
janosg, line 96: `convergence_relative_params_tolerance` -> `convergence_xtol_rel` for alignment.

src/optimagic/optimization/multistart.py
```diff
@@ -12,20 +12,388 @@
...
+    convergence_max_discoveries: int = 2
+    n_cores: int = 1
+    batch_evaluator: Literal["joblib", "pathos"] | Callable = "joblib"
+    batch_size: int | None = None
+    seed: int | np.random.Generator | None = None
+    exploration_error_handling: Literal["raise", "continue"] = "continue"
+    optimization_error_handling: Literal["raise", "continue"] = "continue"
```
janosg, line 103: I think one error handling argument is enough.

tests/optimagic/parameters/test_scaling.py
```diff
@@ -1,6 +1,33 @@
...
 from optimagic.exceptions import InvalidScalingError
-from optimagic.parameters.scaling import ScalingOptions, pre_process_scaling
+from optimagic.parameters.scaling import (
+    ScalingOptions,
+    ScalingOptionsDict,
+    pre_process_scaling,
+)
+
+
+def test_scaling_options_and_dict_have_same_attributes():
```
janosg, line 12: Since this will occur in many places we should have one test file just dedicated to comparing TypedDicts and dataclasses and make the comparison function a bit more abstract.

janosg (commented):
Very nice PR. 

After you are done with the current comments I will go over the how-to guide; Will be easier to make changes there myself than describing changes in comments. 

One more thing: Since it is a breaking change we should convert the `multistart_info` dict in the OptimizeResult to a dataclass. You can add a `__getitem__` method such that it behaves a bit more like a dict.

src/optimagic/optimization/multistart_options.py
```diff
@@ -0,0 +1,482 @@
...
+        "raise",
+        "continue",
+    ):
+        raise InvalidMultistartError(
+            f"Invalid error handling: {options.error_handling}. Error handling must be "
+            "'raise' or 'continue'."
+        )
+
+
+def _replace_and_warn_about_deprecated_attributes(
```
janosg, line 281: Can we move this function into `deprecations.py` (called `replace_and_warn_about_deprecated_multistart_options`)?

src/optimagic/optimization/multistart_options.py
```diff
@@ -0,0 +1,482 @@
...
+    options: MultistartOptions,
+) -> MultistartOptions:
+    replacements: MultistartOptionsDict = {}
+
+    if options.share_optimization is not None:
+        msg = (
+            "The share_optimization attribute is deprecated and will be removed in "
```
janosg, line 288: For users, "share_optimization" is currently not an attribute but an entry in a dictionary. In general I prefer the word "option" instead of "attribute" for this kind of stuff.

janosg (approved):
Thanks a lot. We can merge after two mini comments.

PR #514, opened 2024-08-05
Add `NumdiffResult` object

timmens:
- [x] Update pre-commit hooks and add `numpy` to pre-commit `mypy` environment
- [x] Add `NumdiffResult` object
- [x] Add type hints to the first and second derivative outer interface
- [x] Deprecate Richardson extrapolation options as well as `return_func_value` argument
- [x] Test deprecations
- [x] Update documentation


docs/source/how_to/how_to_first_derivative.ipynb
janosg, line 1: In its current form I would almost say the how-to-guides are redundant. The tutorial goes deeper than the how-to guide when it comes to the different finite-differences methods. And none of them mentions the mathematical background: https://estimagic.org/en/latest/explanations/differentiation/background_numerical_differentiation.html

Maybe we should just improve the tutorial for now and later add specific how-to-guides like "How to choose a finite-differences method" or "How to choose step-sizes for numerical derivatives." But we should make issues so we don't forget about this.

docs/source/tutorials/numdiff_overview.ipynb
janosg, line 1: I suggest two improvements:

Add a reference to https://estimagic.org/en/latest/explanations/differentiation/background_numerical_differentiation.html when you talk about choosing finite differences methods. 

Try to mention parallelization as early as possible. It is one of the coolest features.

src/optimagic/differentiation/derivatives.py
```diff
@@ -1,53 +1,110 @@
...
+    def func_evals(self) -> pd.DataFrame | dict[str, pd.DataFrame | None] | None:
+        deprecations.throw_numdiff_result_func_evals_future_warning()
+        return self._func_evals
+
+    @property
+    def derivative_candidates(self) -> pd.DataFrame | None:
+        deprecations.throw_numdiff_result_derivative_candidates_future_warning()
+        return self._derivative_candidates
+
+    def __getitem__(self, key: str) -> Any:
```
janosg, line 75: I think we should deprecate the dict access right away. It is only there so less old code is broken.

janosg (commented):
Very nice PR. I just have comments about the documentation.

docs/source/tutorials/numdiff_overview.ipynb
```diff
@@ -6,7 +6,9 @@
    "source": [
     "# Numerical differentiation\n",
     "\n",
-    "Using simple examples, This tutorial shows you how to numerical differentiation with optimagic. More details on the topics covered here can be found in the [how to guides](../how_to/index.md)."
+    "In this tutorial, you will learn how to numerically differentiate functions with\n",
+    "optimagic. More details on the topics covered here can be found in the\n",
+    "[how to guides](../how_to/index.md)."
```
janosg, line 11:
```suggestion
    "optimagic."
```

janosg (approved):

PR #517, opened 2024-08-08
Update numdiff option handling

timmens:
- Create a `NumdiffOptions` class
- Update documentation
- Deprecate `numdiff_options` in estimation functions
- Allow PyTrees in derivative arguments


janosg (approved):
Thanks!

PR #519, opened 2024-08-09
Refactor/logging

schroedk:
The proposed logging is described [here](https://estimagic.org/en/latest/development/eep-02-typing.html#logging).

This refactoring introduces the following abstractions:

- `LogOptions` (implementation `SQLiteLogOptions`), configuration
- `LogStore`(implementation `_SQLiteLogStore`, not exposed), write access
-  `LogReader` (implementation `SQLiteLogReader`), read access

The old class `OptimizationLogReader` redirects to `SQLiteLogReader` in a backward-compatible way.


src/optimagic/logging/logger.py
```diff
@@ -0,0 +1,320 @@
...
+            rowid = highest_rowid + iteration + 1
+
+        row_list = self.iteration_store.select(rowid)
+
+        if len(row_list) == 0:
+            raise IndexError(f"Invalid iteration requested: {iteration}")
+        else:
+            data = row_list[0]
+
+        params = tree_unflatten(
```
janosg, line 108: I think there is actually not good reason for storing the parameters in a flat format. 

We define `external_x` (i.e. flattened but not reparametrized params) [here](https://github.com/OpenSourceEconomics/optimagic/blob/354155d38157c861ae8988e45301237a577096d1/src/optimagic/optimization/internal_criterion_template.py#L101) and then just use it for logging. But we also have a non-flat `current_params` defined. If we write that into the database, we can get rid of the unflattening here. 

There might have been a reason in the dashboard code but that reason is definitely no longer valid.

src/optimagic/logging/logger.py
```diff
@@ -0,0 +1,320 @@
...
+                criterion values, and runtimes.
+
+        """
+        raw_res = self.iteration_store.select()
+        params_list = []
+        criterion_list = []
+        runtime_list = []
+        for data in raw_res:
+            if data.value is not None:
+                params = tree_unflatten(
```
janosg, line 128: `tree_unflatten` would also be unnecessary here.

src/optimagic/logging/base.py
```diff
@@ -0,0 +1,214 @@
...
+        ValueError: If input_type or output_type is not a dataclass, or if
+                the primary key is not found in output_type fields.
+
+    """
+
+    def __init__(
+        self,
+        input_type: Type[InputType],
+        output_type: Type[OutputType],
+        primary_key: str,
```
janosg, line 38: The word primary key has been used for a different concept in optimagic. We are getting rid of that concept now but it might still be confusing to some people. However, I don't have a better name and in the database world primary_key is the right name.

src/optimagic/logging/logger.py
```diff
@@ -0,0 +1,320 @@
...
+
+    @staticmethod
+    def _normalize_direction(
+        direction: OptimizationType | OptimizationTypeLiteral,
+    ) -> OptimizationType:
+        if isinstance(direction, str):
+            direction = OptimizationType(direction)
+        return direction
+
+    def _build_history_dataframe(self) -> pd.DataFrame:
```
janosg, line 148: What are the usages of the history DataFrame? I wonder if we could get rid of that format as a DataFrame is not really a good way to store `params`, which can be anything.

src/optimagic/logging/logger.py
```diff
@@ -0,0 +1,320 @@
...
+        )
+
+    def read_start_params(self) -> PyTree:
+        """Read the start parameters form the problem store.
+
+        Returns:
+            A pytree object representing the start parameter.
+
+        """
+        return self.problem_store.select_last_rows(1)[0].params
```
janosg, line 281: Why does `select_last_rows(1)` return the start params? The start params should be the first, not the last entry.

src/optimagic/typing.py
```diff
@@ -12,3 +13,56 @@ class AggregationLevel(Enum):
     SCALAR = "scalar"
     LEAST_SQUARES = "least_squares"
     LIKELIHOOD = "likelihood"
+
+
+@dataclass(frozen=True)
+class DictLikeAccess:
```
janosg, line 19: Very nice, we can use that in many places.

janosg (commented):
Thanks a lot for your first PR!

I am not completely through but here is a first round of comments we can discuss tomorrow. Many are related to quirks of the original logging, i.e. not caused by your PR but hopefully fixed by your PR.

src/optimagic/logging/logger.py
```diff
@@ -0,0 +1,431 @@
...
+            cursor = dbapi_connection.cursor()
+            cursor.execute("PRAGMA journal_mode = WAL")
+            if self._fast_logging:
+                cursor.execute("PRAGMA synchronous = OFF")
+            else:
+                cursor.execute("PRAGMA synchronous = NORMAL")
+            cursor.close()
+
+
+class SQLiteLogger(Logger):
```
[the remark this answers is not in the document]
janosg: Actually, I think there is a problem with default values, now that we have no separation between a Logger and a LogReader. When a Logger is used for reading, it should of course not delete anything. But when it is provided to `minimize`, the most common usage is to replace existing files or tables.

janosg (commented):
I am through now. Thanks a lot for this PR!

One thing that is left to do is adding a `logger` field to the `OptimizeResult`.

src/optimagic/deprecations.py
```diff
@@ -461,6 +466,39 @@ def replace_and_warn_about_deprecated_derivatives(candidate, name):
...
+    log_options_is_compatible = set(log_options.keys()).issubset(compatible_keys)
+
+    if logging_is_path_or_string:
+        if log_options_is_dict and log_options_is_compatible:
+            warnings.warn(
+                f"\nUsing {log_options=} to create an instance of SQLiteLogger. "
+                f"This mechanism will be removed in the future."
+            )
+            return SQLiteLogger(cast(str | Path, logger), **log_options)
+        elif not log_options_is_compatible:
```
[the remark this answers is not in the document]
schroedk: Agree, will do

src/optimagic/logging/logger.py
schroedk: Agree, will do

src/optimagic/logging/logger.py
schroedk: Agree, will do

src/optimagic/logging/__init__.py
```diff
@@ -0,0 +1,2 @@
+from .logger import SQLiteLogger as SQLiteLogger
```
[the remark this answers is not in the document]
schroedk: Agree, will do.

src/optimagic/logging/logger.py
```diff
@@ -0,0 +1,431 @@
...
+            database file. One of “extend”, “replace”, “raise”.
+
+    """
+
+    def __init__(
+        self,
+        path: str | Path,
+        fast_logging: bool = True,
+        if_database_exists: ExistenceStrategy
+        | ExistenceStrategyLiteral = ExistenceStrategy.RAISE,
```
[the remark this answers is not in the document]
schroedk: Outdated

schroedk:
> I am through now. Thanks a lot for this PR!
> 
> One thing that is left to do is adding a `logger` field to the `OptimizeResult`.

Done

tests/optimagic/test_deprecations.py
```diff
@@ -862,6 +865,43 @@ def test_fun_with_dict_return_is_deprecated_in_slice_plot():
...
+
+    incompatibility_msg = "Found string or path for logger argument, but parameter"
+    f" {log_options=} is not compatible "
+    log_options_typo = {"fast_lugging": False}
+
+    with pytest.raises(ValueError, match=incompatibility_msg):
+        handled_logger = handle_log_options_throw_deprecated_warning(
+            log_options_typo, ":memory:"
+        )
+        assert handled_logger == ":memory:"
```
janosg, line 902: We are missing tests that make sure the old `log_options` still work but lead to a `FutureWarning` in:
- minimize
- maximize
- estimate_ml
- estimate_msm

Starting points for testing the estimation functions are [here](https://github.com/OpenSourceEconomics/optimagic/blob/a612912a7a186100fcf2bb1761f8d064ee0b4781/tests/optimagic/test_deprecations.py#L476)

janosg: We also need such a test for using the old `LogReader`

src/optimagic/typing.py
```diff
@@ -12,3 +16,88 @@ class AggregationLevel(Enum):
...
+
+    def __len__(self) -> int:
+        return len(fields(self))
+
+    def __iter__(self) -> Iterator[str]:
+        for field in fields(self):
+            yield getattr(self, field.name)
+
+
+class OptimizationType(str, Enum):
```
janosg, line 66: I have created this same Enum but called it `Direction` because `OptimizationType` could also be multistart vs. single and other things.

src/optimagic/logging/types.py
```diff
@@ -0,0 +1,189 @@
...
+    RAISE = "raise"
+    EXTEND = "extend"
+    REPLACE = "replace"
+
+
+ExistenceStrategyLiteral = Literal["raise", "extend", "replace"]
+
+
+@dataclass(frozen=True)
+class CriterionEvaluationResult(DictLikeAccess):
```
janosg, line 68: Maybe rename to `OptimizerState`, `IterationState` or `OptimizerIterationState`; Actually I think you suggested that at some point and I like it. `CriterionEvaluationResult` is too narrow. Then there might be follow-up renamings (e.g. `iteration_store` to `state_store`

src/optimagic/typing.py
schroedk: I would keep it like it is here, and would like to ask you to resolve the conflict in the upcoming PR:)

src/optimagic/logging/types.py
schroedk: Agree, will do

tests/optimagic/test_deprecations.py
schroedk: Done

janosg (approved):
Thanks a lot! From my side this is ready to merge!

PR #522, opened 2024-08-12
Update constraint interface

timmens:
Update constraints interface according to EEP-02 (changes described [here](https://estimagic.org/en/latest/development/eep-02-typing.html#constraints))

- Also adds some tests for pytree handling in derivatives


src/estimagic/estimate_ml.py
```diff
@@ -179,6 +185,10 @@ def estimate_ml(
     bounds = pre_process_bounds(bounds)
     jacobian_numdiff_options = pre_process_numdiff_options(jacobian_numdiff_options)
     hessian_numdiff_options = pre_process_numdiff_options(hessian_numdiff_options)
+    # TODO: Overwrite constraints once we deprecate dictionary constraints. We currently
+    # pass constraints unprocessed to the optimizer, so that we do not throw warnings
+    # if the new constraints are used.
+    dict_constraints = pre_process_constraints(constraints)
```
janosg, line 191: I would move this function into `deprecations.py` such that we get an error once we delete `deprecations.py`.

src/estimagic/estimate_msm.py
```diff
@@ -175,11 +176,20 @@ def estimate_msm(
...
+
     # ==================================================================================
     # Check and process inputs
     # ==================================================================================
 
     bounds = pre_process_bounds(bounds)
+    # TODO: Overwrite constraints once we deprecate dictionary constraints. We currently
+    # pass constraints unprocessed to the optimizer, so that we do not throw warnings
+    # if the new constraints are used.
+    dict_constraints = pre_process_constraints(constraints)
```
janosg, line 192: Same.

src/optimagic/constraints.py
```diff
@@ -0,0 +1,246 @@
...
+
+        if self.value is not None and not isinstance(self.value, float | int):
+            raise InvalidConstraintError("'value' must be a float or an int.")
+
+
+@dataclass(frozen=True)
+class NonlinearConstraint(Constraint):
+    selector: Callable[[PyTree], PyTree] = identity_selector
+    _: KW_ONLY
+    func: Callable[[PyTree], ArrayLike | "pd.Series[float]" | float] | None = None
```
janosg, line 173: Why `"pd.Series[float]"` instead of `pd.Series[pd.FloatDtype....]`?

src/optimagic/constraints.py
```diff
@@ -0,0 +1,246 @@
...
+        if not callable(self.selector):
+            raise InvalidConstraintError("'selector' must be callable.")
+
+
+def identity_selector(x: PyTree) -> PyTree:
+    return x
+
+
+@dataclass(frozen=True)
+class FixedConstraint(Constraint):
```
janosg, line 32: Can you write docstrings for each constraint? Should be faster now than in a month.

src/optimagic/constraints.py
```diff
@@ -0,0 +1,246 @@
...
+            raise InvalidConstraintError("'tol' must be non-negative.")
+
+        if self.func is None or not callable(self.func):
+            raise InvalidConstraintError("'func' must be callable.")
+
+        if self.derivative is not None and not callable(self.derivative):
+            raise InvalidConstraintError("'derivative' must be callable.")
+
+
+def pre_process_constraints(
```
janosg, line 220: This function should live in deprecations, so we are aware that it needs to be rewritten at some point.

src/optimagic/deprecations.py
```diff
@@ -325,6 +325,15 @@ def throw_key_warning_in_derivatives():
     warnings.warn(msg, FutureWarning)
 
 
+def throw_dict_constraints_future_warning():
+    msg = (
+        "Specifying constraints as a dictionary is deprecated and will be removed in "
+        "optimagic version 0.6.0. Please use a constraint object imported from "
+        "optimagic.constraints instead."
```
janosg, line 332: This should take constraints as argument and give a more constructive error message. Similar to what we do [here](https://github.com/OpenSourceEconomics/optimagic/blob/19029f53dfe4a599bb4f121eea3ed3fabf54f3a3/src/optimagic/deprecations.py#L169) but only warn, not replace.

janosg: Then the control flow to check whether a warning is necessary can also go into the function.

src/optimagic/optimization/create_optimization_problem.py
```diff
@@ -474,8 +481,9 @@ def create_optimization_problem(
         if not isinstance(bounds, Bounds | None):
             raise ValueError("bounds must be a Bounds object or None")
 
-        if not isinstance(constraints, list | dict):
-            raise ValueError("constraints must be a list or a dictionary")
+        if not all(isinstance(c, dict) for c in constraints):
+            # TODO: Only allow list[Constraint]
+            raise ValueError("constraints must be a list of dictionaries")
```
janosg, line 486: This is confusing now that dict are deprecated. Maybe write: "constraints must be a list of **Constraint objects or** dictionaries.")

tests/optimagic/differentiation/test_derivatives.py
janosg, line 1: Did you find anything through these tests?

tests/optimagic/test_deprecations.py
```diff
@@ -856,3 +856,54 @@ def test_fun_with_dict_return_is_deprecated_in_slice_plot():
...
+
+        em.estimate_ml(
+            loglike=loglike,
+            params=np.arange(3),
+            optimize_options={"algorithm": "scipy_lbfgsb"},
+            constraints={"type": "fixed", "loc": [0, 1]},
+        )
+
+
+def test_dict_constraints_are_deprecated_in_estimate_msm():
```
janosg, line 899: What about `check_constraints` and `count_free_params`? They are [here](https://github.com/OpenSourceEconomics/optimagic/blob/0.5.0/src/optimagic/parameters/constraint_tools.py)

src/optimagic/constraints.py
timmens: image

timmens: Mypy can read the information from `"pd.Series[float]"`.

tests/optimagic/differentiation/test_derivatives.py
timmens: No, it worked.

src/estimagic/estimate_ml.py
```diff
@@ -179,6 +184,9 @@ def estimate_ml(
     bounds = pre_process_bounds(bounds)
     jacobian_numdiff_options = pre_process_numdiff_options(jacobian_numdiff_options)
     hessian_numdiff_options = pre_process_numdiff_options(hessian_numdiff_options)
+    # TODO: Replace dict_constraints with constraints, once we deprecate dictionary
+    # constraints.
+    dict_constraints = pre_process_constraints(constraints)
```
janosg, line 189: I would prefer to read `deprecations.pre_process_constraints` to make it even more visible that this is something that will need to be changed. I use `deprecations.func` almost everywhere instead of importing those functions one by one.

src/optimagic/deprecations.py
```diff
@@ -325,6 +326,43 @@ def throw_key_warning_in_derivatives():
...
+    }
+
+    if not isinstance(constraints, list):
+        constraints = [constraints]
+
+    types = []
+    for constraint in constraints:
+        if isinstance(constraint, dict):
+            types.append(constraint.get("type", None))
+
```
janosg, line 353:
```suggestion
    types = list(set(types) - {None})
    
```

src/optimagic/deprecations.py
```diff
@@ -421,3 +459,39 @@ def replace_and_warn_about_deprecated_derivatives(candidate, name):
             out.append(key_to_marker[key](func))
 
     return out
+
+
+def pre_process_constraints(
```
janosg, line 464: This will need a docstring that explains why we convert everything to dict now and that we should go the opposite way once we have time to refactor the internals.

test.py
janosg, line 1: You probably did not want to commit this.

janosg (approved):
Thanks!

PR #531, opened 2024-09-12
small changes to explanation files in documentation

ChristianZimpelmann:
In addition to the minor changes which I already implemented:

- I really like the visualizations in `explanation_of_numerical_optimizers.md`. However, it would be great if there was a way to manually click through the images such that the user can speed it up or revert to a previous figure. With the current implementation via gifs it takes quite a while to see the full cycle.

### DIFF
--- a/docs/source/explanation/explanation_of_numerical_optimizers.md
+++ b/docs/source/explanation/explanation_of_numerical_optimizers.md
@@ -14,8 +14,9 @@ The main principles we describe here are:
 - Derivative free trust region algorithms
 - Derivative free direct search algorithms
 
-This covers a large range of the algorithms that come with optimagic. We do currently
-not cover:
+This covers a large range of the algorithms that come with optimagic. In contrast, the
+following classes of optimizers are also accessible via optimagic, but not yet covered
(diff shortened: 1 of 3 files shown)


docs/source/explanation/explanation_of_numerical_optimizers.md
```diff
@@ -14,8 +14,9 @@ The main principles we describe here are:
 - Derivative free trust region algorithms
 - Derivative free direct search algorithms
 
-This covers a large range of the algorithms that come with optimagic. We do currently
-not cover:
+This covers a large range of the algorithms that come with optimagic. In contrast, the
```
ChristianZimpelmann, line 17: Not sure whether I got this right, but the current version is unclear about where these optimizers are not covered (in this overview or in optimagic in general)

janosg:
Thanks @ChristianZimpelmann. I will merge my current PR first, which should fix the tests. Then I'll have a look. 

I like your suggestion for more interactive plots but I am not sure we will have time to implement it anytime soon.

janosg (approved):

PR #534, opened 2024-09-24
Implement filtered autocomplete for algorithm selection

janosg:
## Short description

After this PR, algorithms can for example be selected as follows:

```python
import optimagic as om
import numpy as np 
 
om.minimize(
    fun=lambda x: x @ x,
    params=np.arange(3),
    algorithm=om.algos.Bounded.scipy_lbfgsb,
)
```

After typing `om.algos.Bounded`, the autocomplete in any modern editor will only suggest algorithms that support bound constraints. 

Categories that can be used to filter algorithms are:
- GradientBased 
- GradientFree
- Global
- Local
- Bounded
- LinearConstrained
- NonlinearConstrained
- Scalar
- LeastSquares
- Likelihood
- Parallel 

Categories can be selected in any order. We make sure that mutually exclusive categories (e.g. Local and Global) cannot be selected at the same time. 

## Detailed description

The changes are described in [EP-02: Static Typing](https://optimagic.readthedocs.io/en/latest/development/ep-02-typing.html#step-3-filtered-autocomplete)

## Deviations from the EP

- The `Unbounded` and `Unconstrained` categories were removed because people rarely want to specifically choose an optimizer that does not support constraints or bounds. 
- `om.algorithms` was renamed to `om.algos` because it is still unambiguous and makes the code shorter, which hopefully improves adoption of using this instead of strings for algorithm selection.


.tools/create_algo_selection_code.py
```diff
@@ -0,0 +1,415 @@
...
+from optimagic.typing import AggregationLevel
+
+
+def main():
+    """Create the source code for algorithms.py.
+
+    The main part are nested dataclasses that enable filtered autocomplete for algorithm
+    selection.
+
+    """
```
timmens, line 20: I would like to see a summary of how the code generation works in the docstring. Like, with what do we start, what do we want, and what are the steps to get there.

.tools/create_algo_selection_code.py
```diff
@@ -0,0 +1,415 @@
...
+        f.write("\n\n")
+        f.write(instantiation_snippet)
+
+
+# ======================================================================================
+# Functions to collect algorithms
+# ======================================================================================
+
+
+def _import_optimizer_modules(package_name):
```
timmens, line 64: Type hints?

.tools/create_algo_selection_code.py
```diff
@@ -0,0 +1,415 @@
...
+        module_parts = module_name.split(".")
+        if all(not part.startswith("_") for part in module_parts) and not is_pkg:
+            module = importlib.import_module(module_name)
+            modules.append(module)
+
+    return modules
+
+
+def _get_all_algorithms(modules: list[ModuleType]) -> dict[str, Type[Algorithm]]:
+    """Collect all algorithms in moudles."""
```
timmens, line 81:
```suggestion
    """Collect all algorithms in modules."""
```

.tools/create_algo_selection_code.py
```diff
@@ -0,0 +1,415 @@
...
+
+    # write code to the file
+
+    with open(OPTIMAGIC_ROOT / "algorithms.py", "w") as f:
+        f.write(docstring)
+        f.write(imports + "\n\n")
+        f.write(parent_class_snippet + "\n")
+        f.write("\n\n".join(dataclass_snippets))
+        f.write("\n\n")
+        f.write(instantiation_snippet)
```
timmens, line 56: Any reason not to combine the code into a single string beforehand, and use pathlib.Path's write functionality?

```python
# Combine all the content into a single string
content = (
    docstring +
    imports + "\n\n" +
    parent_class_snippet + "\n" +
    "\n\n".join(dataclass_snippets) + "\n\n" +
    instantiation_snippet
)

# Write the combined content to the file
with (OPTIMAGIC_ROOT / "algorithms.py").open("w") as f:
    f.write(content)
```

.tools/create_and_format_algo_selection_code.sh
```diff
@@ -0,0 +1,17 @@
+#!/bin/bash
+set -e  # Exit immediately if a command exits with a non-zero status
+
+# # Check if the project is already installed locally
```
timmens, line 4:
```suggestion
# Check if the project is already installed locally
```

tests/optimagic/test_create_algo_selection.py
timmens, line 1: It seems that since pushing this file, you have implemented some of the functionality differently in [.tools/create_algo_selection_code.py](https://github.com/optimagic-dev/optimagic/pull/534/files?diff=unified&w=0#diff-bfdf94afae665f1a8a4daefd1f2fcc5e2c47c4041413130a409fb59a6e60371d). There should probably be only one way of doing it? Could this test be removed entirely?

docs/source/how_to/how_to_specify_algorithm_and_algo_options.md
timmens, line 1: This is not rendered properly:

image

timmens (changes requested):
Looks very good already, only a few things that need updates.

This is a great step towards much better auto complete 🎉

janosg:
Tests are failing due to the release of DFO-LS 1.5.1; This is unrelated to the current changes and would affect all currently open PRs (@timmens, @mpetrosian).

docs/source/how_to/how_to_specify_algorithm_and_algo_options.md
janosg: This was a pre-commit hook that destroyed the markdown formatting. I excluded that file now. Long run we can look for a better solution.

docs/source/how_to/how_to_specify_algorithm_and_algo_options.md
```diff
@@ -1,46 +1,153 @@
...
-implemented in optimagic, or a function that fulfills the interface laid out in
-{ref}`internal_optimizer_interface`.
+There are two ways to specify and configure optimizers. The *optimagic way* and the
+*scipy way*. Both use the `algorithm` argument of `minimize` and `maximize` to specify
+an optimizer and both are super easy to use.
 
-Which algorithms are available in optimagic depends on the packages a user has
-installed. We list all supported algorithms in {ref}`list_of_algorithms`.
+As the name suggests, the *scipy way* is more familiar for users of scipy.optimize. The
+*optimagic way* has much better discoverability and autocomplete. Using the *optimagic
```
timmens, line 15: This sounds slightly negative toward scipy imo. I think you could rephrase like:

```suggestion
*optimagic way* adds discoverability and autocomplete. Using the *optimagic
```

docs/source/how_to/how_to_specify_algorithm_and_algo_options.md
```diff
@@ -1,46 +1,153 @@
...
+for algo in algorithms:
+    minimize(
+        fun=fun,
+        params=np.arange(3),
+        algorithm=algo,
+    )
+```
+
+:::
+:::{tab-item} The scipy way
```
timmens, line 106: Do we accept the scipy method alias, and the scipy method names aliases already? In any case, I think this should be mentioned here. I.e., it is either slightly different or it is possible to go "full scipy." But please convince me from the opposite.

timmens (approved):
Looks very good! I only have two minor comments on the documentation.

docs/source/how_to/how_to_specify_algorithm_and_algo_options.md
janosg: I added a note.

PR #543, opened 2024-10-25
Support numpy>=2

timmens:
- Make optimagic compatible with numpy>=2
- Test optimagic with numpy<2 and numpy>=2:
  1. First test is pandas<2 and numpy<2
  2. Second test is pandas>= 2 and numpy<2
  3. Perform all other tests with numpy>=2
- Pin mypy (should be the same version as used in the pre-commit hook)


janosg (approved):
Thanks!

PR #544, opened 2024-10-28
Fix results processing of nag optimizers

### LINKED ISSUE #538: Result processing crashes if DFOLS starts from optimum
### Bug description

Hi! I noticed that if you start an OLS minimization problem from an initial point that is the optimum (so that no iterations happen), `minimize` crashes with (at least) DFOLS. The issue arises in the processing of results, when trying to figure out the number of iterations the problem ran for.

### To Reproduce

Here is a minimal example:

```
import optimagic as om
import numpy as np

# A simple loss
@om.mark.least_squares
def loss(x):
    return x

x0 = np.array([0.0, 0.0])
res =  om.minimize(
    fun=loss,
    params=x0,
    algorithm="nag_dfols",
    bounds=om.Bounds(np.ones_like(x0) * -1.0, np.ones_like(x0)),
)
```

### Expected behavior

Would have expected to still get a result object, even if no iterations occurred.

### Screenshots/Error messages

The error  is
```
  File "[...]\optimagic\optimizers\nag_optimizers.py", line 633, in nag_dfols_internal
    n_iterations=int(res["n_iterations"]),
                 ^^^^^^^^^^^^^^^^^^^^^^^^
TypeError: int() argument must be a string, a bytes-like object or a real number, not 'NoneType'
```

### System

 - OS: Windows
 - Version: optimagic 0.5.0, dfo-ls 1.3.0

Thanks as always!

timmens:
- Fixes results processing of nag optimizers
- [x] Wait for DFO-LS fix (opened issue [here](https://github.com/numericalalgorithmsgroup/dfols/issues/24))

Closes #538

### DIFF
--- a/.tools/envs/testenv-linux.yml
+++ b/.tools/envs/testenv-linux.yml
@@ -28,7 +28,7 @@ dependencies:
   - jinja2  # dev, tests
   - annotated-types  # dev, tests
   - pip:  # dev, tests, docs
-      - DFO-LS  # dev, tests
+      - DFO-LS>=1.5.3  # dev, tests
       - Py-BOBYQA  # dev, tests
       - fides==0.7.4  # dev, tests
       - kaleido  # dev, tests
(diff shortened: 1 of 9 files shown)


janosg (approved):
Thanks for fixing this!

PR #550, opened 2024-11-12
Polish the algo selection tool

janosg:
- [x] Create a better string representation for algorithm classes
- [x] Allow algorithm classes and instances as dict keys in the results dict for criterion_plot

### DIFF
--- a/.tools/create_algo_selection_code.py
+++ b/.tools/create_algo_selection_code.py
@@ -109,7 +109,7 @@ def _get_algorithms_in_module(module: ModuleType) -> dict[str, Type[Algorithm]]:
     }
     algos = {}
     for candidate in candidate_dict.values():
-        name = candidate.__algo_info__.name
+        name = candidate.algo_info.name
         if issubclass(candidate, Algorithm) and candidate is not Algorithm:
             algos[name] = candidate
     return algos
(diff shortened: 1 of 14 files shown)


tests/optimagic/visualization/test_history_plots.py
timmens, line 1: Can you add a test where names are (1) an algorithm class and (2) an algorithm instance? Or is there a reason why you did not do that?

timmens (approved):

.tools/create_algo_selection_code.py
```diff
@@ -119,47 +119,47 @@ def _get_algorithms_in_module(module: ModuleType) -> dict[str, Type[Algorithm]]:
 # Functions to filter algorithms by selectors
 # ======================================================================================
 def _is_gradient_based(algo: Type[Algorithm]) -> bool:
-    return algo.__algo_info__.needs_jac  # type: ignore
+    return algo.algo_info.needs_jac  # type: ignore
```
timmens, line 122: So mypy still does not understand this. Any ideas why?

janosg: I was surprised too. For mypy this is a method, even though it is a property.

Issue #547, opened 2024-11-05
Improve runtime measures for criterion plot and benchmarking plots


janosg:
## Current Situation / Problem you want to solve

The proposal in this issue concerns the functions `criterion_plot`, `profile_plot` and `convergence_plot`.

- The `criterion_plot` uses the number of function evaluations (`n_evaluations`) as runtime measure
- The `profile_plot` and `convergence_plot` has a `runtime_measure` argument that lets the user switch between `n_evaluations`, `n_batches`, and `walltime`.

Each runtime measure serves a purpose:

- **walltime**: Measures how long it actually takes to achieve a certain progress. This is what a user ultimately cares about in their optimization problem.
- **n_evaluations**: Measures how many evaluations of the objective function it takes to achieve a certain progress. This allows to ignore optimizer overhead and use fast benchmark functions to judge the performance of an optimizer that is designed for expensive objective functions. Moreover, it is deterministic and reproducible across machines. 
- **n_batches**: Similar to n_evaluations. In addition it allows to simulate the performance of a parallel optimizer on small machines. 

`n_evaluations` and `n_batches` measure important aspects but also have a big drawback: They exclusively focus on objective functions and ignore all time that is spent on evaluating derivatives. This is not a problem as long as only derivative free or only derivative based optimizers are compared. But as soon as one compares a derivative free with a derivative based optimizer it becomes misleading. 

### Step 1: Introduce a new runtime measures:

All relevant functions will get a `runtime_measure` argument which can be:

- `"function_time"` (default): The time spent in evaluations of the user provided functions `fun`, `jac`, `fun_and_jac`; Similar to `n_evaluations`, this will ignore the overhead of calculations done in the optimizer. 
- `"batch_function_time"`: The time that would have been spent in evaluations of user provided functions if all evaluations of the same batch were done in parallel (without parallelization overhead). 
- `"walltime"`: The actual time spent (reflecting actual optimizer overheads, parallelization overheads, ...)

We also keep the legacy measures `"n_evaluations"` and `"n_batches"`.

### Step 2: Introduce an optional **cost model**

While `"function_time"` and `"batch_function"` time allow to ignore optimizer overhead, they are not deterministic nor comparable across machines. In order to achieve this, we optionally allow a user to pass a `CostModel` as `runtime_measure`. Using a `CostModel` allows to reproduce all existing measures except for walltime. Moreover, it allows to get reproducible and hardware agnostic runtime measures for almost any situation.

[...]

PR #552, opened 2024-11-30
Refactor and extend History class

timmens:
- Remove `HistoryArrays` class and add these methods to `History` class
- Add `timing` module that contains functionality to create `CostModels` (as described #547)
- Add new methods and properties to `History` class, most importantly `fun_data` and `params_data`, which return tidy data frames of the function evaluations and a user-selected time measure.
- [x] Check visually that plots have not changed!

> [!NOTE]
> 1. For the history processing for visualizations (`history_plots.py`), we must wait for #553 
> 2. For the usage of the new methods `fun_data` and `params_data` in the visualization code we will also wait for #553


src/optimagic/optimization/history.py
```diff
@@ -50,38 +85,210 @@ def add_batch(
...
+        fun_arr = np.array(self.fun, dtype=np.float64)
+        if self.direction == Direction.MINIMIZE:
+            return fun_arr <= self.monotone_fun
+        elif self.direction == Direction.MAXIMIZE:
+            return fun_arr >= self.monotone_fun
+
+    # Parameter data, params, flat params, and flat params names
+    # ----------------------------------------------------------------------------------
+
+    def params_data(self, cost_model: CostModel) -> pd.DataFrame:
```
janosg, line 158: I thought about it again and think that `params_data` does not need a CostModel at all. It is never used to compare multiple optimizations but only to investigate how the parameters are developing in one optimization. Therefore, there is no need to compare speed.

I would therefore also replace the `"time"` column by and `"index"` or `"counter"` column. 

Moreover there should be a way to limit this to evaluations where the function value is not missing (i.e. not pure jac evaluations) and this should be the default. Integrating this into `params_data` is necessary because filtering afterwards makes the `counter` useless (e.g. the filtered counter might have gaps like 1, 2, 8). I am against a very general solution based on EvalTask and would just add a boolean argument like `dropna` or `drop_rows_without_fun`. 

This whole proposal breaks the symmetry between `fun_data` and `params_data` but maybe they are just more different than we thought. 

An alternative proposal would be to keep the cost model; Then the cost model can be used to create an appropriate counter and the filtering can be done afterwards. 

@timmens Maybe we need to discuss this.

src/optimagic/optimization/history.py
```diff
@@ -50,38 +85,272 @@ def add_batch(
...
+
+    # Batches and fast_path
+    # ----------------------------------------------------------------------------------
 
     @property
     def batches(self) -> list[int]:
         return self._batches
 
+    @property
+    def is_serial(self) -> bool:
```
janosg, line 330: I think this should be a private method, not a public property of the History.

src/optimagic/optimization/history.py
```diff
@@ -90,11 +359,157 @@ def criterion(self) -> list[float | None]:
...
+
+def _task_to_categorical(task: list[EvalTask]) -> pd.Categorical:
+    return pd.Categorical(
+        [t.value for t in task], categories=[t.value for t in EvalTask]
+    )
+
+
+def _apply_reduction_to_batches(
+    data: NDArray[np.float64],
+    batch_ids: list[int],
```
janosg, line 453: It's a bit weird that everything is an Array but `batch_ids` is a list.

timmens:
The new plotly version (6.0.0) seems to have a bug. I had to restrict it. I could not find a MRE quickly. However, I already found out that the problem stems from these lines:

```python
# plotting_utilities.py
    if share_yrange_all:
        lb = []
        ub = []
        for f in plots:
            for d in f.data:
>               lb.append(np.min(d["y"]))
>               ub.append(np.max(d["y"]))
```

The problem seems to be that `d["y"]` (and for that matter also `d["x"]` a few lines below) return a dictionary instead of a list of floats.

janosg (approved):

PR #553, opened 2024-11-30
Align plotting from logs and results

janosg:


hmgaudecker:
Only sort-of related, just occurred to me while looking at plots: Shouldn't we rename `criterion_plot` to `fun_plot` or create an alias?

PR #566, opened 2025-03-13
Add wrappers for pyensmallen

gauravmanmode:
added lbfgs barebone

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -16,6 +16,7 @@ dependencies = [
     "sqlalchemy>=1.3",
     "annotated-types",
     "typing-extensions",
+    "pyensmallen",
 ]
 dynamic = ["version"]
 keywords = [
@@ -334,6 +335,7 @@ ignore_errors = true
 
 [[tool.mypy.overrides]]
 module = [
+    "pyensmallen",
     "pybaum",
     "scipy",
     "scipy.linalg",
--- a/src/optimagic/optimizers/pyensmallen_optimizers.py
+++ b/src/optimagic/optimizers/pyensmallen_optimizers.py
@@ -1,0 +1,62 @@
+"""Implement ensmallen optimizers."""
+
+from dataclasses import dataclass
+
+import numpy as np
+import pyensmallen as pye
+from numpy.typing import NDArray
+
+from optimagic import mark
+from optimagic.optimization.algo_options import (
+    MAX_LINE_SEARCH_STEPS,
+    STOPPING_MAXITER,
+)
+from optimagic.optimization.algorithm import Algorithm, InternalOptimizeResult
+from optimagic.optimization.internal_optimization_problem import (
+    InternalOptimizationProblem,
+)
+from optimagic.typing import (
+    AggregationLevel,
+)
+
+
+@mark.minimizer(
+    name="ensmallen_lbfgs",
+    solver_type=AggregationLevel.SCALAR,
+    is_available=True,
+    is_global=False,
+    needs_jac=True,
+    needs_hess=False,
+    supports_parallelism=False,
+    supports_bounds=True,
+    supports_linear_constraints=False,
+    supports_nonlinear_constraints=False,
+    disable_history=False,
+)
+@dataclass(frozen=True)
+class EnsmallenLBFGS(Algorithm):
+    stopping_maxiter = STOPPING_MAXITER
+    max_step_for_line_search = MAX_LINE_SEARCH_STEPS
+    # min_step_for_line_search = MIN_LINE_SEARCH_STEPS
+
+    def _solve_internal_problem(
+        self, problem: InternalOptimizationProblem, x0: NDArray[np.float64]
+    ) -> InternalOptimizeResult:
+        optimizer = pye.L_BFGS()
+
+        print(type(problem.fun(x0)))
+
+        def objective_function(
+            x: NDArray[np.float64], grad: NDArray[np.float64]
+        ) -> np.float64:
+            grad[:] = problem.jac(x)
+            return np.float64(problem.fun(x))
+
+        raw_res = optimizer.optimize(objective_function, x0)
+
+        res = InternalOptimizeResult(
+            x=raw_res,
+            fun=problem.fun(raw_res),
+        )
+
+        return res
--- a/src/optimagic/algorithms.py
+++ b/src/optimagic/algorithms.py
@@ -36,6 +36,7 @@ from optimagic.optimizers.nlopt_optimizers import (
     NloptVAR,
 )
 from optimagic.optimizers.pounders import Pounders
+from optimagic.optimizers.pyensmallen_optimizers import EnsmallenLBFGS
 from optimagic.optimizers.pygmo_optimizers import (
     PygmoBeeColony,
     PygmoCmaes,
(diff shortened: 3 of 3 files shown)


environment.yml
```diff
@@ -40,6 +40,7 @@ dependencies:
       - DFO-LS>=1.5.3  # dev, tests
       - Py-BOBYQA  # dev, tests
       - fides==0.7.4  # dev, tests
+      - pyensmallen
```
janosg, line 43: Please add the comment `# dev, tests` so we know in which environments this is needed.

pyproject.toml
```diff
@@ -16,6 +16,7 @@ dependencies = [
     "sqlalchemy>=1.3",
     "annotated-types",
     "typing-extensions",
+    "pyensmallen",
```
janosg, line 19: `pyensmallen` is going to be an optional dependency, so we don't need to list it in pyproject.toml

src/optimagic/optimizers/pyensmallen_optimizers.py
```diff
@@ -0,0 +1,88 @@
...
+            maxLineSearchTrials=self.max_line_search_trials,
+            minStep=self.min_step_for_line_search,
+            maxStep=self.max_step_for_line_search,
+        )
+
+        def objective_function(
+            x: NDArray[np.float64], grad: NDArray[np.float64]
+        ) -> np.float64:
+            grad[:] = problem.jac(x)
+            return np.float64(problem.fun(x))
```
janosg, line 79: Since both fun and jac are needed, it would be better to use `problem.fun_and_jac`. Then the `InternlOptimizationProblem` can decide which functions to call internally to get there. This is more efficient if a `fun_and_jac` was provided by the user or if we need to calculate numerical derivatives.

src/optimagic/optimizers/pyensmallen_optimizers.py
```diff
@@ -0,0 +1,88 @@
...
+
+        raw = optimizer.optimize(objective_function, x0)
+
+        res = InternalOptimizeResult(
+            x=raw,  # only best x is available
+            fun=problem.fun(raw),  # best f(x) value is not available
```
janosg, line 85: Is there a chance to get more infos (like an iteration counter, etc.) out of the pyensmallen results object? It would also be good to get the optimal function value out of pyensmallen instead of doing an extra function evaluation on top here.

janosg (commented):
Hi @gauravmanmode, thanks for your PR. I made a few comments but this is definitely going into the right direction.

janosg:
Hi @gauravmanmode, @timmens  and I quickly looked into the failures on GitHub actions and found the following:

- Since pyensmallen will be an optional dependency, you have to put all imports into if statements. See [fides](https://github.com/optimagic-dev/optimagic/blob/ba2678753587f91cea54de69ff76cb3dcb4257d4/src/optimagic/optimizers/fides.py#L32) as an example. 
- Most of the failures are expected due to known limitations of the pyensmallen wheels (see e.g. [here](https://github.com/apoorvalal/pyensmallen/issues/8). Here the only thing we can do for now is skip the tests on non-linux systems and for Python >= 3.13; For all tests you are writing you can use pytests `skipif` decorator to enable this. 
- On some systems the failure is already at the environment creation. @timmens  will post a workaround for this.

src/optimagic/optimizers/pyensmallen_optimizers.py
gauravmanmode: currently,  the python wrapper for ensmallen is very minimalistic. the result object is just the best x value. it is possible to get a more detailed report by passing a Report() callback function but it is not implemented in the wrapper yet.

gauravmanmode:
I have done the above said changes.

This is the alignment of tuning parameters for:
### ensmallen_lbfgs
| **Old Name**            | **Proposed Name**                | **from optimizer in optimagic**       |
|------------------------|--------------------------------|----------------|
| `numBasis`            | `limited_memory_max_history`  |  `ipopt`       |
| `maxIterations`       | `stopping_maxiter`            |   `scipy`               |
| `armijoConstant`      | `armijo_constant`             | needs review   |
| `wolfe`               | `wolfe_condition`             | needs review   |
| `minGradientNorm`     | `convergence_gtol_abs`        |    `scipy`              |
| `factr`               | `convergence_ftol_rel`        |     `scipy`              |
| `maxLineSearchTrials` | `max_line_search_trials`      |  needs review               |
| `minStep`             | `min_step_for_line_search`    |   needs review              |
| `maxStep`             | `max_step_for_line_search`    |  `scipy`    |

src/optimagic/optimizers/pyensmallen_optimizers.py
janosg: Ok. This is something we'll need to address before we can merge.

janosg:
> I have done the above said changes.
> 
> This is the alignment of tuning parameters for:
> 
> ### ensmallen_lbfgs
> **Old Name**	**Proposed Name**	**from optimizer in optimagic**
> `numBasis`	`limited_memory_storage_length`	`scipy`
> `maxIterations`	`stopping_maxiter`	`scipy`
> `armijoConstant`	`armijo_constant`	needs review
> `wolfe`	`wolfe_condition`	needs review
> `minGradientNorm`	`convergence_gtol_abs`	`scipy`
> `factr`	`convergence_ftol_rel`	`scipy`
> `maxLineSearchTrials`	`max_line_search_trials`	needs review
> `minStep`	`min_step_for_line_search`	needs review
> `maxStep`	`max_step_for_line_search`	`scipy`
> I have attached a jupyter notebook which i used for testing and exploring the library. [link](https://github.com/gauravmanmode/sharing/blob/main/test_optimagic.ipynb)

Hi @gauravmanmode, thanks for the table. The names are really good. The only thing I would change is `max_line_search_trials` -> `max_line_search_steps` because wa already use that name in `scipy_lbfgsb`. (Unless I'm misunderstanding something and the options are conceptually different).

src/optimagic/optimizers/pyensmallen_optimizers.py
```diff
@@ -0,0 +1,100 @@
...
+"""Parameter for detecting the Wolfe condition."""
+
+STEP_SIZE = 0.001
+"""Step size for each iteration."""
+BATCH_SIZE = 32
+"""Step size for each iteration."""
+EXP_DECAY_RATE_FOR_FIRST_MOMENT = 0.9
+"""Exponential decay rate for the first moment estimates."""
+EXP_DECAY_RATE_FOR_WEIGHTED_INF_NORM = 0.999
+"""Exponential decay rate for the first moment estimates."""
```
janosg, line 43: These four constants are not used below. Are they also tuning parameters of esmallen_lbfgs?

gauravmanmode: sry for that. i was testing something else. it is not relevant to esmallen_lbfgs. i will remove them.

gauravmanmode:
hi,
both `max_line_search_steps` and `max_line_search_trials` are the same.
also, i should be able to get a more detailed OptimizationReport from pyensmallen in a few days.

janosg:
Is there any update on the more detailed OptimizationReport?

gauravmanmode:
I have created a PR [here](https://github.com/apoorvalal/pyensmallen/pull/14) on pyensmallen repo 2 weeks ago but the maintainer is inactive so this has been pending.

PR #567, opened 2025-03-14
make fun optional when fun_and_jac is provided

### LINKED ISSUE #518: Make `fun` optional if `fun_and_jac` is provided
Currently `fun` is a mandatory argument even if `fun_and_jac` is provided. We can make it optional. It is not hard but not a breaking change so I postpone until after the next release. 
 
We should not forget to adjust the how-to-guide on custom derivatives after doing that.

gauravmanmode:
Closes #518 
### Proposed Solution
if fun_and_jac is not None, and fun is not provided, set fun from fun_and_jac.
Also adjusted the docs.
I have added a test also.

### DIFF
--- a/src/optimagic/optimization/create_optimization_problem.py
+++ b/src/optimagic/optimization/create_optimization_problem.py
@@ -140,10 +140,11 @@ def create_optimization_problem(
     # error handling needed as long as fun is an optional argument
     # ==================================================================================
 
-    if fun is None and criterion is None:
+    if fun_and_jac is None and fun is None and criterion is None:
         msg = (
             "Missing objective function. Please provide an objective function as the "
-            "first positional argument or as the keyword argument `fun`."
+            "first positional argument or as the keyword argument `fun` or "
+            " with `fun_and_jac`."
         )
         raise MissingInputError(msg)
 
@@ -160,6 +161,10 @@ def create_optimization_problem(
             "argument or as the keyword argument `algorithm`."
         )
         raise MissingInputError(msg)
+
+    if fun_and_jac is not None:
+        if fun is None:
+            fun = split_fun_and_jac(fun_and_jac, target="fun")
 
     # ==================================================================================
     # deprecations
--- a/tests/optimagic/optimization/test_optimize.py
+++ b/tests/optimagic/optimization/test_optimize.py
@@ -44,3 +44,18 @@ def test_with_invalid_numdiff_options():
             algorithm="scipy_lbfgsb",
             numdiff_options={"bla": 15},
         )
+
+
+# provided fun or fun_and_jac is provided
+def test_with_optional_fun_argument():
+    expected = minimize(
+        fun=lambda x: x @ x,
+        params=np.arange(5),
+        algorithm="scipy_lbfgsb",
+    )
+    res = minimize(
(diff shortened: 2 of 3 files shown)


janosg:
Hi @gauravmanmode, thanks for the PR. 

We want to support this, but there is one edge case that makes it slightly harder: As you can see [here](https://optimagic.readthedocs.io/en/latest/how_to/how_to_derivatives.html#derivatives-that-work-for-scalar-and-least-squares-optimizers), `jac` or `fun_and_jac` can be a list of callables if users want to switch seamlessly between scalar and least_squares optimizers. In that case we would have to select the correct entry from the list before we can call `split_fun_and_jac`. But at the point where you are handling this case we don't even know yet which entry we would need. 

My plan was to wait until version 0.6 where we can remove a lot of deprecated code that currently limits the order in which we can process user inputs. 

But we can keep your PR as a step in the right direction if you raise an error for the (very rare) case that `fun_and_jac` is a list.

gauravmanmode:
you are right. i dont know how to handle this. since you said it is a rare case, i have raised a error and added a test.

src/optimagic/optimization/create_optimization_problem.py
```diff
@@ -161,6 +162,14 @@ def create_optimization_problem(
         )
         raise MissingInputError(msg)
 
+    if fun_and_jac is not None and fun is None:
+        if isinstance(fun_and_jac, list):
+            raise NotImplementedError(
+                "`fun_and_jac` cannot be a list, this is not yet supported."
```
janosg, line 168:
```suggestion
                "If `fun_and_jac` is a list of callables, `fun` is not optional. "
```

tests/optimagic/optimization/test_optimize.py
```diff
@@ -44,3 +44,28 @@ def test_with_invalid_numdiff_options():
...
+        jac=lambda x: 2 * x,
+        params=np.arange(5),
+        algorithm="scipy_lbfgsb",
+    )
+    res = minimize(
+        fun_and_jac=lambda x: (x @ x, 2 * x),
+        params=np.arange(5),
+        algorithm="scipy_lbfgsb",
+    )
+    assert np.all(res.x == expected.x) == np.True_
```
janosg, line 62: Since we know the result of this test I would not run two optimizations (keeping test runtime short is very important). Instead I would prefer something like `expected = np.zeros(5)` and use `assert_array_almost_equal` for the comparison.

src/optimagic/optimization/create_optimization_problem.py
```diff
@@ -161,6 +162,14 @@ def create_optimization_problem(
         )
         raise MissingInputError(msg)
 
+    if fun_and_jac is not None and fun is None:
```
janosg, line 165:
```suggestion
    if fun_and_jac is not None and fun is None and criterion is None:
```

We need the check for `criterion` until the deprecation of `criterion` is complete.

janosg (commented):
Thanks @gauravmanmode, this is looking really good now. I only had a few small suggestions that should be very quick to address.

gauravmanmode:
@janosg thanks for the suggestions. i hope this resolves them.

janosg (approved):
Thanks a lot; This is good to merge now!

PR #568, opened 2025-03-14
added initial wrapper for migrad

spline2hg:
The basic functionality has been tested with the following Python code:

```python
import numpy as np
import optimagic as om

def cost_function_with_two_minima(x):
    return x**4 - x**2 + 1


res = om.minimize(
    fun=cost_function_with_two_minima,
    params=0.1,
    algorithm="iminuit_migrad",
    bounds=om.Bounds(
        lower=0,
        upper=0.6
    )
)

print(res)
```


src/optimagic/optimizers/iminuit_migrad.py
```diff
@@ -33,19 +32,30 @@
...
         def wrapped_objective(x: NDArray[np.float64]) -> float:
             return float(problem.fun(x))
 
-        m = Minuit(
-            wrapped_objective,
-            params,
-        )
+        wrapped_gradient = None
+        jac_func = None
+        if problem.jac is not None:
```
janosg, line 44: You don't need this check. `problem.jac` is never None. If the user did not pass a derivative, it is a numerical one.

src/optimagic/optimizers/iminuit_migrad.py
```diff
@@ -33,19 +32,30 @@
...
+
+            def wrapped_gradient(x: NDArray[np.float64]) -> NDArray[np.float64]:
+                return problem.jac(x)
+
+            jac_func = problem.jac
+        elif problem.fun_and_jac is not None:
+
+            def wrapped_gradient(x: NDArray[np.float64]) -> NDArray[np.float64]:
+                _, jac = problem.fun_and_jac(x)
+                return jac
```
janosg, line 54: Again, `fun_and_jac` is never None for the `InternalOptimizationProblem`. The logic you have here (including much more actually) is already in the `InternalOptimizationProblem` so we don't have to repeat it per optimizer.

src/optimagic/optimizers/iminuit_migrad.py
```diff
@@ -33,19 +32,30 @@
...
+
+            jac_func = lambda x: problem.fun_and_jac(x)[1]
+
+        m = Minuit(wrapped_objective, params, grad=wrapped_gradient)
```
janosg, line 58: I think you can change this to `grad=problem.jac` and then it is the only line you need for passing the gradient to iminuit. Everything else is already done for you in the `InternalOptimizationProblem`.

src/optimagic/optimizers/iminuit_migrad.py
```diff
@@ -0,0 +1,102 @@
...
+from optimagic.optimization.internal_optimization_problem import (
+    InternalOptimizationProblem,
+)
+from optimagic.typing import AggregationLevel
+
+
+@mark.minimizer(
+    name="iminuit_migrad",
+    solver_type=AggregationLevel.SCALAR,
+    is_available=True,
```
janosg, line 22: Only if `iminuit` is installed. See [here](https://optimagic.readthedocs.io/en/latest/how_to/how_to_add_optimizers.html#write-the-minimal-implementation) for how to handle this.

src/optimagic/optimizers/iminuit_migrad.py
```diff
@@ -0,0 +1,102 @@
...
+
+        m = Minuit(wrapped_objective, params, grad=wrapped_gradient)
+
+        if problem.bounds:
+            lower_bounds = problem.bounds.lower
+            upper_bounds = problem.bounds.upper
+
+            if lower_bounds is not None and upper_bounds is not None:
+                for i, (lower, upper) in enumerate(
+                    zip(lower_bounds, upper_bounds, strict=False)
```
janosg, line 66: We would want strict=True here!

src/optimagic/optimizers/iminuit_migrad.py
```diff
@@ -0,0 +1,102 @@
...
+                ):
+                    if lower is not None or upper is not None:
+                        m.limits[i] = (lower, upper)
```
janosg, line 69: Please extrac this into a private function (i.e. a function name that starts with an `_`) and write a test for the bounds conversion. Please also describe in the docstring of that function in which format `iminuit` expects the bounds. 

I'm not sure if the logic is correct. Both `bounds.lower` and `bounds.upper` are numpy arrays, so the elements in it would never be `None` but only `NaN` or `-np.inf` or `np.inf` if there are not bounds. But let's first do the refactoring and then have another look.

src/optimagic/optimizers/iminuit_migrad.py
```diff
@@ -0,0 +1,102 @@
...
+
+        m.migrad(ncall=self.stopping_maxfun)
+
+        res = process_minuit_result(m, jac_func)
+        return res
+
+
+def process_minuit_result(
```
janosg, line 77: Please start the function name with an `_` as this can be a private function.

src/optimagic/optimizers/iminuit_migrad.py
```diff
@@ -0,0 +1,102 @@
...
+    minuit_result: Minuit,
+    jac_fun: Optional[Callable[[NDArray[np.float64]], NDArray[np.float64]]] = None,
+) -> InternalOptimizeResult:
+    jac = None
+    if jac_fun is not None:
+        jac = jac_fun(np.array(minuit_result.values))
```
janosg, line 83: Ideally we don't want to do another gradient evaluation (which could be very costly) here but get one back from the optimizer (if possible) or set it to None.

src/optimagic/optimizers/iminuit_migrad.py
```diff
@@ -0,0 +1,102 @@
...
+
+    res = InternalOptimizeResult(
+        x=np.array(minuit_result.values),
+        fun=minuit_result.fval,
+        success=minuit_result.valid,
+        message=repr(minuit_result.fmin),
+        n_fun_evals=minuit_result.nfcn,
+        n_jac_evals=minuit_result.ngrad,
+        n_hess_evals=None,
+        n_iterations=None,
```
janosg, line 93: We can also set this to `minuit_result.nfcn`

src/optimagic/optimizers/iminuit_migrad.py
```diff
@@ -0,0 +1,102 @@
...
+        status=None,
+        jac=jac,
+        hess=np.array(minuit_result.hesse()),
+        hess_inv=np.array(minuit_result.covariance),
+        max_constraint_violation=None,
+        info={"minos": minuit_result.minos()},
```
janosg, line 99: We don't want to do a potentially costly function call here. Only return things that are already automatically calculated.

janosg (commented):
Thanks for the PR! I made a few comments in the code. 

In addition:
- We need more tests, for example for the bounds conversion
- We need documentation

It looks surprising that the only tuning parameter of the algorithm is `stopping_maxfun`. Are there no other convergence criteria or tuning parameters?

src/optimagic/optimizers/iminuit_migrad.py
spline2hg: Adjusted it. My thought was that if we have strict=True then we wouldn't be able to set just the upper or lower bound individually, but om.Bounds is handling it.

src/optimagic/optimizers/iminuit_migrad.py
spline2hg: should i rename `process_minuit_result function` to `_process_minuit_result`? cause in scipy_optimizers we have not made `process_scipy_result function` private.

src/optimagic/optimizers/iminuit_migrad.py
spline2hg:
```python
optimized_params = np.array(minuit_result.values)
jacobian = minuit_result.grad(optimized_params)  
```  

I found this way to get the Jacobian, but it still performs another gradient evaluation. Since Minuit doesn't provide a direct way to retrieve the Jacobian, I'm setting it to `None`, is this fine?

src/optimagic/optimizers/iminuit_migrad.py
spline2hg: nfcn is the number of function calls on objective function , whereas n_iterations seems to be number of optimizer iteration,are this two the same?

src/optimagic/optimizers/iminuit_migrad.py
spline2hg:
```python
@dataclass(frozen=True)
class InternalBounds(Bounds):
    lower: NDArray[np.float64] | None
    upper: NDArray[np.float64] | None
    soft_lower: None = None
    soft_upper: None = None
```
Currently, bounds.lower and bounds.upper can be None as per the type hint. Should we handle None explicitly, or do we enforce that they must always be NumPy arrays? I’ve used `Optional[NDArray[np.float64]]` is that fine?

spline2hg:
hi @janosg,  

thanks for the review! i will go through all the suggestions and make the necessary changes.  

regarding **`stopping_maxfun`** being the only parameter, i had the same doubt. i found the following things  

1. **uses edm as convergence criterion:**  
   - `iminuit` does not allow setting tolerances for changes in the  parameters. it relies on the edm criterion .  
   - [is it possible to stop `iminuit` by setting a tolerance for changes in the minimized function or parameters?][edm-tolerance]  

2. **`iminuit.minuit` and `migrad` parameters:**  
   - the only parameters that seem compatible with `algo_options` in `iminuit.minuit` and `migrad` are `ncall` (max function calls) and `iterate` (migrad iterations).  
   - i hadn’t set the `iterate` parameter because `STOPPING_MAXITER` has a value of `1,000,000`, and the docs suggest starting with `iter >= 5`. i thought restarting optimization `STOPPING_MAXITER` times if it doesn’t converge wouldn’t be right, but i have set it now. let me know if you have any thoughts.  
   - [`iminuit.minuit` documentation][minuit-docs]  
   - [`migrad` documentation][migrad-docs]  

3. **internal interface:**  
   - the actual low-level process within `iminuit`, which includes parameters like tolerance and precision, is within the `_robust_low_level_fit` function. these parameters are not directly exposed to users via the `iminuit.minuit` interface.  
   - [`iminuit` github source][iminuit-src]  


i have tried to verify these points, but i might have missed something. happy to make further changes if needed!  

[edm-tolerance]: https://scikit-hep.org/iminuit/faq.html#is-it-possible-to-stop-iminuit-by-setting-a-tolerance-for-changes-in-the-minimized-function-or-the-parameters  
[minuit-docs]: https://scikit-hep.org/iminuit/reference.html#iminuit.Minuit  
[migrad-docs]: https://scikit-hep.org/iminuit/reference.html#iminuit.Minuit.migrad  
[iminuit-src]: https://github.com/scikit-hep/iminuit/blob/f8e72b5146642dade56613f4e779379986354928/src/iminuit/minuit.py#L789

src/optimagic/optimizers/iminuit_migrad.py
janosg: Yes, we need to handle the case that entire bounds arrays are None, but it won't happen that individual elements inside those arrays are None. They can only be `-np.inf`, `np.inf` or `np.NaN` because numpy arrays with dtype float cannot contain actual `None` entries. But in line 68 you check if individual array entries are None, if I understand correctly.

src/optimagic/optimizers/iminuit_migrad.py
janosg: Yes. scipy is the exception. (The scipy result object is used by multiple libraries so we use that function in multiple places)

src/optimagic/optimizers/iminuit_migrad.py
janosg: Yes, please set to None then.

src/optimagic/optimizers/iminuit_migrad.py
janosg: Not always but for most local gradient based methods they are.

janosg:
Hi @spline2hg,

Thanks for the detailed reply.

Ok, I am convinced that you did not miss any tuning parameters!

Note that `iter` is very different from what is usually set via `stopping_maxiter`. It is not counting actual optimizer iterations, which are usually in the thousands but actually running the optimizer multiple times. We want to support this option but a better name would be `n_restarts` and the default value should be 1. 

In the very long run we might look into the low-level interface but for now it is ok to just support two options. 

Of course, it's important to document them well and also to document why it is not possible to set additional convergence criteria.

src/optimagic/optimizers/iminuit_migrad.py
```diff
@@ -0,0 +1,154 @@
...
+        For each tuple:
+        - (None, None): Fully unbounded parameter
+        - (value, None): Lower bound only
+        - (None, value): Upper bound only
+        - (min, max): Two-sided constraint
+
+    """
+    for i, (lower, upper) in enumerate(bounds):
+        if lower is not None or upper is not None:
+            m.limits[i] = (lower, upper)
```
janosg, line 154: I would do this directly in `_solve_internal_problem` to avoid having an impure function here. By impure I mean that the function changes the `Minuit` object in place and returns nothing.

src/optimagic/optimizers/iminuit_migrad.py
```diff
@@ -0,0 +1,154 @@
...
+        hess=None,
+        hess_inv=np.array(minuit_result.covariance),
+        max_constraint_violation=None,
+        info=None,
+        history=None,
+    )
+    return res
+
+
+def _convert_bounds_to_minuit_limits(
```
janosg, line 86: This function needs a few unit tests. Please add them in a new file: test/optimagic/optimizers/test_iminuit_migrad.py`

janosg: Sounds good. I'll have a look once you commit.

spline2hg:
@janosg, do I need to add iminuit to the rtd_environment.yml file for the documentation build to pass? If not, could you help me identify the error?

janosg:
It was a temporary readthedocs problem.

janosg (approved):
Thanks a lot!

Issue #565, opened 2025-03-12
Add a batch evaluator that uses threading


janosg:
As seen from #493, multiprocessing is not always the right choice for parallelizing batch evaluators. We therefore need a batch evaluator that uses threading. See #493 for details.

Adding a batch evaluator is quite simple. Just write a function with the same interface as the existing ones [here](https://github.com/optimagic-dev/optimagic/blob/main/src/optimagic/batch_evaluators.py) in `batch_evaluators.py` and adjust `process_batch_evaluator` so it knows about the new batch evaluator. 

As a background, you can watch [this video](https://github.com/optimagic-dev/optimagic/blob/main/src/optimagic/batch_evaluators.py)

As a bonus you can think about other batch evaluators we are missing. Python parallelization has come a long way since we wrote our batch evaluators.

janosg:
Closed via #569

PR #569, opened 2025-03-14
add thread batch evaluator

spline2hg:
added batch evaluator that uses threading #565


spline2hg:
I have done a simple comparison between the three evaluators in this[notebook](https://github.com/spline2hg/optimagic_tutorial/blob/main/bench_batch_evaluators.ipynb) I noticed that the thread evaluator runs comparatively faster than the process evaluators. Could this be due to lower computational demands? Additionally, is the faster re-running due to some form of caching?

janosg:
Hi @spline2hg which one runs faster depends on the application. A thread based batch evaluator will be faster for IO bounds functions or very fast functions due to its lower overhead. Process based batch evaluators will be good for CPU bound long running functions. But that's the whole reason why we offer multiple batch evaluators, so everyone can pick the best for their problem 🙂

janosg:
Please also add tests for your new batch evaluator. Since the tests are already parametrized this should be as simple as adding "threading" to the list [here](https://github.com/optimagic-dev/optimagic/blob/ba2678753587f91cea54de69ff76cb3dcb4257d4/tests/optimagic/test_batch_evaluators.py#L8).

spline2hg:
just a small note the pathos batch evaluator is not added in the testing batch evaluator [here](https://github.com/optimagic-dev/optimagic/blob/ba2678753587f91cea54de69ff76cb3dcb4257d4/tests/optimagic/test_batch_evaluators.py#L8), we could keep it this way if thats what we wanted, i have made the said changes, let me know if any more changes are required

spline2hg:
hi @janosg , sorry, i might have overlooked something. i have fixed it, and it should most likely pass the ci run.

janosg (approved):
Thanks a lot!

PR #570, opened 2025-03-16
Add how-to add an algorithm guide and improve documentation of some internal classes

janosg:


mpetrosian (commented):
In `how_to_add_optimizers.ipynb`:
- `import pygmo as pg` results in ModuleNotFoundError.
- A few typos:
    - First sentence of # How to tuning parameters: We alread saw -> We already saw
    - # 2. Understand how the algorithm works: Here we wand -> Here we want
    - missing question marks at the end of the first two bullet points
    - # How does it work (intuitively): in the last sentence of the first paragraph, add generations in brackets after "..so that in later simulations" because in the next paragraph you talk about generations and it was introduced before.
    - # Which tuning parameters does it have?: 
        - missing periods at the end of the 1st, 7th, and last bullet points 
        - start with lower case after column in the second bullet point

mpetrosian (approved):
Many thanks, Janos. 
A very clear and concise how-to guide!

PR #572, opened 2025-03-23
Let `_fail_if_duplicates` raise the correct error.

hmgaudecker:
A skillmodels-based project came up with `selected = [1.0, 1.0, 44.0]`, so `_fail_if_duplicates` failed with

```
*** TypeError: list indices must be integers or slices, not float
```

instead of providing information on the problematic constraint.

Fix is to do the conversion to int first, rather than in the line just after the call to `_fail_if_duplicates`.


janosg (approved):
Thanks. This is much better!

PR #573, opened 2025-03-24
Convert to array in order to allow for Boolean indexing.

hmgaudecker:
Another triviality. Since `parnames` is a list and `is_invalid` Boolean array, no useful message comes out without the change.


timmens:
Thanks!!

PR #578, opened 2025-04-05
MAINT: using intersphinx-registry

Schefflera-Arboricola:
Found the broken numpy doc link in `intersphinx_mapping`, so used [intersphinx-registry](https://pypi.org/project/intersphinx-registry/) to avoid needing to update these links in the future. Also added `scipy` in the list of packages. Please LMK if this is a good change and/or anything needs to be changed.

Thank you :)


timmens (approved):
Thanks for the PR!

This is a nice change. If you like these kind of tasks, then you could also look into adding a sphinx-build job to our GitHub Actions that checks the links using the linkcheck option.

PR #579, opened 2025-04-07
Add wrapper for Nevergrad PSO optimizer

### LINKED ISSUE #560: Add wrappers for Nevergrad
Wrap the gradient free optimizers from [nevergrad](https://github.com/facebookresearch/nevergrad) in optimagic.

Nevergrad implements the following algorithms:

- `NgIohTuned` is "meta"-optimizer which adapts to the provided settings (budget, number of workers, parametrization) and should therefore be a good default.
- `TwoPointsDE` is excellent in many cases, including very high `num_workers`.
- `PortfolioDiscreteOnePlusOne` is excellent in discrete settings of mixed settings when high precision on parameters is not relevant; it's possibly a good choice for hyperparameter choice.
- `OnePlusOne` is a simple robust method for continuous parameters with `num_workers` < 8.
- `CMA` is excellent for control (e.g. neurocontrol) when the environment is not very noisy (num_workers ~50 ok) and when the budget is large (e.g. 1000 x the dimension).
- `TBPSA` is excellent for problems corrupted by noise, in particular overparameterized (neural) ones; very high `num_workers` ok).
- `PSO` is excellent in terms of robustness, high `num_workers` ok.
- `ScrHammersleySearchPlusMiddlePoint` is excellent for super parallel cases (fully one-shot, i.e. `num_workers` = budget included) or for very multimodal cases (such as some of our MLDA problems); don't use softmax with this optimizer.
- `RandomSearch` is the classical random search baseline; don't use softmax with this optimizer.

In the long run we want to wrap all of them, **but if you are tackling this as your first issue you should focus on one. In that case please coment below which optimizer you are going to work on so we don't duplicate efforts.**

r3kste:
Implements the Particle Swarm Optimization algorithm from `nevergrad`.
Partially fixes #560

### DIFF
--- a/src/optimagic/optimizers/nevergrad_optimizers.py
+++ b/src/optimagic/optimizers/nevergrad_optimizers.py
@@ -39,6 +39,7 @@ class NevergradPSO(Algorithm):
     transform: Literal["arctan", "gaussian", "identity"] = "arctan"
     population_size: int | None = None
     n_cores: int = 1
+    seed: int | None = None
     stopping_maxfun: PositiveInt = STOPPING_MAXFUN_GLOBAL
     inertia: float = 0.5 / math.log(2.0)
     cognitive: float = 0.5 + math.log(2.0)
@@ -63,6 +64,9 @@ class NevergradPSO(Algorithm):
                 shape=x0.shape, lower=problem.bounds.lower, upper=problem.bounds.upper
             )
         )
+
+        if self.seed is not None:
+            instrum.random_state.seed(self.seed)
 
         optimizer = ng.optimizers.ConfPSO(
             transform=self.transform,
--- a/src/optimagic/optimizers/nevergrad_optimizers.py
+++ b/src/optimagic/optimizers/nevergrad_optimizers.py
@@ -1,0 +1,91 @@
+"""Implement nevergrad optimizers."""
+
+import math
+from dataclasses import dataclass
+from typing import Literal
+
+import numpy as np
+from numpy.typing import NDArray
+
+from optimagic import mark
+from optimagic.config import IS_NEVERGRAD_INSTALLED
+from optimagic.exceptions import NotInstalledError
+from optimagic.optimization.algo_options import STOPPING_MAXFUN_GLOBAL
+from optimagic.optimization.algorithm import Algorithm, InternalOptimizeResult
+from optimagic.optimization.internal_optimization_problem import (
+    InternalOptimizationProblem,
+)
+from optimagic.typing import AggregationLevel, PositiveInt
+
+if IS_NEVERGRAD_INSTALLED:
+    import nevergrad as ng
+
+
+@mark.minimizer(
+    name="nevergrad_pso",
+    solver_type=AggregationLevel.SCALAR,
+    is_available=IS_NEVERGRAD_INSTALLED,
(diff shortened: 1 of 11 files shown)


r3kste:
Hello @janosg Is there anything left to implement in this PR? I would appreciate it if the CI gets approval to begin.

gulshan-123:
@r3kste , I am also working on similar PR: about the addition of OnePlusOne algorithm from nevergrad. I am curious if we can also add non-linear constraint in wrapper as suggested here: https://facebookresearch.github.io/nevergrad/optimization.html#optimization-with-constraints, can you please look into this?

src/optimagic/optimizers/nevergrad_optimizers.py
```diff
@@ -0,0 +1,104 @@
...
+            qo=self.quasi_opp_init,
+            sqo=self.speed_quasi_opp_init,
+            so=self.special_speed_quasi_opp_init,
+        )(
+            parametrization=instrum,
+            budget=self.stopping_maxfun,
+            num_workers=self.n_cores,
+        )
+
+        while optimizer.num_ask < optimizer.budget:
```
janosg, line 82: Minor thing, but I think it would be more readable to see `while optimizer.num_ask <= self.stopping_maxfun`

src/optimagic/optimizers/nevergrad_optimizers.py
```diff
@@ -0,0 +1,104 @@
...
+            for x, loss in zip(x_list, losses, strict=True):
+                optimizer.tell(x, loss)
+
+        recommendation = optimizer.provide_recommendation()
+
+        result = InternalOptimizeResult(
+            x=recommendation.value[0][0],
+            fun=recommendation.loss,
+            success=True,
+            n_fun_evals=optimizer.num_ask,
```
janosg, line 101: Can you try to define a few more fields of the `InternalOptimizeResult`, even if they are trivial (e.g. `n_jac_evals` could be 0 instead of None).

src/optimagic/optimizers/nevergrad_optimizers.py
```diff
@@ -0,0 +1,104 @@
...
+            parametrization=instrum,
+            budget=self.stopping_maxfun,
+            num_workers=self.n_cores,
+        )
+
+        while optimizer.num_ask < optimizer.budget:
+            x_list = [
+                optimizer.ask()
+                for _ in range(
+                    min(optimizer.num_workers, optimizer.budget - optimizer.num_ask)
```
janosg, line 86: Similarly, I would prefer `self.n_cores` over `optimizer.num_workers` here. For readers who are familiar with optimagic but not nevergrad (which is going to be the majority) it is easier to understand the optimagic terminology than the nevergrad terminology.

janosg (commented):
Hi @r3kste, thanks for the PR. The changes look really good. I just made a few minor stylistic comments. 

Currently the tests are not passing but I think this is not related to a problem in your code. It rather seems like nevergrad has a few things that are not yet compatible with numpy 2.0. 

Can you pin numpy to <2.0 in the environment to confirm this? If it works we can see how to deal with the situation.

janosg (approved):
Thanks a lot for the great PR!

PR #581, opened 2025-04-13
3d Visualizations of optimization problems

shammeer-s:
**This PR extends the existing `slice_plot` functionality to support 2D and 3D visualizations of objective functions. The main enhancements include:**

- Support for 3D surface plots and 2D contour plots using Plotly
- Pairwise visualization grid similar to `seaborn.pairplot` for higher-dimensional problems
- Separation of concerns: Function evaluations are decoupled from plotting logic
- Parallelized evaluations using batch evaluators for efficient performance
- Modular function evaluation handled in `plot_data.py`
- Unified interface: Users can choose between "slice", "3d", or "contour" via the projection argument in slice_plot

**Future Improvements**

- Clean up repeated code by moving common plot logic into helper functions.
- Simplify 3D subplot layout and camera setup to make the code easier to manage.
- Add interactive elements or show useful plots (like 1D slices) on the diagonal of pairwise plots.
- Move the 3D and contour plot logic into their own well-named functions for better structure.
- Rename some function parameters to make their purpose easier to understand.
- Use an enum to clearly define chart types and make the code safer and easier to follow.

### DIFF
--- a/src/optimagic/visualization/plot_data.py
+++ b/src/optimagic/visualization/plot_data.py
@@ -1,0 +1,230 @@
+import warnings
+from functools import partial
+
+import numpy as np
+import pandas as pd
+from pybaum import tree_just_flatten
+
+from optimagic import deprecations
+from optimagic.batch_evaluators import process_batch_evaluator
+from optimagic.deprecations import replace_and_warn_about_deprecated_bounds
+from optimagic.optimization.fun_value import (
+    convert_fun_output_to_function_value,
+    enforce_return_type,
+)
+from optimagic.parameters.bounds import pre_process_bounds
+from optimagic.parameters.conversion import get_converter
+from optimagic.parameters.tree_registry import get_registry
+from optimagic.shared.process_user_function import infer_aggregation_level
+from optimagic.typing import AggregationLevel
+
+
+def evaluate_func(params, func, func_kwargs):  # type: ignore
+    """Evaluate a user-defined function, handling deprecated dictionary output.
+
+    Args:
+        params: Input parameters for the function.
+        func: The user-defined objective function.
+        func_kwargs: Optional dictionary of keyword arguments to pass to the function.
+
+    Returns:
+        A tuple of (possibly wrapped) function and its evaluated output.
+
+    """
+    if func_kwargs:
+        func = partial(func, **func_kwargs)
+
+    func_eval = func(params)
+
+    if deprecations.is_dict_output(func_eval):
+        warnings.warn(
+            "Functions that return dictionaries are deprecated and will "
+            "raise an error in future versions.",
+            FutureWarning,
+        )
+        func_eval = deprecations.convert_dict_to_function_value(func_eval)
+        func = deprecations.replace_dict_output(func)
+
+    problem_type = (
+        deprecations.infer_problem_type_from_dict_output(func_eval)
+        if deprecations.is_dict_output(func_eval)
+        else infer_aggregation_level(func)
+    )
+    func_eval = convert_fun_output_to_function_value(func_eval, problem_type)
+    func = enforce_return_type(problem_type)(func)
+    return func, func_eval
+
+
+def process_bounds(bounds, lower_bounds, upper_bounds):  # type: ignore
+    """Process parameter bounds, replacing deprecated formats if necessary.
+
+    Args:
+        bounds: Bound object or structure.
+        lower_bounds: Deprecated lower bounds.
+        upper_bounds: Deprecated upper bounds.
+
+    Returns:
+        Processed and validated bounds.
+
+    """
+    bounds = replace_and_warn_about_deprecated_bounds(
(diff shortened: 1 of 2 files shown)


timmens (commented):
Hey,

Thanks for the PR!

Before I start a thorough review of your changes, lets tackle some high-level issues:

1. Can you move the code from `plot_data.py` into the `slice_plot.py` module, unless it is also used in other plotting functions.
2. Can you re-order the functions such that the function that is exported by the module (slice_plot) is at the beginning, then the private slice plot functions, and lastly the plotting data code.
3. Can you create a small notebook where you showcase the usage of the new features (this does not need to pushed, you can just upload it in a comment here).

Thanks a lot! If you have any questions, let me know here or on Zulip!

shammeer-s:
Hey @timmens,

Thanks for the feedback!

Sure, I’ll move the code from `plot_data.py` into `slice_plot.py` as suggested, since it’s not being used elsewhere. I’ll also re-order the functions so that the exported slice_plot function comes first.

I'll work on a small notebook to demonstrate the new features and share it here once it’s ready.

Also, I’ll take this opportunity to improve the modularity of the plotting logic a bit—cleaning it up should make the flow easier to understand.

Let me know if there's anything else I should add or keep in mind!

timmens (changes requested):
Thanks so much—this is looking great already! A few things that need to be addressed before we continue:

#### General

- **Type annotations:** Mypy is complaining because there are missing annotations. Could you add argument and return type hints wherever the types are clear? For the exported function interfaces, feel free to silence the error (if you are unsure) with `# type: ignore[no-untyped-def]`.
- **Pre-commit hooks:** Ruff is flagging some issues. Please fix those and make sure all pre-commit hooks run and pass before pushing.
- **Tests:** Add `test_slice_plot_3d.py` to cover `slice_plot_3d.py` (use `test_slice_plot.py` as a template). Ideally include:
    - End-to-end tests calling `slice_plot_3d()` with various inputs.
    - Unit tests for each helper function.
- **Code:** To gain a better understanding how your code differs from our current `slice_plot` implementation, could you please align your code with that of `slice_plot`. Currently you are doing two things at the same time: (1) you are adding the 3D feature, and (2) you are refactoring the `slice_plot` code. It is hard to focus on (1) when you are also doing (2). For example, the function `evaluate_func` does not exist in the original implementation, and it is hard for me to see whether you added functionality there.

#### Visualization comments

> These comments arose after looking at the how-to notebook 

- **1D Slice Plot**
	- Remove the plot title `"Slice Plot"`.
	- Remove the legend entry `"trace 1"`.
- **3D Surface Plot (Two parameters)**
	- Remove the plot title `"Slice Plot"`.
	- Ensure both axes are fully visible (currently cropped, especially in the lower center).    
	- Rename the axes from `"x"`/`"y"` to `"alpha"`/`"beta"`.
- **2D Contour Plot (Two parameters)**
	- Remove the plot title `"Slice Plot"`.
	- Label the axes `"alpha"` and `"beta"`. 
	- Add a color legend so it’s clear which color matches each contour level.
- **Grid View (Multiple parameters)**
	- Use a selector to focus on the first three parameters.
	- Once the individual plots are fixed, also ensure:
    - No plot title `"Slice Plot"`.
    - No legend `"trace 1"`.
    - All axes visible in the 3D plots (no cropping).
	- Display parameter names outside the grid—once along the left edge and once along the bottom, similar to `sns.pairplot`.
- **Behavior change for `projection`:**
    - If `projection` is a string (or Enum), e.g., `"surface"`:
        - Diagonal: univariate slice plots
        - Lower triangle: 2D/3D plots based on that string
        - Upper triangle: empty
    - If `projection` is a dictionary `{"lower": "surface", "upper": "contour"}`:
        - Diagonal: univariate slice plots        
        - Lower triangle: 3D surface plots
        - Upper triangle: 2D contour plots
    - I’d recommend parsing `projection` early into `{"lower":…, "upper":…}`, where `None` means no plot (e.g. `"surface"` → `{"lower":"surface","upper":None}`). Given your implementation the dictionary values can be parsed to the Projection Enum at this point as well.
    
Thanks a lot! If you have any questions let me know!

shammeer-s:
Thank you for the feedback.

The notebook has been updated as requested. The following changes have been incorporated:

1. Replaced the `slice` projection keyword with `univariate`. Added an explicit example demonstrating how to use the projection argument as a dictionary to assign different projection types for the upper and lower triangular parts of the grid (e.g., `{"upper": "contour", "lower": "surface"}`).

2. Additionally, the concern regarding redundant grid evaluations has been addressed by introducing a **hashmap to cache** and reuse computed evaluation values for each unique parameter pair. This significantly improves computational efficiency by avoiding repeated evaluations of the same grid slices.

Please let me know if any further refinement is needed.

timmens (approved):
Thanks a lot @shammeer-s for the changes!

The PR is finally fully approved and will be merged 🎉

PR #589, opened 2025-04-22
Update algo selection pre-commit hook from within Python

timmens:
Previously we updated the algo selection via a bash script which was called through a local pre-commit hook. This does not work on Windows.

In this PR, I implement the bash script mechanism in Python directly.


janosg (approved):
Thanks! Surprising that no windows users complained before.

PR #594, opened 2025-05-05
Polish new algorithms

janosg:
Polish and align a few things that slipped through the reviews for new algorithms. 

- [x] Improve error message if iminuit is not installed
- [x] Set budget for PSO higher to reduce random failures
- [ ] Test stochastic algorithms separately from deterministic algorithms in `test_many_algorithms`. Stochastic algorithms would be all algorithms that have a `seed` option. They should be tested with a fixed seed to avoid spurious failures.

### DIFF
--- a/src/optimagic/optimizers/iminuit_migrad.py
+++ b/src/optimagic/optimizers/iminuit_migrad.py
@@ -1,11 +1,12 @@
 from dataclasses import dataclass
-from typing import Optional
+from typing import TYPE_CHECKING, Any, Optional
 
 import numpy as np
 from numpy.typing import NDArray
 
 from optimagic import mark
 from optimagic.config import IS_IMINUIT_INSTALLED
+from optimagic.exceptions import NotInstalledError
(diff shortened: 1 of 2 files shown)


timmens:
I've implemented point 3 by splitting all test functions into a stochastic and deterministic one. Alternatively, we could also define `algo_options` conditionally inside the test function depending on whether the algorithm is stochastic or not.

janosg:
> I've implemented point 3 by splitting all test functions into a stochastic and deterministic one. Alternatively, we could also define `algo_options` conditionally inside the test function depending on whether the algorithm is stochastic or not.

Thinking about that: We could just have called every algorithm with a seed. Optimagic would discard it automatically where it doesn't apply 🤦 

On the other hand: Separating the stochastic algorithms will allow us to use something like [flaky](https://pypi.org/project/flaky/) to re-run stochastic tests which is something we definitely don't want to do for deterministic algorithms. 

So I would say we keep the separation.

timmens (approved):

PR #598, opened 2025-05-17
Add/fix links to github in a few more places

hamogu:
### What problem do you want to solve?

Reference the issue or discussion, if there is any. Provide a description of your
proposed solution.

### Todo

- [ ] Target the right branch and pick an appropriate title.
- [ ] Put `Closes #XXXX` in the first PR comment to auto-close the relevant issue once
  the PR is accepted. This is not applicable if there is no corresponding issue.
- [ ] Any steps that still need to be done.


hamogu:
@janosg I think this is ready. RTDs passed and looks good and there are no code changes, so it doesn't matter that GH Actions didn't run yet (because it needs approval). I don't have the power to "request a review", so this comment here will have to do.

janosg (approved):
Hi @hamogu, thank you so much for the PR and welcome to the optimagic community. 

Also thanks for the feedback in the other issues. We will work on them as quickly as possible. It is super valuable to get a fresh perspective on the documentation!

PR #599, opened 2025-05-19
Implementation of `matplotlib` backend for `criterion_plot()`

r3kste:
## Summary of Changes

1. Added `backend` parameter to `criterion_plot()` which accepts "plotly" or "matplotlib"
2. `criterion_plot()` now returns the figure object corresponding to the chosen backend.
3. Added `backends.py` which serves as a registry for all backends

##  To-Do

It is very much still a work in progress.

- [ ] Discuss on whether the dispatch mechanism can be improved.
- [ ] Add examples in documentation
- [ ] Add suitable tests

### Plot

|Plotly|Matplotlib|
|-|-|
|t|t|

### DIFF
--- a/src/optimagic/visualization/history_plots.py
+++ b/src/optimagic/visualization/history_plots.py
@@ -1,5 +1,6 @@
 import inspect
 import itertools
+from dataclasses import dataclass
 from pathlib import Path
 from typing import Any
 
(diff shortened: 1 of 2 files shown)


src/optimagic/visualization/backends.py
```diff
@@ -0,0 +1,111 @@
...
+        return decorator
+
+    @classmethod
+    def get_backend_wrapper(cls, backend_name: str) -> type[BackendWrapper]:
+        if backend_name not in cls._registry:
+            raise ValueError(
+                f"Backend '{backend_name}' is not supported. "
+                f"Supported backends are: {', '.join(cls._registry.keys())}."
+            )
+        return cls._registry[backend_name]
```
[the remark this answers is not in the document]
r3kste: > ```python
> def get_plotting_backend_class(backend: str) -> PlottingWrapper:
>     backend_to_class = {
>         "plotly": PlotlyBackend,
>         "matplotlib": MatplotlibBackend,
>     }
> ```

Is there any specific reason why the dictionary should be defined within the function rather than as a global object?

src/optimagic/visualization/backends.py
```diff
@@ -0,0 +1,111 @@
...
+    def line_plot(self, lines: list[LineData]) -> None:
+        pass
+
+    @abc.abstractmethod
+    def label(self, **kwargs):
+        pass
+
+    @abc.abstractmethod
+    def return_obj(self):
+        pass
```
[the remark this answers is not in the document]
r3kste: > To update the legend we should probably add a `set_legend_position` method.

I agree. I think `set_legend_props` might be a better fit as it can also be used for any additional properties specific to the legend.

> Where do we need to set the template? Can we set the template in the init?

1. plotly requires it to be _after_ creating the figure and 
2. matplotlib requires it _before_ creating the figure.

As we create the figure within the init, we can also set template in init.

pyproject.toml
timmens, line 1: I believe this was a mis-communication from my side. We should not add all backend as mandatory dependencies to optimagic -- they should be optional.

For now let's keep `plotly` as a mandatory dependency, but let's make `matplotlib` optional. That is,

1. We need to create the flag `IS_MATPLOTLIB_INSTALLED` (in `config.py`)
2. In the function `get_plot_backend_class` in `backends.py` we need check whether the backend is installed and if not raise a helpful error message (see below). I'd prefer an independent function that does this.

We can keep matplotlib in the testing environments!

```python
# error message example from tao_optimizers.py
if not IS_PETSC4PY_INSTALLED:
    raise NotInstalledError(
        "The 'tao_pounders' algorithm requires the petsc4py package to be "
        "installed. If you are using Linux or MacOS, install the package with "
        "'conda install -c conda-forge petsc4py'. The package is not available on "
        "Windows. Windows users can use optimagics 'pounders' algorithm instead."
    )
 ```

File: src/optimagic/visualization/backends.py [hunk 3]
```diff
@@ -0,0 +1,107 @@
...
+    def set_labels(self, xlabel: str | None = None, ylabel: str | None = None) -> None:
+        self._fig.update_layout(xaxis_title_text=xlabel, yaxis_title_text=ylabel)
+
+    def set_legend_props(self, legend_props: dict[str, Any]) -> None:
+        self._fig.update_layout(legend=legend_props)
+
+
+class MatplotlibBackend(PlotBackend):
+    default_template: str = "default"
+    default_palette: list = list(mpl.colormaps["Set2"].colors)
```
timmens, line 69: This line raises a mypy error. Can you fix that? If necessary just ignore it.

timmens (commented):
Thanks! This looks very very good now for the first proper version.

I only have a few comments. Most importantly, we need to make sure that `matplotlib` is treated as an optional dependency. Once this is handled we can move on to the other functions 🎉

src/optimagic/visualization/history_plots.py
```diff
@@ -8,36 +8,51 @@
...
 def criterion_plot(
-    results: OptimizeResultOrPath
-    | list[OptimizeResultOrPath]
-    | dict[str, OptimizeResultOrPath],
+    results: ResultOrPath | list[ResultOrPath] | dict[str, ResultOrPath],
     names: list[str] | str | None = None,
     max_evaluations: int | None = None,
-    template: str = PLOTLY_TEMPLATE,
-    palette: list[str] | str = PLOTLY_PALETTE,
+    backend: str = "plotly",
```
timmens, line 41: Can you make `backend` a literal with the respective entries? When experimenting I realized that there is not autocomplete for the backend yet.

src/optimagic/visualization/history_plots.py
timmens, line 1: I realized that we do not test the `matplotlib` backend. Can you choose one or two appropriate tests from `test_history_plots.py` and parametrize them over the implemented backends?

src/optimagic/visualization/backends.py
```diff
@@ -0,0 +1,131 @@
...
+    def set_labels(self, xlabel: str | None = None, ylabel: str | None = None) -> None:
+        self._fig.update_layout(xaxis_title_text=xlabel, yaxis_title_text=ylabel)
+
+    def set_legend_properties(self, legend_properties: dict[str, Any]) -> None:
+        self._fig.update_layout(legend=legend_properties)
+
+
+if IS_MATPLOTLIB_INSTALLED:
+
+    class MatplotlibBackend(PlotBackend):
```
timmens, line 73: If I experiment in a notebook, the matplotlib backend is shown twice, while the plotly figure is only shown once. Also if I assign the figure to a variable, plotly does not show anything (wanted behavior) while with the matplotlib backend the figure is still shown once.

This behavior should be aligned across backends, and we want the plotly bahavior.

timmens (commented):
Very nice, thanks a lot! Only a few clarifying comments, and then we are hopefully through with this.

src/optimagic/visualization/backends.py
```diff
@@ -0,0 +1,173 @@
...
+if IS_MATPLOTLIB_INSTALLED:
+    import matplotlib as mpl
+    import matplotlib.pyplot as plt
+
+    # Handle the case where matplotlib is used in notebooks (inline backend)
+    # to ensure that interactive mode is disabled to avoid double plotting.
+    # (See: https://github.com/matplotlib/matplotlib/issues/26221)
+    if mpl.get_backend() == "module://matplotlib_inline.backend_inline":
+        plt.install_repl_displayhook()
+        plt.ioff()
```
timmens, line 18: As discussed, we will check how other libraries solve this problem before continuing with this approach.

timmens (changes requested):
Looks very nice now! Thanks a lot. I only have a few comments on top of what we discussed in the call.

timmens (commented):
Looks very good, thanks!

Only two minor comments.

timmens (approved):
Approved, looks very good now! 🎉 

Thanks a lot!!

---

There actually seems to be an issue with the "tao_pounders" optimizer on Python 3.11, so we will have to wait until that is fixed before we merge.

PR #607, opened 2025-07-03
Ensure `exploration_results` is a numpy array

r3kste:
This PR ensures that `exploration_results` in `MultistartInfo` is always a numpy array.

### Summary

1. Fix a case where exploration_results is converted to a list.
2. Test that it is a numpy array of dtype `np.float64`.
3. Change type hint of exploration_results from `list[float]` to `NDArray[np.float64]`.

### Possible Changes
I would like to get feedback on whether the following changes should be included.

- Use a less strict check for dtype of `exploration_results`. For example, a [subdtype](https://numpy.org/doc/stable/reference/generated/numpy.issubdtype.html) of `np.floating`
- Add a check to assert that `exploration_results` is positive for `minimize` and negative for `maximize`.

### DIFF
--- a/src/optimagic/optimization/optimize_result.py
+++ b/src/optimagic/optimization/optimize_result.py
@@ -4,6 +4,7 @@ from typing import Any, Dict, Optional
 
 import numpy as np
 import pandas as pd
+from numpy.typing import NDArray
 
 from optimagic import deprecations
 from optimagic.logging.logger import LogReader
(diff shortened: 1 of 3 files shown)


timmens:
Hey @r3kste!

I discussed the problem / inconsistency with Janos. We realized that instead of adhering to the output of `run_explorations`, we'd rather change the output type! That is, we believe it is more fitting and in line with the rest of the codebase if `run_explorations` returns a dictionary of `list[float]`'s.

Can you rename the PR and make the required changes for this?

janosg:
The exploration result has fixed fields, so it should not be a dict at all but a dataclass (say InternalExplorationResult) so we get type checking for attribute access. 

The reason why I would have preferred a list is that this is in principle a history and for histories we usually use lists of floats or lists of arrays as this is the most intuitive way to store histories. However, I don't have a strong opinion here. So you can choose if InternalExplorationResult has list or array attributes but we definitely don't want to use a dict here.

PR #608, opened 2025-07-04
Add support for plotly >= 6

hmgaudecker:
In the current envs, incompatible versions of plotly and kaleido are installed. See docs here:

https://optimagic.readthedocs.io/en/latest/tutorials/optimization_overview.html#you-can-compare-optimizers

Updating them here.

There are some test failures, but AFAICT in tests that check deprecations / are for deprecated behavior. In particular, it looks like the `slice_plot` requires some plotly-internal representation of data, which seems to have changed. 

It certainly does not seem worth it to fix those tests -- two routes, I guess:
1. Just remove the tests and corresponding code if the deprecations are scheduled to be enforced in the next release
2. Pin `plotly < 6.0` and `kaleido < 0.3` instead of the update done here.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -127,6 +127,25 @@ jobs:
         run: |
           micromamba activate optimagic
           pytest -m "not slow and not jax"
+  run-tests-with-old-plotly:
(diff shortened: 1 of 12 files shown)


janosg:
Tim will work on this and then re-assign the review to me. We don't want to pin an old version but fix the code.

PR #615, opened 2025-07-11
Add pre-commit hook that checks mypy version consistency

timmens:
In this PR we add a local pre-commit hooks that checks whether the mypy version in the pre-commit config and the (conda) environment are the same.

This is supposed to help us keep these two versions the same, as otherwise we cannot ensure consistent results of calling mypy.


janosg (approved):
Thanks, looks good! I like that we just flag the problem so we can decide ourselves how to fix it.

PR #616, opened 2025-07-15
Pygad optimizer

spline2hg:


spline2hg:
| **PyGAD Parameter**            | **Optimagic Mapping**                                             |
|-------------------------------|--------------------------------------------------------------------|
| **User-facing**               |                                                                    |
| `num_generations`            | `num_generations`                                                  |
| `sol_per_pop`                | `population_size`                                                  |
| `num_parents_mating`         | `num_parents_mating`                                               |
| `initial_population`         | `initial_population`                                               |
| `parent_selection_type`      | `parent_selection_type`                                            |
| `keep_parents`               | `keep_parents`                                                     |
| `keep_elitism`               | `keep_elitism`                                                     |
| `K_tournament`               | `K_tournament`                                                     |
| `crossover_type`             | `crossover_type`                                                   |
| `crossover_probability`      | `crossover_probability`                                            |
| `mutation_type`              | `mutation_type`                                                    |
| `mutation_probability`       | `mutation_probability`                                             |
| `mutation_percent_genes`     | `mutation_percent_genes`                                           |
| `mutation_num_genes`         | `mutation_num_genes`                                               |
| `mutation_by_replacement`    | `mutation_by_replacement`                                          |
| `random_mutation_min_val`    | `random_mutation_min_val`                                          |
| `random_mutation_max_val`    | `random_mutation_max_val`                                          |
| `allow_duplicate_genes`      | `allow_duplicate_genes`                                            |
| `fitness_batch_size`         | `fitness_batch_size`                                               |
| `stop_criteria`              | `stop_criteria`                                                    |
| `random_seed`                | `random_seed`                                                      |
| `n_cores` *(Optimagic-only)* | `n_cores`                                                          |
| **Internal**                 |                                                                    |
| `fitness_func`              | Auto-generated from `problem.fun` / `problem.batch_fun`            |
| `num_genes`                 | Auto-detected from `len(x0)`                                       |
| `gene_type`                 | Fixed to `float64`                                                 |
| `gene_space`                | Auto-built from `problem.bounds.lower` and `problem.bounds.upper` |
| `parallel_processing`       | Fixed to `None` (parallelism handled via Optimagic)                |
| **Ignored**                 |                                                                    |
| `init_range_low`            |                                                                    |
| `init_range_high`           |                                                                    |
| `gene_constraint`           |                                                                    |
| `save_best_solutions`       |                                                                    |
| `save_solutions`            |                                                                    |
| `suppress_warnings`         |                                                                    |
| `on_start`                  |                                                                    |
| `on_fitness`                |                                                                    |
| `on_parents`                |                                                                    |
| `on_crossover`              |                                                                    |
| `on_mutation`               |                                                                    |
| `on_generation`             |                                                                    |
| `on_stop`                   |                                                                    |

src/optimagic/optimizers/pygad_optimizer.py
```diff
@@ -0,0 +1,292 @@
...
+    mutation_type: (
+        Literal["random", "swap", "inversion", "scramble", "adaptive"]
+        | MutationFunction
+        | None
+    ) = "random"
+    mutation_probability: (
+        NonNegativeFloat
+        | list[NonNegativeFloat]
+        | tuple[NonNegativeFloat, NonNegativeFloat]
+        | NDArray[np.float64]
```
janosg, line 78: From the current documentation I don't understand the cases other than a scalar probability. How many entries would have to be in the list. What happens if I specify a tuple with two values? How long would the array have to be?

At the moment I am not sure if just the documentation is missing or we might need to simplify which cases are allowed. 

The same comment applies to the next two options.

src/optimagic/optimizers/pygad_optimizer.py
```diff
@@ -0,0 +1,292 @@
...
+    mutation_num_genes: (
+        PositiveInt
+        | list[PositiveInt]
+        | tuple[PositiveInt, PositiveInt]
+        | NDArray[np.int_]
+        | None
+    ) = None
+    mutation_by_replacement: bool = False
+    random_mutation_min_val: float | list[float] | NDArray[np.float64] = -1.0
+    random_mutation_max_val: float | list[float] | NDArray[np.float64] = 1.0
```
janosg, line 97: In the array case: Does the array have one entry per parameter? If so, we would need o more general PyTree handling with the converter here.

src/optimagic/optimizers/pygad_optimizer.py
```diff
@@ -0,0 +1,292 @@
...
+
+    allow_duplicate_genes: bool = True
+
+    gene_constraint: list[GeneConstraintFunction | None] | None = None
+    sample_size: PositiveInt = 100
+
+    fitness_batch_size: PositiveInt | None = None
+    stop_criteria: str | list[str] | None = None
```
janosg, line 105: This option is insufficiently documented. 

Once I understand which values are allowed here we can think about aligning the handling of stopping criteria better with other optimizers in optimagic.

src/optimagic/optimizers/pygad_optimizer.py
```diff
@@ -0,0 +1,292 @@
...
+            warnings.warn(
+                f"fitness_batch_size ({fitness_batch_size}) is smaller than "
+                f"n_cores ({n_cores}). This may reduce parallel efficiency. "
+                f"Consider setting fitness_batch_size >= n_cores."
+            )
+        return fitness_batch_size
+    elif n_cores > 1:
+        return n_cores
+    else:
+        return None
```
janosg, line 256: Please use one return statement at the end. I.e. instead of returning multiple times, assign to a variable called `out` or `result` and then return that at the end.

src/optimagic/optimizers/pygad_optimizer.py
janosg: Then I think we should only support `ProbabilityFloat`, `None` and `tuple[...]`; Not the array and list ones.

janosg: I read your explanation a bit more carefully now: We want to avoid a situation like this where the **value** of one argument determines which **type** is valid for another one. We also want to avoid situations where the value of one argument determines whether another argument is used or ignored. 

This is the same situation we solve [here](https://optimagic.readthedocs.io/en/latest/development/ep-02-typing.html#configured-algorithms) by introducing configured algorithms. 

Since `mutation_functions` can be callables, the same solution will work. 

Concretely this means:

- For every mutation function supported by pygad we introduce a callable dataclass. The attributes of the class are the options supported by that mutation function. The `__call__` method just calls the corresponding function from pygad. 
- If a user wants to just select a mutation function but otherwise use default values they can pass a string (as now) or a class. 
- If a user wants to configure the mutation further, they need to pass an instance of the mutation functions. 
- If a user wants a completely customized mutation function they can either create a subclass or follow the protocol.
- The argument `mutation_type` should then be renamed to `mutation`.  

Usage example for a configured mutation function:

```python
algo = om.algos.pygad(
    mutation=om.pygad.AdaptiveMutation(
        probability_bad=0.3, # for low fitness values
        probability_good=0.1, # for high fitness values
    )
)
```

Name wise, we need to avoid "high" and "low" because the algorithms can be used in maximize and minimize as well as "fitness" because optimagic users know this as "fun". 

Let me know if this explanation was clear, otherwise we can have a short call again. A similar solution might also be necessary for parent selection functions etc.

spline2hg:
@janosg @timmens When building Sphinx docs for pygad_optimizer , type hints like:

```python
PositiveInt | None = None
````

are rendered as:

```
Annotated[int, Gt(gt=0)] | None = None
```

So we are getting kind of decomposed view ,This happens even if I simplify the annotation to just `PositiveInt` without unions or default values.

Interestingly, similar type hints render correctly for SciPy’s documentation.

Example build: [https://optimagic--616.org.readthedocs.build/en/616/algorithms.html#pygad-optimizer](https://optimagic--616.org.readthedocs.build/en/616/algorithms.html#pygad-optimizer)

This might be related to this Sphinx issue:
[https://github.com/sphinx-doc/sphinx/issues/10785](https://github.com/sphinx-doc/sphinx/issues/10785)

Do you have any suggestions for how to prevent Sphinx from expanding PositiveInt into its Annotated form in the rendered docs?

image

**Comment by timmens** [comment]
Hey! Can you try and add `from __future__ import annotations` to `pygad_optimizers.py`? I believe that is missing here, compared to the scipy optimizers. (I think I forgot to mention that in the how-to-document-optimizers guide.)

**If that fixes you problem:**
Could you add that to the how-to guide somewhere? (I.e., think about where in the guide it would be appropriate to mention that one has to add `from __future__ import annotations` to the import section in the optimizer module.) Feel free to tag me for feedback on the guide.

File: src/optimagic/optimization/algo_options.py [hunk 5]
```diff
@@ -108,6 +108,25 @@
...
+    this target. The definition of "better" depends on the optimization direction.
+
+    - Minimization: criterion <= target
+    - Maximization: criterion >= target
+
+    Used in population-based algorithms like genetic algorithms.
+    To disable, set to None.
+"""
+
+CONVERGENCE_SATURATE_GENERATIONS = None
```
janosg, line 122: Let's rename to `convergence_generations_noimprove` to align it better with [this](https://github.com/optimagic-dev/optimagic/blob/1f95f25223933a693ce9ca7e25cd73459ae93aed/src/optimagic/optimizers/nevergrad_optimizers.py#L153)

src/optimagic/optimization/algo_options.py
```diff
@@ -108,6 +108,25 @@
 
 """
 
+CONVERGENCE_TARGET_CRITERION = None
```
janosg, line 111: Let's rename to `CONVERGENCE_TARGET_VALUE`

janosg (commented):
This looks very good. Just a few minor comments.

PR #618, opened 2025-07-16
Improve bounds processing if no bounds are specified

### LINKED ISSUE #603: Improve bounds processing if no bounds are specified
## Current Situation

If no bounds are specified, the bounds arrays in the internal optimization problem are set to `-np.inf` and `np.inf`. 

## Problem 

- Array creation is slow so adding these arrays that are not very useful adds overhead on very fast problems 
- It becomes harder to check whether bounds were specified
- Some algorithms accept None as bounds but cannot handle infinite bounds 

## Solutions 

Set the bounds arrays to None if no bounds were specified.

timmens:
Closes #603

### DIFF
--- a/tests/optimagic/optimization/test_history_collection.py
+++ b/tests/optimagic/optimization/test_history_collection.py
@@ -33,12 +33,18 @@ def test_history_collection_with_parallelization(algorithm, tmp_path):
 
     path = tmp_path / "log.db"
 
+    algo_options = {"n_cores": 2}
+    if algorithm == "nevergrad_pso":
+        algo_options["stopping_maxfun"] = 15
+    else:
+        algo_options["stopping_maxiter"] = 3
+
     collected_hist = minimize(
         fun=mark.least_squares(lambda x: x),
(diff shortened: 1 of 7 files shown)


tests/optimagic/optimization/test_history_collection.py
```diff
@@ -33,12 +33,18 @@ def test_history_collection_with_parallelization(algorithm, tmp_path):
...
+    else:
+        algo_options["stopping_maxiter"] = 3
+
     collected_hist = minimize(
         fun=mark.least_squares(lambda x: x),
         params=np.arange(5),
         algorithm=algorithm,
         bounds=Bounds(lower=lb, upper=ub),
-        algo_options={"n_cores": 2, "stopping_maxiter": 3, "stopping_maxfun": 6},
+        algo_options=algo_options,
```
timmens, line 47: The current version raises warnings because the algorithms over which we loop in this test only have one of the stopping options. I would prefer a warning-free pytest output.

PR #620, opened 2025-07-17
Use interactive plotly figures in docs

timmens:
In this PR, we remove kaleido from the docs environment and simply render the plotly figures using a notebook renderer, which on top has the side effect of the figures being interactive.

In the past we included plotly figures in the documentation by rendering them as PNG via kaleido. kaleido >= 1 now requires Google Chrome to be installed, which we want to avoid.

> [!NOTE]
> Because of the mentioned kaleido issue, all plotly figures in the latest docs are currently broken.


docs/source/conf.py
```diff
@@ -171,7 +171,11 @@
 # "default.css" will overwrite the built-in "default.css".
 html_css_files = ["css/termynal.css", "css/termynal_custom.css", "css/custom.css"]
 
-html_js_files = ["js/termynal.js", "js/custom.js"]
+html_js_files = [
+    "js/termynal.js",
+    "js/custom.js",
+    "js/require.js",
```
timmens, line 177: We need `require.js` for plotly to work with sphinx. See [here](https://github.com/readthedocs/sphinx_rtd_theme/issues/788).

PR #621, opened 2025-07-22
feat: implement lazy loading for optional dependencies

### LINKED ISSUE #617: Move import of some optional dependencies into functions or methods
## Current situation

Imports of optimizer libraries we wrap are at the top of files inside `if IS_LIBRARY_INSTALLED` conditions. This means that all installed libraries are imported, even if they are not used, which slows down the general import time for optimagic. 

## Desired Situation

Most imports of optional dependencies should be inside the `_solve_internal_problem` method

## Exceptions

We want exceptions for non optional dependencies and optimizers designed for cheap cost functions. There the overhead of doing imports each time could have a too high runtime cost. Thus the exceptions are:

- scipy (non-optional dependency)
- pyensmallen (designed for every fast objective functions)
- nlopt (will become a non-optional dependency in the future)

## Note

We already have this for pyensmallen and bayesian optimization because it made type hinting easier. See discussion in #602 

@spline2hg would you like to tackle this?

spline2hg:
## Overview

This PR implements **lazy loading for optional dependencies** . Previously, many optional dependencies were imported at the top level of modules, which increased import times even when those libraries were not needed for a given run. With this change, optional dependencies are only imported within the functions or methods that actually require them.

## Key Changes

- Moved imports of optional dependencies (e.g., `pybobyqa`, `dfols`, `cyipopt`, `pygmo`, `nevergrad`, `tranquilo`, `petsc4py`, `fides`) into the specific functions or methods that use them, rather than at the top of the module.
- Updated `src/optimagic/config.py` to use `importlib.util.find_spec()` via a new `_is_installed()` helper function for dependency checks. As a result, the `IS_..._INSTALLED` flags are now set without actually importing the modules, which prevents unnecessary overhead at import time.
- No changes were made for libraries that are non-optional or designed for extremely fast objective functions (`scipy`, `nlopt`, `pyensmallen`), as discussed in #617.
- Type-checking (`TYPE_CHECKING`) is used for type hints where necessary.

## Related Issues

- Closes #617

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -7,15 +7,35 @@ releases are available on [Anaconda.org](https://anaconda.org/optimagic-dev/opti
 
 ## 0.5.2
 
-This minor release includes several bug fixes and small improvements. Many contributions
-in this release were made by Google Summer of Code (GSoC) 2025 applicants, with
-@gauravmanmode and @spline2hg being the accepted contributors.
+This minor release adds support for two additional optimizer libraries:
 
+- [Nevergrad](https://github.com/facebookresearch/nevergrad): A library for
+  gradient-free optimization developed by Facebook Research.
+- [Bayesian
+  Optimization](https://github.com/bayesian-optimization/BayesianOptimization): A
+  library for constrained bayesian global optimization with Gaussian processes.
+
+In addition, this release includes several bug fixes and improvements to the
+documentation. Many contributions in this release were made by Google Summer of Code
+(GSoC) 2025 applicants, with @gauravmanmode and @spline2hg being the accepted
+contributors.
+
+- {gh}`620` Uses interactive plotly figures in documentation ({ghuser}`timmens`).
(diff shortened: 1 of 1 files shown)


spline2hg:
### Update Status of Optimizer Libraries

| **Library**              | **Status**                          |
|--------------------------|--------------------------------------|
| `pygmo`                  | Updated                              |
| `cyipopt`                | Updated                              |
| `fides`                  | Updated                              |
| `petsc4py`               | Updated                              |
| `tranquilo`              | Updated                              |
| `pybobyqa`, `dfols`      | Updated                              |
| `bayes_opt`              | Already correct                      |
| `iminuit`                | Already correct                      |
| `scipy`                  | No change needed                     |
| `nlopt`                  | No change needed                     |
| `pyensmallen`            | No change needed                     |
| `neldermead`             | Not required                         |
| `bhhh`                   | Not required                         |
| `pounders`               | Not required                         |
| `nevergrad`              | Pending                              |

PR #633, opened 2025-08-19
Migrate bayesian-optimizer docs to new style

spline2hg:


janosg (commented):
The tutorial is very nice. I just have a few comments. Since it is hard to make line-by-line comments in a notebook, I'll add all of them here:

- You start by saying: "This tutorial demonstrates how to use the bayesian-optimizer ...". Since our string name is "bayes_opt" we should always refer to it as "bayes_opt" optimizer. 
- optimagic should always be written with a lowercase o 
- You state: "# bayesian-optimizer is primarily a maximization algorithm, optimagic provides a minimizer interface by negating the objective function" -> This is confusing and makes it sound like the optimizer is better suited for maximization problems than minimization problems but this is not the case as the two are mathematically equivalent after negating the objective function.

docs/source/tutorials/index.md
```diff
@@ -48,6 +48,25 @@ Learn numerical differentiation with estimagic.
...
+```{button-link} bayesian_optimizer_tutorial.html
+---
+click-parent:
+ref-type: ref
+class: stretched-link index-card-link sd-text-primary
+---
+Bayesian Optimizer
+```
+
+Tutorial on bayesian-optimizer in opimagic.
```
janosg, line 66: Change to "Tutorial on the `"bayes_opt"` optimizer in optimagic."

janosg (approved):
Thanks for the fixes!

janosg:
You can decide. Use what looks best.

PR #637, opened 2025-08-22
Add note about __future__ import

spline2hg:
Added a note in the `how_to_document_optimizers` guide about using `from __future__ import annotations` to ensure type hints are rendered correctly in Sphinx.


janosg:
The display is working now.

PR #639, opened 2025-08-27
Add optimizers from PySwarms

spline2hg:
This PR introduces the following optimizers from PySwarms:

- `pyswarms_global_best`
- `pyswarms_local_best`
- `pyswarms_general`

### DIFF
--- a/src/optimagic/optimizers/pyswarms_optimizers.py
+++ b/src/optimagic/optimizers/pyswarms_optimizers.py
@@ -218,6 +218,7 @@ class PySwarmsGlobalBestPSO(Algorithm):
 
     This algorithm is an adaptation of the original Particle Swarm Optimization method
     by :cite:`Kennedy1995`
+
     """
 
     n_particles: PositiveInt = 50
@@ -244,12 +245,20 @@ class PySwarmsGlobalBestPSO(Algorithm):
     boundary_strategy: Literal[
         "periodic", "reflective", "shrink", "random", "intermediate"
     ] = "periodic"
-    """Strategy for handling out-of-bounds particles. Available options: periodic (default),
-    reflective, shrink, random, intermediate."""
+    """Strategy for handling out-of-bounds particles.
+
+    Available options: periodic (default),
+    reflective, shrink, random, intermediate.
+
+    """
 
     velocity_strategy: Literal["unmodified", "adjust", "invert", "zero"] = "unmodified"
-    """Strategy for handling out-of-bounds velocities. Available options: unmodified (default),
-    adjust, invert, zero."""
+    """Strategy for handling out-of-bounds velocities.
+
+    Available options: unmodified (default),
+    adjust, invert, zero.
+
+    """
 
     velocity_clamp_min: float | None = None
     """Minimum velocity limit for particles."""
@@ -370,7 +379,9 @@ class PySwarmsLocalBestPSO(Algorithm):
         - :math:`\hat{y}_{lj}(t)`: local best position in particle i's neighborhood
 
     The algorithm is based on the original Particle Swarm Optimization method by
-    :cite:`Kennedy1995` and the local best concept introduced in :cite:`EberhartKennedy1995`.
+    :cite:`Kennedy1995` and the local best concept introduced in
+    :cite:`EberhartKennedy1995`.
+
     """
 
     n_particles: PositiveInt = 50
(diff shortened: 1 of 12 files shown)


src/optimagic/optimizers/pyswarms_optimizers.py
```diff
@@ -0,0 +1,913 @@
...
+    convergence_ftol_rel: NonNegativeFloat = CONVERGENCE_FTOL_REL
+    """Stop when relative change in objective function is less than this value."""
+
+    convergence_ftol_iter: PositiveInt = 1
+    """Number of iterations to check for convergence."""
+
+    stopping_maxiter: PositiveInt = STOPPING_MAXITER
+    """Maximum number of iterations."""
+
+    initial_population: NDArray[np.float64] | None = None
```
janosg, line 245: `list[PyTree]`, then use the converter to convert this to a list of arrays and then a 2d array from there.

spline2hg:
@janosg could you review this?

A few points:

1. Could you check the failing run on Python 3.11? I’m not sure what’s causing the issue.
2. PySwarms has stochastic algorithms but doesn’t expose a seed parameter. Because of that, they were being treated as deterministic in the tests. I’ve added a seed parameter with a random number generator for it to be treated as stochastic.
3. The default value CONVERGENCE_FTOL_REL = 2e-9 is causing the algorithm to converge earlier than expected(how should we handle this?).

src/optimagic/optimizers/pyswarms_optimizers.py
```diff
@@ -0,0 +1,904 @@
...
+
+    This algorithm is an adaptation of the original Particle Swarm Optimization method
+    by :cite:`Kennedy1995`
+
+    """
+
+    n_particles: PositiveInt = 50
+    """Number of particles in the swarm."""
+
+    options: PSOOptions = PSOOptions()
```
janosg, line 217: I don't like the nested structure. Instead the attributes of `PSOOptions` should be direct attributes of the algorithm. We had the exact same situation [here](https://github.com/optimagic-dev/optimagic/blob/381285002cff8edfb6c981f95a1b780fc9f19b64/src/optimagic/optimizers/gfo_optimizers.py#L39) and solved it via inheritance. I.e. in your case, all PSO algorithms would inherit from PSOOptions. 

[here](https://github.com/optimagic-dev/optimagic/pull/624#discussion_r2311826434) we are still discussing how to render a nice documentation in this case.

janosg:
> @janosg could you review this?
> 
> A few points:
> 
> 1. Could you check the failing run on Python 3.11? I’m not sure what’s causing the issue.
I also don't see the reason but it seems completely unrelated to your PR so you can ignore it for now. 

> 2. PySwarms has stochastic algorithms but doesn’t expose a seed parameter. Because of that, they were being treated as deterministic in the tests. I’ve added a seed parameter with a random number generator for it to be treated as stochastic.

If I see correctly, you only use the rng for inital positions. Is this the only stochastic part of the algorithm or are ther other stochastic parts that make the optimization non-deterministic even if a seed is set?

> 3. The default value CONVERGENCE_FTOL_REL = 2e-9 is causing the algorithm to converge earlier than expected(how should we handle this?).

You can deviate from this default if there are good reasons. Especially for a global optimizer I don't think it is problematic to set this to 0 because global optimizers are usually expected to run until `maxiter` or `maxfun` are reached. Does setting `convergence_ftol_rel=0` fix the problem?

src/optimagic/optimizers/pyswarms_optimizers.py
janosg: Wait, it looks like almost all options are the same for all three algorithms. Then we could extract all of those into a PySwarmsCommonOptions class.

src/optimagic/optimizers/pyswarms_optimizers.py
```diff
@@ -0,0 +1,904 @@
...
+            "options": pso_options_dict,
+            "static": self.topology.static,
+        }
+
+        res = _pyswarms_internal(
+            problem=problem,
+            x0=x0,
+            optimizer_class=ps.single.LocalBestPSO,
+            optimizer_kwargs=optimizer_kwargs,
+            n_particles=self.n_particles,
```
janosg, line 461: It might also make sense to pass a CommonOptions class (i.e. self) to `_pyswarms_internal` to avoid this long list of arguments. Would be similar to [here](https://github.com/optimagic-dev/optimagic/blob/381285002cff8edfb6c981f95a1b780fc9f19b64/src/optimagic/optimizers/gfo_optimizers.py#L692)

janosg (commented):
I don't have time to completely finish the review but this should give you some pointers to continue working on the PR.

spline2hg:
>If I see correctly, you only use the rng for inital positions. Is this the only stochastic part of the algorithm or are ther other stochastic parts that make the optimization non-deterministic even if a seed is set?

Stochastic parts in PySwarms:

1. Random topology and random bounds handling strategy.
2. Random components (r1, r2) in the velocity update equation, making the movement inherently stochastic.

>You can deviate from this default if there are good reasons. Especially for a global optimizer I don't think it is problematic to set this to 0 because global optimizers are usually expected to run until maxiter or maxfun are reached. Does setting convergence_ftol_rel=0 fix the problem?

If we set convergence_ftol_rel to 0, it disables early convergence and forces the algorithm to run for STOPPING_MAXITER iterations by default if stopping_maxiter is not set. However, running optimization for STOPPING_MAXITER times i.e (1_000_000) takes a long time.

janosg:
The failing test should be fixed after you update the branch.

src/optimagic/optimizers/pyswarms_optimizers.py
spline2hg: Yes, initially I thought having a common class would make documentation difficult, but it seems we do have a proper way to handle it.

janosg:
Sure! You can click on merge after checking!

spline2hg:
@janosg this looks good. The CI also passed, I’m merging it. Thanks for letting me work on this and for the help along the way. Feel free to tag me if anything related comes up.

PR #640, opened 2025-09-01
pin petsc4py to <= 3.23.4,  failing test python 3.11

gauravmanmode:
pin petsc4py to <= 3.23.4,
Failing test on python 3.11 with error
```
E   [0] Unable to find requested Tao type pounders
petsc4py/PETSc/TAO.pyx:190: Error
```


janosg (approved):
Thanks!

PR #647, opened 2025-10-18
Implementation of backend plotting for `convergence_plot`

r3kste:
## Summary of changes

1. Added `grid_line_plot` function that takes a `list[list[LineData]]` and returns a grid plot.
2. Modified `line_plot` to take in an optional parameter `subplot`. This allows `grid_line_plot` to re use `line_plot` functionality.
3. Refactored `convergence_plot` to use the backend plotting.

### Plot Changes

| Before | After |
| - | - |
| old | new |

### Backend Comparision

| Plotly | Matplotlib |
| - | - |
| new | output |

## PR Checklist

- [x] Compared images before and after making changes
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/src/optimagic/visualization/backends.py
+++ b/src/optimagic/visualization/backends.py
@@ -33,6 +33,8 @@ class LinePlotFunction(Protocol):
         """Protocol of the line_plot function used for type checking.
 
         Args:
+            ...: All other argument descriptions can be found in the docstring of the
+                `line_plot` function.
             subplot: The subplot to which the lines should be plotted. The type of this
                 argument depends on the backend used. If not provided, a new figure is
                 created.
@@ -76,6 +78,8 @@ def _line_plot_plotly(
     """Create a line plot using Plotly.
 
     Args:
+        ...: All other argument descriptions can be found in the docstring of the
+            `line_plot` function.
         subplot: A tuple specifying the subplot to which the lines should be plotted.
             The tuple contains the Plotly `Figure` object, the row index, and the column
             index of the subplot. If not provided, a new `Figure` object is created.
@@ -152,6 +156,7 @@ def _grid_line_plot_plotly(
         rows=n_rows,
         cols=n_cols,
         subplot_titles=titles,
+        horizontal_spacing=0.3 / n_cols,
     )
 
     for lines, (row, col) in zip(
@@ -193,6 +198,8 @@ def _line_plot_matplotlib(
     """Create a line plot using Matplotlib.
 
     Args:
+        ...: All other argument descriptions can be found in the docstring of the
+            `line_plot` function.
         subplot: A Matplotlib `Axes` object to which the lines should be plotted.
             If provided, the plot is drawn on the given `Axes`. If not provided,
             a new `Figure` and `Axes` are created.
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -173,6 +173,7 @@ else:
 nb_execution_mode = "force"  # "off", "force", "cache", "auto"
 nb_execution_allow_errors = False
 nb_merge_streams = True
+nb_scroll_outputs = True
 
 # Notebook cell execution timeout; defaults to 30.
 nb_execution_timeout = 1000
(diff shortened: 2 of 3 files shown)


tests/optimagic/visualization/test_convergence_plot.py
```diff
@@ -15,33 +37,20 @@
     {"normalize_distance": False},
     {"runtime_measure": "walltime"},
     {"runtime_measure": "n_batches"},
-    {"stopping_criterion": None},
+    # {"stopping_criterion": None},
```
r3kste: Hey @timmens 

I am not sure whether `stopping_criterion=None` is supported for `convergence_plot`, and by extension for `process_benchmark_results`?

The docstrings for these functions mention that it is indeed possible. However, passing `stopping_criterion=None` to convergence_plot or process_benchmark_results throws the following error:
`UnboundLocalError: local variable 'is_converged' referenced before assignment`

This behavior 'seems' to have been tested in this test, but **previously** due to the incorrect usage of `zip`, only the first two options were tested and the remaining options (including the `stopping_criterion=None`) were never executed by pytest.

I am not entirely sure as to how exactly `stopping_criterion` is used. So, I would appreciate it if you could clarify what the expected behavior should be in this case. Thanks!

src/optimagic/visualization/backends.py
```diff
@@ -25,6 +27,26 @@ def __call__(
...
+        n_cols: int,
+        titles: list[str] | None,
+        xlabel: str | None,
+        ylabel: str | None,
+        template: str | None,
+        height: int | None,
+        width: int | None,
+        legend_properties: dict[str, Any] | None,
+        margin_properties: dict[str, Any] | None,
     ) -> Any: ...
```
timmens: Can you add a descriptive docstring to this Protocol and the LinePlotFunction Protocol?

The idea would be that given the Protocol and the docstring, it should be clear to someone implementing a new backend on what to do...

src/optimagic/visualization/backends.py
```diff
@@ -25,6 +27,26 @@ def __call__(
         legend_properties: dict[str, Any] | None,
         margin_properties: dict[str, Any] | None,
         horizontal_line: float | None,
+        subplot: Any | None = None,
```
timmens: I like this `subplot` argument and how you use it a lot! Can you make sure that it is clear how to use it. That is, here in the protocol docstring you should explain it on an abstract level, and then I would expect to see a description of the argument in the docstring of the backend-specific functions. The docstrings of the backend-specific functions do not need to repeat all of the argument descriptions that are the same across backends, but only the ones that differ between backends.

r3kste:
> When going through the how-to guide where the convergence plot is shown, it seems like the figure goes over the page margin.

It occurs on my environment too. Moreover, it seems like it has been this way previously as well. 

One solution is to update the dimensions from 320x1000 to 320x840, so as to fit the figure within the margins. However, I also noticed that the dataframes being displayed in [Convergence report](https://optimagic.readthedocs.io/en/latest/how_to/how_to_benchmarking.html#a-convergence-report) also goes over the page margin.

So, I believe the ideal way to fix this issue for both cases would be to enable scrolling using myst-nb's [`nb_scroll_outputs` ](https://myst-nb.readthedocs.io/en/latest/configuration.html#nb_scroll_outputs:~:text=code%20cell%20outputs-,nb_scroll_outputs,-bool) option.
```diff
diff --git a/docs/source/conf.py b/docs/source/conf.py
index 89f9104..8c77a28 100644
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -173,6 +173,7 @@ else:
 nb_execution_mode = "force"  # "off", "force", "cache", "auto"
 nb_execution_allow_errors = False
 nb_merge_streams = True
+nb_scroll_outputs = True

 # Notebook cell execution timeout; defaults to 30.
 nb_execution_timeout = 1000
```

I would like to know your thoughts on this. Thanks!
Unrelated, but it looks like alternate rows in the displayed dataframe are not properly visible in dark mode.

PR #659, opened 2025-12-16
Fix UnboundLocalError when stopping_criterion=None in convergence_plot

### LINKED ISSUE #648: `UnboundLocalError` when `stopping_criterion = None` in convergence_plot().
### Bug description

Passing `stopping_criterion = None` to `convergence_plot()` or `process_benchmark_results()` raises 
`UnboundLocalError: local variable 'is_converged' referenced before assignment`, even though the [API](https://optimagic.readthedocs.io/en/latest/reference/index.html#optimagic.criterion_plot) mentions that `None` can be passed.

### To Reproduce

```python
import plotly.io as pio

import optimagic as om

problems = om.get_benchmark_problems("example")
optimizers = [
    "nag_dfols",
    "scipy_neldermead",
    "scipy_truncated_newton",
]
results = om.run_benchmark(
    problems,
    optimizers,
)

fig = om.convergence_plot(
    problems=problems,
    results=results,
    n_cols=2,
    problem_subset=["rosenbrock_good_start", "box_3d"],
    stopping_criterion=None,
)

fig.show()
```

### Expected behavior

The figure should be generated without any error. I am not sure whether the implementation is incorrect, or whether the docstring is outdated.

### Additional Information

The error occurs in the `_process_one_result()` function in `process_benchmark_results.py`, due to the absence of logic to handle the case when `stopping_criterion = None`

### Proposed Fix

This could be fixed by one of:

1. The case of `stopping_criterion = None` should be handled in the `_process_one_result()`,
2. The documentation should be updated to reflect that `stopping_criterion` can not be 'None'.

szd5654125:
### Summary
This PR fixes a bug where passing `stopping_criterion=None` to
`convergence_plot` / `process_benchmark_results` raises an
`UnboundLocalError`, although the API explicitly documents `None`
as a valid option.

### Changes
- Handle `stopping_criterion=None` explicitly in `_process_one_result`
  and ensure `is_converged` is always defined.
- Ensure histories are not truncated when no stopping criterion is applied.
- Update tests to reflect the documented behavior.

### Related issue
Closes #648

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -6,7 +6,7 @@ repos:
       - id: check-useless-excludes
         # - id: identity  # Prints all files passed to pre-commits. Debugging.
   - repo: https://github.com/lyz-code/yamlfix
-    rev: 1.19.0
+    rev: 1.19.1
     hooks:
       - id: yamlfix
         exclude: tests/optimagic/optimizers/_pounders/fixtures
@@ -89,7 +89,7 @@ repos:
           - --blank
         exclude: src/optimagic/optimization/algo_options.py
   - repo: https://github.com/astral-sh/ruff-pre-commit
-    rev: v0.14.8
+    rev: v0.14.10
     hooks:
       # Run the linter.
       - id: ruff
@@ -141,7 +141,7 @@ repos:
         args:
           - --drop-empty-cells
   - repo: https://github.com/pre-commit/mirrors-mypy
-    rev: v1.19.0
+    rev: v1.19.1
     hooks:
       - id: mypy
         files: src|tests
--- a/.tools/envs/testenv-linux.yml
+++ b/.tools/envs/testenv-linux.yml
@@ -26,7 +26,7 @@ dependencies:
   - scipy>=1.2.1  # run, tests
   - sqlalchemy  # run, tests
   - seaborn  # dev, tests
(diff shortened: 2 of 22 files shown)


r3kste (commented):
After looking through this, I am unsure whether `stopping_criterion` is actually allowed to be None. On one hand, the docstring of `process_benchmark_results` mentions that passing None is valid, although the logic for handling it was not present. On the other hand, `profile_plot` explicitly raises an error in this case.

Personally, I think that the docstring was wrong/outdated, and passing None to `stopping_criterion` was never supported. This change was made in [commit 68d941](https://github.com/optimagic-dev/optimagic/commit/68d941194b95f4cefd65afa336797c8abe8e429c), but it is not clear why.

@janosg It would be great if you could clarify whether `stopping_criterion` can be None, and if so, what the expected behavior should be.

janosg:
Thanks @r3kste and @szd5654125. 

I think the None case should actually not be supported. If someone wanted to disable all convergence criteria, they could set the precisions to extremely small values or even 0. 

Could we change this PR to improve the type hints and documentation for this?

szd5654125:
I agree that `stopping_criterion=None` should not be supported.
Given this, I can update this PR to:
clarify the documentation,
Restore the previous test cases,

I'll push an update shortly.

janosg (approved):

PR #679, opened 2026-04-01
Migrate pybaum to optree

abelaba:
# Summary

  - Added wrapper methods that use optree for all pytree operations used in optimagic (`tree_unflatten`, `tree_flatten`, `tree_leaves`(Previously `tree_just_flatten`), `tree_map`, `tree_equal`, `leaf_names`)
  - Removed pybaum as a dependency and added optree>=0.19.0.
  - Added namespace constants (`OPTREE_NAMESPACES` tuple and `DEFAULT_NAMESPACE`) in typing.py 
  - tree_registry.py: Rewritten to use `optree.register_pytree_node`, including support for numpy arrays, pandas DataFrame/Series, and JAX arrays.
  - Removed the `get_registry` method. So now, when extending the set of types that are considered internal nodes is desired, one would pass a namespace that exists in `OPTREE_NAMESPACES` as an argument to the tree methods.

#### Extending the set of types that are considered internal nodes for numpy arrays, pandas DataFrame/Series, and JAX arrays.
Previously
```python
registry = get_registry(extended=True) # data_col="value" is a default parameter inside this method
tree_flatten(tree, registry=registry)
```
Now
```python
tree_flatten(tree, namespace=<namespace>) # <namespace> is a string in OPTREE_NAMESPACES
```

#### When custom node method parsing is not desired.
Previously
```python
tree_flatten(tree)
```
Now
```python
tree_flatten(tree)
```

### Why I added wrapper methods on top of optree tree methods instead of directly using them throughout the project.

  1. **Flattening**: optree does not preserve insertion order of dictionaries by default ([documentation](https://github.com/metaopt/optree?tab=readme-ov-file#key-ordering-for-dictionaries)). The wrapper ensures insertion order for dictionaries is maintained for the registered namespaces in `OPTREE_NAMESPACES` and `DEFAULT_NAMESPACE`. (The following tests in the screenshot below fail if the order is not maintained for a dictionary)
Screenshot 2026-04-02 at 18 26 42

  2. **Unflattening**: pybaum and optree have different tree structures. The `tree_unflatten` wrapper maintains backward compatibility for methods that use pybaum's tree structure.

# Remaining
  - Flattening and unflattening for torch tensors.
 
# Test plan
  - All existing fast tests pass with the new optree backend.
  - Added additional tests for verifying the outputs of the tree methods.
     - Check tree method outputs for default namespace and registered namespaces.
     - Verify warning is raised for an unregistered namespace.
     - Tree method outputs when tree is None.
     - Verify dictionary insertion ordering is respected for all registered namespaces and the default namespace.


src/optimagic/differentiation/derivatives.py
```diff
@@ -214,24 +216,23 @@ def first_derivative(
...
     # Convert scalar | pytree arguments to 1d arrays of floats
     # ==================================================================================
-    registry = get_registry(extended=True)
 
     is_fast_path = _is_1d_array(params)
 
     if not is_fast_path:
-        x, params_treedef = tree_flatten(params, registry=registry)
-        x = np.array(x, dtype=np.float64)
+        params_leaves, params_treedef = tree_flatten(params, namespace=VALUE_NAMESPACE)
```
abelaba, line 223: Changed the name from `x` to `params_leaves` to fix type hint error.

src/optimagic/differentiation/derivatives.py
```diff
@@ -271,7 +272,7 @@ def first_derivative(
     step_size = cast(NDArray[np.float64], step_size)
 
     # generate parameter vectors at which func has to be evaluated as numpy arrays
-    evaluation_points = []
+    evaluation_points: list[float | np.ndarray] = []
```
abelaba, line 275: Added this type hint because I was getting mypy type error.

src/optimagic/differentiation/derivatives.py
```diff
@@ -529,24 +530,23 @@ def second_derivative(
...
     # Convert scalar | pytree arguments to 1d arrays of floats
     # ==================================================================================
-    registry = get_registry(extended=True)
 
     is_fast_path = _is_1d_array(params)
 
     if not is_fast_path:
-        x, params_treedef = tree_flatten(params, registry=registry)
-        x = np.array(x, dtype=np.float64)
+        params_leaves, params_treedef = tree_flatten(params, namespace=VALUE_NAMESPACE)
```
abelaba, line 537: Changed the name from `x` to `params_leaves` to fix type hint error.

src/optimagic/optimization/fun_value.py
```diff
@@ -124,8 +123,8 @@ def _get_flat_value(value: PyTree) -> NDArray[np.float64]:
     elif isinstance(value, np.ndarray):
         flat = value.flatten()
     else:
-        registry = get_registry(extended=True)
-        flat = tree_just_flatten(value, registry=registry)
+        value_leaves = tree_leaves(value, namespace=VALUE_NAMESPACE)
+        flat = np.asarray(value_leaves, dtype=np.float64)
```
abelaba, line 127: Converting to numpy array after flattening to fix mypy type error.

src/optimagic/parameters/tree_registry.py
```diff
@@ -1,50 +1,209 @@
...
+    equality_checkers[pd.Series.__name__] = lambda a, b: a.equals(b)
+    equality_checkers[pd.DataFrame.__name__] = lambda a, b: a.equals(b)
+
+    if _has_jax:
+        equality_checkers[jnp.ndarray.__name__] = lambda a, b: bool((a == b).all())
+
+    return equality_checkers
+
+
+def _check_namespace(namespace: str) -> None:
```
abelaba, line 153: Should I update this method to change the namespace to `DEFAULT_NAMESPACE` and raise a warning if an unregistered namespace is used?

PR #681, opened 2026-04-07
chore: set permission levels to read and pin specific commit hashes

abelaba:
## Summary
  - Set top-level permissions `contents: read` for main.yml workflow.
  - Pinned all third-party GitHub Actions to specific commit SHAs instead of mutable tags to prevent supply-chain attacks via tag reassignment using `pinact`.
  - Added `persist-credentials: false` to all actions/checkout steps to avoid exposing the `GITHUB_TOKEN` to subsequent steps.
  - Split the build-n-publish job into separate jobs to avoid sharing permissions.
  - Removed the PyPI secret token to use trusted publishing.
## Action Items before merging
  - Trusted publishing must be configured on PyPI. [Documentation for setting up PyPI Trusted publishing](https://docs.pypi.org/trusted-publishers/adding-a-publisher/)
       - The workflow name is `publish-to-pypi.yml`.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -8,9 +8,3 @@ updates:
       interval: weekly
     cooldown:
       default-days: 7
-  - package-ecosystem: pre-commit
-    directory: /
-    schedule:
-      interval: weekly
-    cooldown:
-      default-days: 7
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,16 @@
+---
+version: 2
+# update once a week, with a 7-day cooldown
+updates:
+  - package-ecosystem: github-actions
+    directory: /
+    schedule:
+      interval: weekly
+    cooldown:
+      default-days: 7
+  - package-ecosystem: pre-commit
+    directory: /
+    schedule:
+      interval: weekly
+    cooldown:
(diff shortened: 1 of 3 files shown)


AdrianoKF:
@abelaba - if you need help in setting up Trusted Publishing, just ping me! After the first release, let's also validate that attestations are correctly uploaded to PyPI for the release files.

abelaba:
@AdrianoKF Thank you for the review. I used `pinact run` and didn't update the actions to the latest versions because I wasn't sure if it would bring any breaking changes.  
I also directly used the versions for `actions/upload_artifact` and `actions/download_artifact` from the `lakefs-spec` repo workflow, but I see there are newer versions and I they weren't in the project previously so will update those to the latest versions.

abelaba:
## New changes
 - Added dependabot configuration for automatically updating GitHub actions `which updates once a week, with a 7-day cooldown`.
 - Inspired by these documentations:
   - https://astral.sh/blog/open-source-security-at-astral.
   - https://blog.yossarian.net/2025/11/21/We-should-all-be-using-dependency-cooldowns

abelaba:
I reverted the changes because creating a separate job for codecov only is excessive.

PR #683, opened 2026-04-30
Allow fixed entries in probability constraints

hmgaudecker:
In an [application](https://github.com/OpenSourceEconomics/skillmodels/pull/89), I need probability constraints  where one element is fixed. This is currently rejected by optimagic; this PR fixes that.


tests/optimagic/optimization/test_with_constraints.py
```diff
@@ -345,6 +345,52 @@ def criterion(x):
...
+        algorithm="scipy_lbfgsb",
+        constraints=[
+            om.ProbabilityConstraint(lambda x: x[[0, 1, 2, 3, 4]]),
+            om.FixedConstraint(lambda x: x[[0, 2]]),
+        ],
+    )
+
+    assert res.params[0] == 0.0
+    assert res.params[2] == 0.0
+    aaae(res.params[[1, 3, 4]].sum(), 1.0)
```
janosg: The first three checks are redundant since the last checks the exact values of the result vector.

tests/optimagic/optimization/test_with_constraints.py
```diff
@@ -345,6 +345,52 @@ def criterion(x):
...
+def test_probability_constraint_with_non_zero_fix_on_selector_element():
+    """One selected entry is fixed to 0.2; the remaining free entries sum to 0.8."""
+
+    def criterion(params):
+        target = np.array([0.2, 0.2, 0.3, 0.3])
+        return np.sum((params - target) ** 2)
+
+    res = minimize(
+        fun=criterion,
+        params=np.array([0.2, 0.3, 0.2, 0.3]),
```
janosg: The start vector shoudl be harder, i.e. more different from the target.

tests/optimagic/optimization/test_with_constraints.py
```diff
@@ -345,6 +345,52 @@ def criterion(x):
...
+        algorithm="scipy_lbfgsb",
+        constraints=[
+            om.ProbabilityConstraint(lambda x: x[[0, 1, 2, 3]]),
+            om.FixedConstraint(lambda x: x[[0]]),
+        ],
+    )
+
+    assert res.params[0] == 0.2
+    aaae(res.params[1:].sum(), 0.8)
```
janosg: redundant

tests/optimagic/optimization/test_with_constraints.py
```diff
@@ -345,6 +345,52 @@ def criterion(x):
...
+    aaae(res.params, [0.2, 0.2, 0.3, 0.3], decimal=4)
+
```
janosg, line 416: Let's add one test case with more than one fixed parameter

tests/optimagic/optimization/test_with_constraints.py
```diff
@@ -345,6 +345,52 @@ def criterion(x):
         )
 
 
+def test_probability_constraint_with_zero_fix_on_selector_element():
```
janosg, line 348: In all of the tests below the probability constraint is not really binding because the unconstrained optimum of the function happens to satisfy the constraint. We need harder test cases!

timmens (approved):
Found one functional issue: folding a fixed original pivot could leave a zero free entry as the new pivot, causing division by zero in probability_to_internal. Commit 005ea3a selects a positive free pivot when required and adds focused plus end-to-end regression coverage. It also corrects the Jacobian formula documentation.

PR #684, opened 2026-06-25
Add user-suppliable batch_evaluator argument to minimize/maximize

hmgaudecker:
## What

Threads a `batch_evaluator` keyword from the public `minimize`/`maximize` API through
`create_optimization_problem` and the `OptimizationProblem` container to
`InternalOptimizationProblem`, replacing the hardcoded `process_batch_evaluator("joblib")`
in `_optimize`. Users can now pass either a built-in name (`"joblib"`, `"pathos"`,
`"threading"`) **or a callable** conforming to the `BatchEvaluator` protocol.

Also:
- Removes the dead `batch_evaluator` `Literal` field from the `Tranquilo` / `TranquiloLS`
  dataclasses. It was never read — the evaluator lives on the optimization problem, not
  the algorithm — so setting it silently no-op'd.
- Exports `BatchEvaluator` from the top-level namespace so users can type their custom
  evaluators against the protocol.

## Why

The batched criterion evaluations a trust-region optimizer like `tranquilo` requests were
locked to the in-process `joblib` evaluator. Some workloads need a different execution
backend — e.g. a distributed, MPI-backed pool that fans each batch of candidate points
out across nodes/GPUs for an expensive simulation-based criterion. There was already a
`process_batch_evaluator` resolver and a `BatchEvaluator` protocol; this just lets the
caller reach them.

This is also the foundation for a follow-up: a first-class **executor-backed**
(`concurrent.futures` / `mpi4py.futures`) distributed evaluator, where `minimize` itself
parks non-root ranks in a worker pool so the optimizer runs on a single driver. That
addresses a real footgun — under MPI, `srun` starts N copies of the program, so a naive
"run `minimize` on every rank with a collective evaluator" design ends up with N
optimizers kept in lockstep, which diverges under floating-point nondeterminism. This PR
is the minimal, backward-compatible plumbing that makes the safe design expressible.

## Tests

Adds `test_minimize_uses_custom_batch_evaluator`,
`test_maximize_uses_custom_batch_evaluator` (a spy callable is the one optimagic actually
invokes), and `test_minimize_rejects_unknown_batch_evaluator` (clear error on an unknown
name).

## Backward compatibility

Fully backward compatible — `batch_evaluator` defaults to `"joblib"`, the prior behavior.

---

Happy to add a `CHANGES.md` entry referencing this PR number under the appropriate
next-version heading (0.5.4 vs 0.6.0 — your call).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -5,6 +5,12 @@ chronological order. We follow [semantic versioning](https://semver.org/) and al
 releases are available on [Anaconda.org](https://anaconda.org/optimagic-dev/optimagic).
 
 
+## 0.6.0
+
+- {gh}`684` Adds a user-suppliable `batch_evaluator` argument to `minimize` and
+  `maximize` ({ghuser}`hmgaudecker`).
+
+
 ## 0.5.3
 
 This release introduces **multi-backend plotting** with support for matplotlib, bokeh,
--- a/src/optimagic/__init__.py
+++ b/src/optimagic/__init__.py
@@ -47,6 +47,7 @@ from optimagic.optimizers import pygad
 from optimagic.parameters.bounds import Bounds
 from optimagic.parameters.constraint_tools import check_constraints, count_free_params
 from optimagic.parameters.scaling import ScalingOptions
+from optimagic.typing import BatchEvaluator
 from optimagic.visualization.convergence_plot import convergence_plot
 from optimagic.visualization.history_plots import criterion_plot, params_plot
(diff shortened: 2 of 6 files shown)


timmens (commented):
Looks very good, thanks!

Did you discuss already that this should be part of 0.6.0 instead of 0.5.4 (which we could probably release a bit faster than 0.6.0)?

PR #690, opened 2026-07-04
Convert all algorithm documentation to the new format.

janosg:
Previously, algorithm documentation was in algrithms.md; Now it's in docstrings of the algorithm classes. The details of the new format are documented [here](https://optimagic.readthedocs.io/en/latest/how_to/how_to_document_optimizers.html). During conversion, we discovered and fixed the following issues:

### scipy

- **Stale constraint-support notes.** The old text said "SLSQP's / COBYLA's /
  trust_constr's general nonlinear constraints are not supported yet by optimagic".
  All three have `supports_nonlinear_constraints=True` today. The new docstrings
  instead explain how optimagic converts equality constraints to inequality pairs
  for COBYLA (which natively only handles inequalities).
- **Copy-paste error in `scipy_newton_cg`.** The warning advised refining "an optimum
  found with Powell" with another optimizer — it meant an optimum found with
  newton_cg.
- **Wrong default for `scipy_direct`.** The old text claimed the volume tolerance
  `eps` defaults to 1e-6, "differing from scipy's default 1e-4". The actual optimagic
  default is `CONVERGENCE_FTOL_REL = 2e-9`.
- **Nonexistent options documented.** `scipy_truncated_newton` was documented with
  `stopping.maxiter`, but the class only has `stopping_maxfun`; `scipy_ls_lm` was
  documented with `tr_solver`/`tr_solver_options`, which only exist on the trf and
  dogbox least-squares solvers. Dropped.
- Typo `finitie_difference_precision`; several places where optimagic defaults
  silently differ from SciPy are now stated explicitly (Nelder-Mead ftol 1e-8 vs
  SciPy 1e-4, SLSQP 1e-8 vs 1e-6, BFGS xrtol 1e-5 vs 0, trust_constr xtol 1e-5 vs
  1e-8, dual_annealing maxfun 1e6 vs 1e7, differential_evolution atol 1e-8 vs 0).

### nlopt

- **MMA/CCSAQ constraint support was backwards.** The old text said these support
  "nonlinear equality (but not inequality) constraints". NLopt supports the
  opposite: nonlinear *inequality* (not equality) constraints. In optimagic, MMA
  gets equalities via conversion to inequality pairs, and CCSAQ currently exposes no
  nonlinear constraints at all (`supports_nonlinear_constraints=False`).
- **Wrong dropdown title.** The section was titled `nlopt_lbfgs`, but the registered
  algorithm name is `nlopt_lbfgsb`.
[...]

PR #691, opened 2026-07-05
Fix algo option type conversion being silently skipped with postponed annotations

janosg:
## Problem

`Algorithm.__post_init__` converts and validates algo option values by looking up each dataclass field's type in `TYPE_CONVERTERS`, which is keyed by type objects (`PositiveInt`, `int`, ...). In modules with `from __future__ import annotations` — which the [how-to-document-optimizers guide](https://optimagic.readthedocs.io/en/latest/how_to/how_to_document_optimizers.html) requires so that autodoc renders type aliases readably — `field.type` is the annotation *string* (`"PositiveInt"`), so the lookup never matched and coercion/validation was silently skipped.

This already affected every optimizer module that uses the import on main (scipy, gfo, nevergrad, pyswarms, pygad, bayesian, pygmo, tranquilo, iminuit) and surfaced as a test failure in #690 when `ipopt.py` gained the import: `test_ipopt_algo_options` passes `5.5` for an integer-typed Ipopt option and relies on coercion to int.

## Fix

- `TYPE_CONVERTERS_BY_NAME` in `type_conversion.py`: a string-keyed companion table with the same converters.
- `Algorithm.__post_init__` branches on whether `field.type` is a string and uses the matching table, so coercion works identically with and without the future import.

## Tests

- `test_algorithm_future_annotations.py` (new): a dummy algorithm defined under postponed annotations, mirroring the existing type conversion tests in `test_algorithm.py`, plus a guard test asserting the field types really are annotation strings.
- `test_algorithm.py`: a mirror guard asserting the existing dummy's field types are type objects (so the two modules keep covering both code paths), a float→int coercion test, and `test_int_options_are_coerced_for_all_algorithms`, which passes floats for all int-typed options of every algorithm in `ALL_ALGORITHMS` and asserts they come back as ints.

Without the fix, 97 of the new tests fail (the postponed-annotations module and the registry-wide test for ~90 algorithms); with it, everything passes.

#690 is rebased on top of this branch.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/src/optimagic/optimization/algorithm.py
+++ b/src/optimagic/optimization/algorithm.py
@@ -14,7 +14,7 @@ from optimagic.optimization.history import History
 from optimagic.optimization.internal_optimization_problem import (
     InternalOptimizationProblem,
 )
-from optimagic.type_conversion import TYPE_CONVERTERS
+from optimagic.type_conversion import TYPE_CONVERTERS, TYPE_CONVERTERS_BY_NAME
 from optimagic.typing import AggregationLevel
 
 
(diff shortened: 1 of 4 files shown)


janosg:
Closed in favor of #696

PR #693, opened 2026-08-02
Docs: explain AggregationLevel / problem types

### LINKED ISSUE #595: Docs: Explain problem types (`AggregationLevel`) in the docs
In the docs, the [How to write objective functions](https://optimagic.readthedocs.io/en/latest/how_to/how_to_criterion_function.html?highlight=least_squares) section has this intriguing sentence: 
> To use least-squares optimizers in optimagic, you need to mark your function with a decorator and return the least-squares residuals instead of the aggregated function value.

However, there is no more detail what this does and what it's good for. So, I decided to start reading the code and found that it's an `AggregationLevel`, but, going back to the docs (as a user, I try to avoid spending too much time to read the code to find out what it does), `AggregationLevel` is only mentioned [in passing](https://optimagic.readthedocs.io/en/latest/how_to/how_to_add_optimizers.html?highlight=aggregationlevel) in an example with the note "the type of problem this optimizer can solve -> scalar problems; Other optimizers solve likelihood or least_squares problems."

At this point, at least I found that there are in fact three Aggregationlevels: scalar, likelihood, and least-squares. 

Some of my objective functions will have a least-square structure, others are log-likehoods, so I wonder if I should mark those. I don't think this needs separate examples, a short (1-3 sentence each) explanation in the docs about the following would suffice:
- What is the purpose of setting those levels / use the decorators to mark objective functions (as far as I've found out so far, they change the return format expected from the objective function (scalar vs. array) and also the penalty function)?
- E.g. for the likelihood, do you expect the objective function to return a number or an array? The summed log-likelihood? A positive or negative number (in my field people commonly use the negative log-likelihood, because then you can call common minimizers and don't have to worry about "maximizers")?
- Similar 1-2 sentences for scalar and least-square.

A good place for that might be in the "How to write an objective function" guide, but it could also live in the API docs, if it's called out in the objective function guide.

Pragati5-DEBUG:
## Summary
- Adds an Explanation page for scalar, least-squares, and likelihood problem types (`AggregationLevel`)
- Clarifies what each `@om.mark.*` decorator expects as a return value, and maximize vs minimize for likelihood
- Links the new page from the “How to write objective functions” guide

Closes #595

## Test plan
- [ ] Build/docs preview: Explanation → Problem types (`AggregationLevel`)
- [ ] Confirm link from How to write objective functions (least-squares section)
- [ ] Check that the page answers the questions raised in #595 (purpose of marks, return types, likelihood sign)

### DIFF
--- a/docs/source/explanation/aggregation_level.md
+++ b/docs/source/explanation/aggregation_level.md
@@ -15,8 +15,7 @@ optionally `@om.mark.scalar`). That mark changes:
 1. **Which specialized optimizers you can use** (for example pounders for least-squares,
    or BHHH for likelihood).
 1. **How error penalties and derivatives are interpreted** when something goes wrong
-   (see {ref}`how-to-errors`) or when a scalar optimizer is used on a specialized
-   problem.
+   (see {ref}`how-to-errors`).
 
 Any marked function can still be solved with a normal scalar optimizer; optimagic
 aggregates the vector output when needed (sum of squares for least-squares, sum of
--- a/docs/source/explanation/aggregation_level.md
+++ b/docs/source/explanation/aggregation_level.md
@@ -12,9 +12,9 @@ optionally `@om.mark.scalar`). That mark changes:
 
 1. **What your function should return** (a single number vs a vector of contributions or
    residuals).
-2. **Which specialized optimizers you can use** (for example pounders for least-squares,
+1. **Which specialized optimizers you can use** (for example pounders for least-squares,
    or BHHH for likelihood).
-3. **How error penalties and derivatives are interpreted** when something goes wrong
+1. **How error penalties and derivatives are interpreted** when something goes wrong
    (see {ref}`how-to-errors`) or when a scalar optimizer is used on a specialized
    problem.
 
--- a/docs/source/explanation/aggregation_level.md
+++ b/docs/source/explanation/aggregation_level.md
@@ -14,12 +14,18 @@ optionally `@om.mark.scalar`). That mark changes:
    residuals).
 1. **Which specialized optimizers you can use** (for example pounders for least-squares,
    or BHHH for likelihood).
+<<<<<<< HEAD
 1. **How error penalties and derivatives are interpreted** when something goes wrong or
    when a scalar optimizer is used on a specialized problem.
+=======
+3. **How error penalties and derivatives are interpreted** when something goes wrong
+   (see {ref}`how-to-errors`) or when a scalar optimizer is used on a specialized
+   problem.
+>>>>>>> 04ce067 (docs: address review on AggregationLevel explanation)
 
-The same marked function can often still be solved with a normal scalar optimizer;
-optimagic aggregates the vector output when needed (sum of squares for least-squares,
-sum of contributions for likelihood).
+Any marked function can still be solved with a normal scalar optimizer; optimagic
+aggregates the vector output when needed (sum of squares for least-squares, sum of
+contributions for likelihood).
 
 ## Scalar problems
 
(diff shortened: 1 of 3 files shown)


docs/source/explanation/aggregation_level.md
```diff
@@ -0,0 +1,94 @@
...
+optionally `@om.mark.scalar`). That mark changes:
+
+1. **What your function should return** (a single number vs a vector of contributions or
+   residuals).
+1. **Which specialized optimizers you can use** (for example pounders for least-squares,
+   or BHHH for likelihood).
+1. **How error penalties and derivatives are interpreted** when something goes wrong or
+   when a scalar optimizer is used on a specialized problem.
+
+The same marked function can often still be solved with a normal scalar optimizer;
```
janosg: Not just often. All function can be solved with scalar optimizers.

docs/source/explanation/aggregation_level.md
```diff
@@ -0,0 +1,94 @@
...
+You tell optimagic which kind you have by marking the objective function with a
+decorator from `optimagic.mark` (`@om.mark.least_squares`, `@om.mark.likelihood`, or
+optionally `@om.mark.scalar`). That mark changes:
+
+1. **What your function should return** (a single number vs a vector of contributions or
+   residuals).
+1. **Which specialized optimizers you can use** (for example pounders for least-squares,
+   or BHHH for likelihood).
+1. **How error penalties and derivatives are interpreted** when something goes wrong or
+   when a scalar optimizer is used on a specialized problem.
```
janosg: We should link to the error handling docs page here.

docs/source/explanation/aggregation_level.md
```diff
@@ -0,0 +1,94 @@
...
+**Sign / maximize vs minimize:** return the actual log-likelihood contributions (the
+quantities you would sum to get $\ell(\theta)$). Prefer {func}`~optimagic.maximize` for
+maximum likelihood; optimagic flips the sign internally for the solver. If you prefer
+{func}`~optimagic.minimize`, return **negative** log-likelihood contributions instead.
+
+Do **not** return only the summed scalar log-likelihood if you want likelihood-specific
+optimizers — they need the contributions.
+
+For estimation workflows built on likelihood functions, see also the estimagic
+tutorials.
```
janosg: We should add a link here.

docs/source/explanation/aggregation_level.md
```diff
@@ -0,0 +1,98 @@
...
+   residuals).
+1. **Which specialized optimizers you can use** (for example pounders for least-squares,
+   or BHHH for likelihood). \<<\<<\<<< HEAD
+1. **How error penalties and derivatives are interpreted** when something goes wrong or
+   when a scalar optimizer is used on a specialized problem. =======
+1. **How error penalties and derivatives are interpreted** when something goes wrong
+   (see {ref}`how-to-errors`) or when a scalar optimizer is used on a specialized
+   problem.
+
+> > > > > > > 04ce067 (docs: address review on AggregationLevel explanation)
```
janosg: There is a not cleaned-up merge conflict.

docs/source/explanation/aggregation_level.md
```diff
@@ -0,0 +1,94 @@
...
+decorator from `optimagic.mark` (`@om.mark.least_squares`, `@om.mark.likelihood`, or
+optionally `@om.mark.scalar`). That mark changes:
+
+1. **What your function should return** (a single number vs a vector of contributions or
+   residuals).
+1. **Which specialized optimizers you can use** (for example pounders for least-squares,
+   or BHHH for likelihood).
+1. **How error penalties and derivatives are interpreted** when something goes wrong
+   (see {ref}`how-to-errors`) or when a scalar optimizer is used on a specialized
+   problem.
```
janosg:
```suggestion
   (see {ref}`how-to-errors`).
```

janosg (approved):
Thanks for the PR. I made one tiny change and will merge it soon.

Issue #695, opened 2026-08-02
Use pydantic validation for constraint classes (after constraints refactor lands)


janosg:
## Background

Algorithm options are now validated by pydantic via `validated_dataclass` in `optimagic/typing.py` (see also #694 for the option dataclasses).

`optimagic/constraints.py` contains ~20 `__post_init__` methods. A part of what they do is per-attribute type checking that pydantic could take over, so the constraint classes are natural candidates for the same treatment.

## Why this should wait

The typed-constraints refactor is currently in flight as a stack of branches touching exactly this file. Migrating the validation now would create constant rebase conflicts. This issue should be tackled **after the constraints refactor stack has been merged**.

## Notes for implementation

- Most constraint `__post_init__` logic is semantic (bounds relations, weight lengths, index compatibility), which stays in `__post_init__` — pydantic dataclasses run it after field validation. The win is smaller than for the option dataclasses, so evaluate per class whether the migration pays off.
- Constraint fields hold PyTrees and selector callables; fields where pydantic adds no value can keep plain annotations (`Any`, `Callable`) and rely on the semantic checks.
- Translate `pydantic.ValidationError` into `InvalidConstraintError` to keep the public exception contract.

m-akhil-reddy:
@janosg @timmens 
hi janosg,
myself akhil, i would like to contribute to optimagic.
how can i help you and contribute to this repo.

janosg:
Hi @m-akhil-reddy, currently the best way to contribute to optimagic ist to use it, stress test it and suggests improvements or new features. We are currently discussing a policy for code contributions in the age of AI internally and will document it as soon as we are done.

Issue #694, opened 2026-08-02
Use pydantic validation for MultistartOptions, NumdiffOptions and ScalingOptions


janosg:
## Background

Algorithm options are now validated and converted by pydantic: `mark.minimizer` re-creates each algorithm class as a pydantic dataclass via `validated_dataclass` in `optimagic/typing.py`. This replaced the hand-written converter table and works with postponed annotation evaluation and optional types like `PositiveInt | None`.

The same approach can replace the manual validation boilerplate in the user-facing option dataclasses:

- `MultistartOptions` (`optimization/multistart_options.py`): `_validate_attribute_types_and_values` is ~160 lines of isinstance checks that duplicate information already present in the precise field annotations.
- `NumdiffOptions` (`differentiation/numdiff_options.py`): same pattern.
- `ScalingOptions` (`parameters/scaling.py`): same pattern.

## Suggested approach

- Decorate the option classes with `validated_dataclass` (lax config `DEFAULT_PYDANTIC_CONFIG`, matching the convention that user-facing inputs are converted generously), translating `pydantic.ValidationError` into the existing exception types (`InvalidMultistartError`, ...) so the public exception contract is unchanged.
- Keep cross-field semantic checks (e.g. `batch_size >= n_cores`) in `__post_init__`; pydantic dataclasses run `__post_init__` after field validation.
- The `pre_process_*` functions keep handling the generous union inputs (`bool | dict | Options`); only the per-attribute type checking moves to pydantic.
- `BatchEvaluator` is already `@runtime_checkable`, so protocol-typed fields validate fine.

## Payoff

- Deletes several hundred lines of validation boilerplate.
- Aggregated error reports (all invalid attributes at once instead of first-fail).
- Values are converted (e.g. `"2"` or `2.0` for an int field), not just checked.

PR #696, opened 2026-08-02
Use pydantic for algo option validation (instead of our own validation logict).

janosg:
## Current situation

Algorithm options are converted and validated by a hand-written converter table (TYPE_CONVERTERS in type_conversion.py) that Algorithm.__post_init__ looks up per field, keyed by type objects for eight supported types. AlgoInfo and InternalOptimizeResult carry ~50 lines of manual isinstance checks each.

## Two problems are hard to solve within this design:

- In modules with from __future__ import annotations — which the how-to-document-optimizers guide requires — field.type is an annotation string, so the table lookup silently fails and options are neither coerced nor validated. This already affects most optimizer modules.
- Anything beyond the eight exact types is skipped: PositiveInt | None, Literals, unions, nested option types. Supporting them would mean reimplementing generic type introspection and coercion — exactly what pydantic provides. Notably, our type aliases are already built on annotated-types (Annotated[int, Gt(0)]), which is pydantic's native constraint vocabulary.

## Proposed Implementation

Replace our validation logic with pydantic. 

## Backwards compatibility

- Algorithm classes (including user-defined ones) are still written as plain frozen dataclasses; validation is added by `mark.minimizer`, so no algorithm definition changes.
- All public exception types and message prefixes are preserved (`InvalidAlgoOptionError`, `InvalidAlgoInfoError`, `TypeError` for `InternalOptimizeResult`), including for unknown constructor kwargs (`extra="forbid"`).
- The internal classes use strict mode, mirroring the old isinstance semantics.

Remaining breaking changes:

- Fractional floats for int options raise instead of being silently truncated (stopping_maxiter=5.5 was 5; integral floats like 1e6 still coerce). Numeric strings in float notation ("1e6") are rejected for int options; plain int strings still work.
- Fields that were previously unvalidated (Literals, unions, str, arrays, ...) are now enforced, so invalid values that used to pass silently now raise. 

Alternatives

PR #691 patches the string-annotation symptom by adding a second, string-keyed converter table (TYPE_CONVERTERS_BY_NAME) and branching on the annotation kind in __post_init__. That fixes the silently-skipped coercion but keeps everything else: optional and union types remain unvalidated, coverage stays limited to eight hand-written converters, and two parallel tables must be kept in sync with typing.py forever. This PR solves the whole class of problems, deletes code instead of adding it, and extends naturally to the other validation boilerplate in the codebase (#694, #695). If this is merged, #691 should be closed and #690 rebased onto main afterwards.


timmens (approved):
Very nice; I couldn't find anything to flag.

PR #697, opened 2026-08-02
ENH: support SciPy-style callback(xk) during optimization

Pragati5-DEBUG:
## Summary
- Support SciPy-style `callback(xk)` in `minimize` / `maximize`
- Pass it into `InternalOptimizationProblem` and call it on objective evaluations (works for all optimizers, not via SciPy's own callback)
- `callback(intermediate_result)` left for a follow-up

## Test plan
- [x] `pytest tests/optimagic/optimization/test_scipy_aliases.py::test_callback_xk_is_called`
- [x] Related optimization tests (96 passed)


src/optimagic/optimization/create_optimization_problem.py
```diff
@@ -86,6 +86,7 @@ class OptimizationProblem:
     skip_checks: bool
     direction: Direction
     fun_eval: SpecificFunctionValue
+    callback: Callable[[Any], Any] | None
```
janosg: This type hint could be much tighter; Currently we only accept callables that take a PyTree (the current parameter Vector) and doesn't return anything. 

Note: In contrast to scipy xk is not necessarily a numpy array, it can be an arbitrary PyTree.

src/optimagic/optimization/create_optimization_problem.py
```diff
@@ -551,6 +544,7 @@ def create_optimization_problem(
         skip_checks=skip_checks,
         direction=direction,
         fun_eval=fun_eval,
+        callback=callback,
```
janosg, line 562: Currently you don't do any validation of the user passed callback. For example, if the user passes a callable with the wrong signature, this would lead to a hard to understand error in the first iteration. We need the same validation for callbacks as we have for all other user evaluated functions like fun, jac, etc. 

We also need tests that invalid callbacks are handled well.

src/optimagic/optimization/internal_optimization_problem.py
```diff
@@ -265,6 +267,11 @@ def with_step_id(self, step_id: int) -> Self:
         new._step_id = step_id
         return new
 
+    def _maybe_call_callback(self, x: NDArray[np.float64]) -> None:
+        """Call the optional SciPy-style ``callback(xk)`` if one was provided."""
+        if self._callback is not None:
+            self._callback(x)
```
janosg: SciPy callbacks can raise StopIteration, which aborts an optimization gracefully. This is probably a bit out of scope for this PR, but we need to document that it is not handled.

src/optimagic/optimization/internal_optimization_problem.py
```diff
@@ -652,6 +660,7 @@ def func(x: NDArray[np.float64]) -> SpecificFunctionValue:
             exceptions=traceback,
         )
 
+        self._maybe_call_callback(x)
```
janosg: We should not evaluate the callback at derivative evaluations. Only at objective evaluations.

src/optimagic/optimization/internal_optimization_problem.py
```diff
@@ -504,6 +511,7 @@ def _pure_evaluate_fun(
...
```
janosg: You are calling the callback with the internal parameter vector (a flat numpy array with entries that a user might not be able to interpret). Instead the callback should be called with the external parameters (which could be dictionaries or arbitrary PyTrees. We have this pattern for any user supplied function.

janosg (changes requested):
Hi @Pragati5-DEBUG 

I left a few comments. After reading up a bit more on SciPy callbacks, I am also realizing there are two conceptual problems with the implementation sketch I gave you. 

1. SciPy callbacks are evaluated after each Iteration, not after each function evaluation. For gradient based optimizers this is typically the same but for gradient free ones it is not. Optimagic cannot know what an algorithm calls an iteration, therefore any algorithm agnostic implementation of callbacks can only work on function evaluations, which is useful but not exactly what SciPy does. 

I therefore suggest that for all optimizers that support SciPy style callbacks (we'll need a flag in AlgoInfo for that), we use the algorithm's callback mechanism instead of bypassing it and just add parameter conversion to get from the internal parameter vector of the optimizer to the external one the callback expects. For all other optimizers we go with our own callback implementation that works on function evaluations. That way we get full SciPy compatibility and a useful extension for all non scipy optimizers.  

2. The `_pure` functions are actually not the right level to implement callbacks. We would get problems with parallelizing optimizers. With n_cores > 1, the callback executes in worker processes. Common callback state such as xs.append(...), counters, or in-memory progress objects will not be updated in the parent process; the callback must also be serializable, and side effects can occur concurrently. This is the exact same reason why history collection is not implemented inside the `_pure... ` functions. A rule of thumb is that callback evaluations should probably happen next to history append operations to avoid this problem. 

I am realizing that this makes the feature much harder than anticipated. Let me and @timmens  know if you want us to take over the PR.

Pragati5-DEBUG:
Thanks @janosg for the detailed review — that all makes sense.

I’d like to keep working on this. Could I please have some time to read up on SciPy callbacks and understand the design better before implementing the changes? I’ll update the PR once I’m ready.

janosg:
Sure, take as long as you need.

Pragati5-DEBUG:
Thanks @janosg — I’ve started on the redesign in two phases.

**Phase 1 (in progress / about to push):** the fun-eval fallback path, made parallel-safe and consistent with your review comments:
- move callback next to history append (not in `_pure_*`)
- pass external PyTree params
- call only on objective evaluations (not jac-only)
- tighten types + validate callback like `fun`/`jac`
- document that `StopIteration` is not handled yet
- add/update tests

**Phase 2 (follow-up):** full SciPy compatibility where possible:
- `AlgoInfo` flag for optimizers that support SciPy-style callbacks
- use the algorithm’s own callback + internal→external conversion for those
- keep the Phase 1 fun-eval callback for all other optimizers

I’ll push Phase 1 first for review, then tackle Phase 2. Happy to adjust the split if you’d prefer a different order.

Pragati5-DEBUG:
Hi @janosg, Phase 1 is ready for re-review.

I've rebased onto latest `main`. Changes from your feedback:
- Callback next to history append (not in `_pure_*`) — parallel-safe with `n_cores > 1`
- External PyTree params (not internal flat `x`)
- Only on objective evals (not jac-only)
- Callback validated like `fun` / `jac`
- `StopIteration` documented as not handled yet
- Tests updated; related scipy-alias tests pass locally

Phase 2 (AlgoInfo + native SciPy callbacks) as follow-up after this lands.

Thanks!

src/optimagic/optimization/internal_optimization_problem.py
```diff
@@ -265,6 +273,20 @@ def with_step_id(self, step_id: int) -> Self:
         new._step_id = step_id
         return new
 
+    def _maybe_call_callback(self, params: PyTree) -> None:
+        """Call the optional SciPy-style ``callback(xk)`` if one was provided.
+
+        Called next to history append (not inside ``_pure_*``) so it runs in the
+        parent process when ``n_cores > 1``. ``params`` are external user-facing
+        parameters (a PyTree), not the internal flat parameter vector.
```
janosg: The docstring of a function should not explain where this function is called or not called. Only what the function does. Then a potential user of the function can decide where to call it.

src/optimagic/optimization/optimize.py
```diff
@@ -76,8 +76,8 @@
 JacType = Callable[..., PyTree]
 FunAndJacType = Callable[..., tuple[float | PyTree | FunctionValue, PyTree]]
 HessType = Callable[..., PyTree]
-# TODO: refine this type
-CallbackType = Callable[..., Any]
+# SciPy-style callback(xk); xk is the external parameter PyTree. Returns None.
```
janosg: The comment is not needed because the type hint is informative enough now.

src/optimagic/optimization/optimize.py
```diff
@@ -216,7 +216,13 @@ def maximize(
...
         hess: Not yet supported.
         hessp: Not yet supported.
-        callback: Not yet supported.
+        callback: Optional SciPy-style callback with signature ``callback(xk)``, where
+            ``xk`` is the current **external** parameter value (a PyTree; for array
+            ``params`` this is a numpy array). Called next to history collection after
+            objective evaluations (not inside parallel worker ``_pure_*`` functions, and
+            not on derivative-only evaluations). Raising ``StopIteration`` to abort
+            optimization (as in SciPy) is not handled yet. The
+            ``callback(intermediate_result)`` interface is not yet supported.
```
janosg: This is the user facing documentation of the callback feature. So it should not mention any implementation details. It should tell the user what they can do with callbacks and which callbacks are accepted but not mention unrelated features like history collection. 

You can take the scipy documentation as inspiration.

src/optimagic/optimization/optimize.py
```diff
@@ -413,7 +419,13 @@ def minimize(
...
```
janosg: Same.

tests/optimagic/optimization/test_scipy_aliases.py
```diff
@@ -120,14 +120,107 @@ def test_exception_for_hessp():
         )
 
 
-def test_exception_for_callback():
-    msg = "The callback argument is not yet supported"
-    with pytest.raises(NotImplementedError, match=msg):
+def test_callback_xk_is_called():
```
janosg: These tests should be in a new test file called `test_callback.py`

tests/optimagic/optimization/test_scipy_aliases.py
```diff
@@ -120,14 +120,107 @@ def test_exception_for_hessp():
...
+    """SciPy-style callback(xk) is invoked on objective evaluations."""
+    xs = []
+
+    def callback(xk):
+        xs.append(np.asarray(xk).copy())
```
janosg: You don't need a copy here.

tests/optimagic/optimization/test_scipy_aliases.py
```diff
@@ -120,14 +120,107 @@ def test_exception_for_hessp():
...
+    res = om.minimize(
+        fun=lambda x: x @ x,
+        x0=np.arange(3, dtype=float),
+        algorithm="scipy_neldermead",
+        callback=callback,
+    )
+
+    assert len(xs) >= 1
+    assert xs[0].shape == (3,)
+    aaae(res.x, np.zeros(3), decimal=5)
```
janosg: You can actually test this against the exact values of the collected history.

tests/optimagic/optimization/test_scipy_aliases.py
```diff
@@ -120,14 +120,107 @@ def test_exception_for_hessp():
...
+        )
+
+
+def test_invalid_callback_not_callable():
+    with pytest.raises(InvalidFunctionError, match="callback must be a callable"):
+        om.minimize(
+            fun=lambda x: x @ x,
+            x0=np.arange(3, dtype=float),
+            algorithm="scipy_neldermead",
+            callback="not-a-callable",
```
janosg: Can you please add a test that implements history collection via a callback for a parallelizing optimizer and compares it against the history you get from optimagic?

janosg (changes requested):
Hi @Pragati5-DEBUG, this goes into the right direction. I left a few more comments.

Pragati5-DEBUG:
@janosg Thanks for the re-review
Ready for another look when you have time. Thanks!

